PR #110, opened 2023-04-16
allow timestamp to be passed from command line [skip ci]

wolfv:


wolfv:
Closing this for now, but I want to make a `rebuild mypkg.tar.bz2` command instead that copies prefix-placeholder and timestamp from an existing package.

PR #131, opened 2023-04-26
Feature/source local path

### LINKED ISSUE #108: Implement `local` source
Implement fetching data from a local path

ruben-arts:
Add the ability to use a local path as source in a recipe. 

closes #108

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 book
 output/
 target/
+.idea/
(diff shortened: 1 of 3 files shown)


ruben-arts:
> If we want to have something even more fancy (with the potential of adding a progress bar) we could look into `fs_extra`

Yeah sure if we need that later on. Think this should be fine for now :smile:

PR #137, opened 2023-05-01
Feature/source git

ruben-arts:
Add git as a source for the package building as specified [here](https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html?highlight=git#source-from-git)

What is still missing is `git-lfs` integration and `git_depth` cloning.

### DIFF
--- a/src/source.rs
+++ b/src/source.rs
@@ -370,11 +370,11 @@ pub async fn fetch_sources(
 
 #[cfg(test)]
 mod tests {
-    use std::str::FromStr;
-    use git2::Repository;
     use super::*;
-    use url::Url;
     use crate::metadata::GitRev;
+    use git2::Repository;
+    use std::str::FromStr;
+    use url::Url;
 
     #[test]
     fn test_split_filename() {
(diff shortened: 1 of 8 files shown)


src/source.rs
```diff
@@ -129,8 +132,84 @@ async fn url_src(
     Ok(cache_name)
 }
 
-fn git_src(_source: &GitSrc) {
-    todo!("Git source support not implemented yet!");
+/// Fetch the repo and its specific refspecs.
```
wolfv, line 135: Can you document how it works with the `src_cache` dir?

wolfv:
I think cloning with depth is still a bit out: https://github.com/libgit2/libgit2/pull/6396

In general, I am wondering if we should use `git2` or just invoke `git` on the command line. The reasoning for the latter:

- Is probably configured properly with SSH keys etc. (e.g. in Github Actions, local computers, ...)
- Super widely used & available
- It's what `conda-build` does
- No additional dependencies means smaller install size of rattler-build

On the other hand, using `git2` means:

- no dependencies on host needed
- potential interesting future possibilities ... e.g. when we need to search for older recipes in repositories or things like that
- more control, probably!?

Regardless, I'm happy to merge this PR once we have a little more documentation (I would like to understand how / if the source_cache works with git repositories).

ruben-arts:
> I am wondering if we should use git2 or just invoke git on the command line.

@wolfv I'm also not sure. I continued this direction to hopefully avoid OS-specific differences in the cli. Plus the error handling is probably easier as now you just have everything in rust. But I get what you saying. I could try to redo it in another branch to see if that would work better. 
@baszalmstra @tdejager do you guys have an opinion?

baszalmstra:
I think cargo is also originally using `git2` so Im pretty sure it will work more than good enough for our usecase. So I would just stick to that. 

However, I read somewhere that `cargo` is moving to [`gix`](https://github.com/Byron/gitoxide). I think its behind a feature flag atm.

PR #142, opened 2023-05-21
fix: relative git_url paths

### LINKED ISSUE #140: Relative paths with git url
It was reported that relative paths don't seem to work properly with the git source yet.

ruben-arts:
My bad there was no canonicalization of the path. 
closes #140

### DIFF
--- a/src/source.rs
+++ b/src/source.rs
@@ -168,11 +168,21 @@ fn fetch_repo(repo: &Repository, refspecs: &[String]) -> Result<(), git2::Error>
 ///
 /// # Returns
 /// - A Result containing the PathBuf to the cache, or a SourceError if an error occurs during the process.
-fn git_src<'a>(source: &'a GitSrc, cache_dir: &'a Path) -> Result<PathBuf, SourceError> {
+fn git_src<'a>(
+    source: &'a GitSrc,
+    cache_dir: &'a Path,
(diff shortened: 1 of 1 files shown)


src/source.rs
```diff
@@ -168,11 +168,21 @@ fn fetch_repo(repo: &Repository, refspecs: &[String]) -> Result<(), git2::Error>
...
+    source: &'a GitSrc,
+    cache_dir: &'a Path,
+    recipe_dir: &'a Path,
+) -> Result<PathBuf, SourceError> {
     // Create cache path based on given cache dir and name of the source package.
     let filename = match &source.git_url {
         GitUrl::Url(url) => url.path_segments().unwrap().last().unwrap().to_string(),
-        GitUrl::Path(path) => path.file_name().unwrap().to_string_lossy().to_string(),
+        GitUrl::Path(path) => recipe_dir
+            .join(path)
```
wolfv, line 180: is this necessary? I think filename should also work on a relative path?
Or maybe one should find out if the path is relative first, and then run the `join` code?

wolfv: Hmm, the docs say: "If path is absolute, it replaces the current path." -- in that case I guess I am fine with this code :)

wolfv (approved):

PR #143, opened 2023-05-25
improve error reporting when parsing files and remove some unwrap

wolfv:


src/main.rs
```diff
@@ -272,7 +272,20 @@ async fn run_build_from_args(args: BuildOpts) -> anyhow::Result<()> {
...
+            Result::Err(e) => {
+                match &e {
+                    render::recipe::RecipeRenderError::InvalidYaml(inner) => {
+                        tracing::error!("Failed to parse recipe YAML: {}", inner.to_string());
+                    }
+                    render::recipe::RecipeRenderError::YamlNotMapping => {
+                        tracing::error!("{}", e);
+                    }
+                }
+                return Err(e.into());
```
ruben-arts, line 285: returning it again as an error prints it twice in the terminal. The tracing one is nice as it prints in red plus the source crate is mentioned.

src/render/recipe.rs
```diff
@@ -127,17 +126,28 @@ fn render_context(yaml_context: &serde_yaml::Mapping) -> BTreeMap<String, Value>
     context
 }
 
+#[derive(Debug, thiserror::Error)]
+pub enum RecipeRenderError {
+    #[error("Invalid YAML")]
+    InvalidYaml(#[from] serde_yaml::Error),
+
+    #[error(
+        "YAML does not follow regular recipe structure (map with build, requirements, outputs...)"
```
ruben-arts, line 135: This is not testing a certain schema right? The error can be a little more generic
```suggestion
        "Invalid Recipe File Format. The recipe file you've provided is not correctly formatted as a YAML mapping value. Please ensure your file adheres to the YAML standard and try again."
```

ruben-arts (dismissed):

PR #146, opened 2023-06-07
Fix vinca recipes

ruben-arts:
@wolfv


wolfv:
I think you forgot to push the file `flatten_3.yml`

PR #160, opened 2023-07-06
Update README.md

raybellwaves:
am I right this should be --recipe instead of --recipe-file ?


ruben-arts:
Hey you're completely right!

PR #174, opened 2023-07-26
copy_dir with gitignore options

wolfv:
This will adhere to the gitignore file found in the directory to copy and also allows include / exclude globs to be specified for further filtering.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -364,6 +364,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "bstr"
+version = "1.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


matthiasbeyer:
I am overtaking this PR as discussed.

To summarize what we discussed and what I understood (please correct if necessary):

* The `ignore` crate shouldn't be used anymore because it is tailored towards git and we possibly need another syntax (because the `!` prefix for negation is a yaml builtin)
* The `copy_dir` function should be integrated into recipe processing so that the `license_file` key can use a list of pathes and filters and that gets processed properly. Later this integration should also be integrated for other keys in the recipe.

wolfv:
Yep, that sounds good!

PR #181, opened 2023-09-14
fix: minor code nit

swarnimarun:
Pretty useless functionally but still manual implementation of a transpose, is redundant. Hope no one minds this. `:P`


wolfv:
LGTM!

PR #202, opened 2023-10-10
WIP Multi outputs

wolfv:
This is the start of implementing support for multiple outputs (and also similar to implement support for multiple _recipes_).

Some points for discussion:

- evaluate and order the build tree, which is dependent on finding the edges of the tree. The edges can be strongly or loosely connected by a `pin_subpackage` expression. If the `pin_subpackage` pins "exact" we might have to build more variants. I am going to create a diagram for this.
- The PR takes the recipe with multiple outputs and splits it into multiple recipes. It follows some merging rules:
  - if the output does not define a version, use the version of the toplevel recipe
  - Should we make the top-level "name" optional? conda-forge uses it for the feedstock name I think
  - We merge the keys of the `about` section, as well as `extra` and `build`, although for `build` the main relevant fields are `number`, `string` and maybe `script`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1403,9 +1403,9 @@ dependencies = [
 
 [[package]]
 name = "insta"
-version = "1.33.0"
+version = "1.34.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1aa511b2e298cd49b1856746f6bb73e17036bcd66b25f5e92cdcdbec9bd75686"
+checksum = "5d64600be34b2fcfc267740a243fa7744441bb4947a619ac4e5bb6507f35fbfc"
 dependencies = [
  "console",
  "lazy_static",
@@ -3397,9 +3397,9 @@ checksum = "1f3ccbac311fea05f86f61904b462b55fb3df8837a366dfc601a0161d0532f20"
 
 [[package]]
 name = "tokio"
-version = "1.32.0"
+version = "1.33.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "17ed6077ed6cd6c74735e21f37eb16dc3935f96878b1fe961074089cc80893f9"
+checksum = "4f38200e3ef7995e5ef13baec2f432a6da0aa9ac495b2c0e8f3b7eec2c92d653"
 dependencies = [
  "backtrace",
  "bytes 1.5.0",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ sha2 = "0.10.8"
 hex = "0.4.3"
 serde_json = "1.0.107"
 reqwest = "0.11.22"
-tokio = {version = "1.32.0", features = ["rt", "macros", "rt-multi-thread"] }
+tokio = {version = "1.33.0", features = ["rt", "macros", "rt-multi-thread"] }
 itertools = "0.11.0"
 content_inspector = "0.2.4"
 serde_with = "3.3.0"
@@ -61,7 +61,7 @@ clap-verbosity-flag = "2.0.1"
 tracing-core = "0.1.31"
 
 [dev-dependencies]
-insta = {version = "1.33.0", features = ["yaml"] }
+insta = {version = "1.34.0", features = ["yaml"] }
 rstest = "0.18.2"
 
 [profile.dev.package."*"]
--- a/src/main.rs
+++ b/src/main.rs
@@ -239,114 +239,119 @@ async fn run_build_from_args(args: BuildOpts, multi_progress: MultiProgress) -> 
 
     let variant_config = VariantConfig::from_files(&args.variant_config, &selector_config)?;
 
-    let variants = variant_config
-        .find_variants(&recipe_text, &selector_config)
+    let outputs_variants = variant_config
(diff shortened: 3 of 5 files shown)


wolfv:
Scenario 1 where we have a Python specific libmambapy and a Python specific mamba that are connected by a `pin_subpackage(..., exact=True). In this case we need to expand the matrix, but only for "solveable" combinations (python=3.8 & libmambapy=*=*py38, ...).

Screenshot 2023-10-10 at 10 50 13

Scenario 2 is a bit simpler, here we only have two variant packages on top. The exact pinned lower packages should also produce two variants.

Screenshot 2023-10-10 at 11 22 16

Or, if they are not pinned exact, it's enough to produce a single output for the final package.

Screenshot 2023-10-10 at 11 02 32

wolfv:
This is how we compute the "variant overlap" and build tree order in `boa`: https://github.com/mamba-org/boa/blob/d332420fc2c764e0077e2b02071469a6f0b73d23/boa/core/variant_arithmetic.py#L225-L278

wolfv:
Variant computation algorithm! 


- For a single package, it's modestly simple. There are two inputs:
    - The used variables and dependencies (incl. run_exports) of the package
    - The conda_build_config.yaml that contains the variant configuration and `zip_keys`.

For the following examples, let's consider the following variant config:

```yaml
python:
  - 3.8
  - 3.9
  - 3.10
numpy:
  - 1.19
  - 1.20
  - 2
```

- Simple variants: if our recipe depends on `python`, we create 3 variants with 3.8, 3.9, and 3.10.
- Zip Keys "zip up" some variants into a single variant. If our variant configuration would contain the statement
  `zip_keys: [python, numpy]` the first condition is that we have the same number of `numpy` variants as `python`.
  They are then zipped, so that we end up with a "joint variant" of `python_numpy: [[3.8, 1.19], [3.9, 1.20], [3.10, 2]]`.

## Multiple outputs
- Multiple outputs behave like independent recipes that can be "strongly" connected with `pin_subpackage` dependencies. 
  When strong `pin_subpackage` connections are introduced, variant computation also needs to be altered.
  Let's consider the following recipe:

```yaml
outputs:
  - package:
      name: libfoo
    requiremnts:
      host:
        - python
  - package:
      name: foofoo
    requiremnts:
      host:
        - python
        - ${{ pin_subpackage('libfoo', exact=True) }}
```

  We build `libfoo py38`, `libfoo py39` and `libfoo py310`. 
  Now for `foofoo`, we expect to build `foofoo py38`, `foofoo py39` and `foofoo py310` (stemming from python variants). However, since there is also
  an exact pin on `libfoo`, we also expect 3 variant packages for `libfoo` to be built. We need to find the shared variants between the exactly pinned subpackages 
  and create a joint variant for them. In this case, we would end up with 3 variants for `libfoo` and 9 variants for `foofoo`. Of the 9 variants for foofoo only 3 are solvable.
  We create a pin_subpackage_zip to create only the solvable subset of variants:

  - py38  libfoo-py38
  - py38  libfoo-py39 (UNSOLVABLE! as are the others)
  - py39  libfoo-py39
  - py310 libfoo-py310

  So the pin_subpackage_zip looks like `python_libfoo: [[3.8, libfoo py38], [3.9, libfoo py39], [3.10, libfoo py310]]`.


```yaml
outputs:
  - package:
      name: libfoo
    requiremnts:
      host:
        - python
  - package:
      name: foofoo
    requiremnts:
      host:
        - python
        - numpy   # here we need to be careful to still `zip` the variants with `python` and `numpy` together
        - ${{ pin_subpackage('libfoo', exact=True) }}
```

- py38  [numpy 1.19] [libfoo py38]
- py39  [numpy 1.20] [libfoo py39]
- py310 [numpy 2]    [libfoo py39]

Let's assume we also have two boost versions (17 and 18), and only libfoo depends on them.

```yaml
outputs:
  - package:
      name: libfoo
    requiremnts:
      host:
        - python
        - boost
  - package:
      name: foofoo
    requiremnts:
      host:
        - python
        - ${{ pin_subpackage('libfoo', exact=True) }}
```

Now the variants look like:

- libfoo py38  boost 17
- libfoo py38  boost 18
- libfoo py39  boost 17
- libfoo py39  boost 18
- libfoo py310 boost 17
- libfoo py310 boost 18

And the variants for foofoo are:

- py38  [libfoo py38  boost 17]
- py38  [libfoo py38  boost 18]
- py39  [libfoo py39  boost 17]
- py39  [libfoo py39  boost 18]
- py310 [libfoo py310 boost 17]
- py310 [libfoo py310 boost 18]

Okay now I am considering to build the full variant tree, and then subtract everything that is not solvable based on examining the run dependencies / constraints of the package.

Jaime's case:

```yaml
package:
  version: 1.0.0
outputs:
  - package:
      name: libfoo
    build:
      string: ${{PKG_HASH}}_{GPU_OR_CPU}
  - package:
      name: foofoo
    requiremnts:
      host:
        - ${{ pin_subpackage('libfoo', exact=True) }}
```

In this case we get

- libfoo 1.0.0  GPU
- libfoo 1.0.0  CPU

And for foofoo we get

- foofoo 1.0.0  [libfoo 1.0.0 GPU]
- foofoo 1.0.0  [libfoo 1.0.0 CPU]

PR #209, opened 2023-10-18
implement python post processing for INSTALLER and add PYTHON variabl…

wolfv:
…e to env vars

### DIFF
--- a/src/post.rs
+++ b/src/post.rs
@@ -12,7 +12,7 @@ use std::{
     path::{Path, PathBuf},
 };
 
-use rattler_conda_types::Platform;
+use rattler_conda_types::{PackageName, Platform};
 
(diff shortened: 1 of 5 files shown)


src/post.rs
```diff
@@ -83,3 +83,45 @@ pub fn relink(
...
+) -> Result<(), std::io::Error> {
+    let metadata_glob = globset::Glob::new("**/*.dist-info/METADATA")
+        .unwrap()
+        .compile_matcher();
+
+    if let Some(p) = paths.iter().find(|p| metadata_glob.is_match(p)) {
+        // unwraps are OK because we already globbed
+        let distinfo = p
+            .parent()
+            .unwrap()
```
baszalmstra, line 102: Please use expect instead of unwrap with a message on why this should never fail. That makes your assumption more explicit

PR #230, opened 2023-10-20
feat: implement test file copying with copy dir and improve url source

wolfv:
This adds a `file_name` field to URL source (different from `fn` in conda's meta.yaml).

I also had to modify copy_dir a bit to not delete the existing files in the target directory.

During the work on this, I also discovered that we are currently hard coding a `bash` invocation to execute the tests. While that works great on Unix, I think we need to make some additional adjustments for Windows and rewrite that part a little bit.

### DIFF
--- a/src/packaging.rs
+++ b/src/packaging.rs
@@ -26,6 +26,7 @@ use rattler_package_streaming::write::{
 
 use crate::macos;
 use crate::metadata::Output;
+use crate::source::copy_dir;
 use crate::{linux, post};
 
 #[derive(Debug, thiserror::Error)]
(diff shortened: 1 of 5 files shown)


wolfv:
@matthiasbeyer I've changed the copy_dir function a little bit more here (specifically to not erase the contents of the existing target directory).

Maybe - if you do the redesign of this function - you can consider this?

wolfv:
@matthiasbeyer I copied the include / exclude globs logic from you but I think for the exclude globs we need to add a `trim_start_matches('~')`, right? I don't think that can work as it is right now :)

PR #234, opened 2023-10-23
feat(recipe): Rendering yaml values before parsing

GrayJack:
- Moved the rendering of selectors before parsing
- Added support for `source.patches` multiple entries
- Added more checks for invalid field
- Extend error variants for better error messages
- Improve some error messages with help message


src/recipe/custom_yaml/rendered.rs
```diff
@@ -0,0 +1,719 @@
...
+    recipe::{
+        error::{ErrorKind, ParsingError, PartialParsingError},
+        jinja::Jinja,
+        Render,
+    },
+};
+
+use super::{HasSpan, MappingNode, Node, ScalarNode, SequenceNode, SequenceNodeInternal};
+
+/// A marked new Conda Recipe YAML node
```
baszalmstra, line 21: What does "marked" mean in this context?

src/recipe/custom_yaml/rendered.rs
```diff
@@ -0,0 +1,719 @@
...
+//                 } else {
+//                     Ok(RenderedNode::Null(RenderedScalarNode::new(
+//                         *if_sel.span(),
+//                         "".to_string(),
+//                     )))
+//                 }
+//             }
+//         }
+//     }
+// }
```
baszalmstra, line 719: Please just remove unused code.

src/recipe/parser.rs
```diff
@@ -0,0 +1,253 @@
...
+
+        // Set the build string to the package hash if it is not set
+        if recipe.build.string.is_none() {
+            recipe.build.string = Some(format!("{}_{}", default_pkg_hash, recipe.build.number));
+        }
+
+        Ok(recipe)
+    }
+
+    /// WIP
```
baszalmstra, line 79: Maybe add something here? :)

src/recipe/parser.rs
```diff
@@ -0,0 +1,253 @@
...
+        let mut about = About::default();
+
+        for (key, value) in rendered_node.iter() {
+            match key.as_str() {
+                "package" => package = Some(value.try_convert("package")?),
+                "source" => source = value.try_convert("source")?,
+                "build" => build = value.try_convert("build")?,
+                "requirements" => requirements = value.try_convert("requirements")?,
+                "test" => test = value.try_convert("test")?,
+                "about" => about = value.try_convert("about")?,
```
baszalmstra, line 142: Why not use `key` for the names here. :) Easier to not make mistakes!

src/recipe/parser/about.rs
```diff
@@ -149,40 +85,96 @@ impl About {
...
+            .and_then(|m| m.try_convert(name))
+    }
+}
+
+impl TryConvertNode<About> for RenderedMappingNode {
+    fn try_convert(&self, name: &str) -> Result<About, PartialParsingError> {
+        let mut about = About::default();
+        for (key, value) in self.iter() {
+            match key.as_str() {
+                "homepage" | "home" => about.homepage = Some(value.try_convert("homepage")?),
```
baszalmstra, line 101: Same, if you use `key` here for the names its less likely to make mistakes. And its more correct in the context!

src/recipe/parser/build.rs
```diff
@@ -0,0 +1,436 @@
...
+    }
+}
+
+impl TryConvertNode<Build> for RenderedMappingNode {
+    fn try_convert(&self, _name: &str) -> Result<Build, PartialParsingError> {
+        let mut build = Build::default();
+
+        for (key, value) in self.iter() {
+            match key.as_str() {
+                "number" => {
```
baszalmstra, line 124: Here as well. :)

src/recipe/parser/build.rs
```diff
@@ -0,0 +1,436 @@
...
+    }
+}
+
+impl TryConvertNode<RunExports> for RenderedMappingNode {
+    fn try_convert(&self, name: &str) -> Result<RunExports, PartialParsingError> {
+        let mut run_exports = RunExports::default();
+
+        for (key, value) in self.iter() {
+            match key.as_str() {
+                "noarch" => {
```
baszalmstra, line 363: And here as well!

src/recipe/parser/build.rs
```diff
@@ -0,0 +1,436 @@
...
+                    let deps = value.try_convert("noarch")?;
+                    run_exports.noarch = deps;
```
baszalmstra, line 365: Why the intermediate variable?

src/recipe/parser/package.rs
```diff
@@ -0,0 +1,138 @@
...
+        let mut version = None;
+        let span = *self.span();
+
+        for (key, value) in self.iter() {
+            match key.as_str() {
+                "name" => {
+                    name_val = Some(value.try_convert("name")?);
+                }
+                "version" => {
+                    version = Some(value.try_convert("version")?);
```
baszalmstra, line 55: Same here with the `key`

src/recipe/parser/package.rs
```diff
@@ -0,0 +1,138 @@
...
+    fn missing_fields() {
+        let raw_recipe = r#"
+        package:
+            name: test
+        "#;
+
+        let recipe = Recipe::from_yaml(raw_recipe, SelectorConfig::default());
+        assert!(recipe.is_err());
+
+        let err = recipe.unwrap_err();
```
baszalmstra, line 119: This is a bit duplicate, if there is no error the test will already panic. This happens in a few different places as well.

src/recipe/parser/requirements.rs
```diff
@@ -120,6 +83,51 @@ impl Requirements {
...
+impl TryConvertNode<Requirements> for RenderedMappingNode {
+    fn try_convert(&self, _name: &str) -> Result<Requirements, PartialParsingError> {
+        let mut build = Vec::new();
+        let mut host = Vec::new();
+        let mut run = Vec::new();
+        let mut run_constrained = Vec::new();
+
+        for (key, value) in self.iter() {
+            match key.as_str() {
+                "build" => build = value.try_convert("build")?,
```
baszalmstra, line 109: Same with the key.

src/recipe/error.rs
```diff
@@ -160,6 +188,10 @@ impl fmt::Display for ErrorKind {
             ErrorKind::IfSelectorMissingThen => {
                 write!(f, "Missing `then` field in the `if` selector.")
             }
+            ErrorKind::InvalidMd5 => write!(f, "Invalid MD5 checksum."),
+            ErrorKind::InvalidSha256 => write!(f, "Invalid SHA256 checksum."),
+            ErrorKind::InvalidField(s) => write!(f, "Invalid field `{s}`."),
+            ErrorKind::MissingField(s) => write!(f, "Missing field `{s}`"),
```
baszalmstra, line 194: I think its good practice to start error messages lowercase. This is more correct when you concatenate it to another string like:

```
ParseError: invalid md5 checksum.
```

src/recipe/parser/snapshots/rattler_build__recipe__parser__package__tests__missing_fields.snap
```diff
@@ -0,0 +1,14 @@
+---
+source: src/recipe/parser/package.rs
+expression: err
+---
+  × Parsing: Missing field `version`
+   ╭─[2:1]
+ 2 │         package:
+ 3 │             name: test
```
wolfv, line 8: Is this the correct position for the span?
I am wondering if underlining `package` would be better?

baszalmstra (commented):
I like this refactor a lot! Great work! 

I left a few comments to help reduce copy-paste-errors. 

`Recipe` is a struct that you pass when actually building a package. But now, because the fields are all private the only way to create the struct is by going through the recipe parsing. However, I think it is important that `Recipe`s can also be constructed via other methods. (Straight from a pixi.toml for instance). I would opt to make the Recipe struct public or perhaps we should provide a builder. WDYT?

src/recipe/parser/source.rs
```diff
@@ -0,0 +1,409 @@
...
+        let mut patches = Vec::new();
+        let mut folder = None;
+        let mut file_name = None;
+
+        for (key, value) in self.iter() {
+            match key.as_str() {
+                "url" => url = Some(value.try_convert("url")?),
+                "sha256" => {
+                    let sha256_str: String = value.try_convert("sha256")?;
+                    if sha256_str.len() != 64 {
```
wolfv, line 286: There are actually some more constraints. Maybe we could use `rattler_digest::Md5Hash` and `rattler_digest::Sha256Hash` that would parse it to the right types straight away?

wolfv:
@baszalmstra I am not sure wether we need to create `recipe`'s from scratch. It's important that we can create the `Output` struct though (and is what we do for pixi.toml in the pixi build PR).

src/recipe/parser/snapshots/rattler_build__recipe__parser__package__tests__missing_fields.snap
GrayJack: I think underlining `package` would be better as well. But by making the conversion generic over a trait (as it is right now) we lose the ability to reference.

One possible option is to change the trait definition to receive a `parent_span` as a parameter, to reference it if needed. What do you think?

src/recipe/parser/about.rs
```diff
@@ -149,40 +85,96 @@ impl About {
...
+                "repository" | "dev_url" => {
```
baszalmstra, line 102: These are aliases for `repository`. It should be an illegal to have both `repository` and `dev_url`.

baszalmstra: Actually, we should just get rid of the old crap as Wolf would say.

src/recipe/parser/source.rs
GrayJack: That is a good idea!

src/recipe/parser/about.rs
GrayJack: Removed

PR #246, opened 2023-10-30
store the original recipe and rendered recipe

wolfv:


wolfv:
I've uploaded two packages that have the potential to be "reproducible" on MacOS ARM64 computers:

https://drive.google.com/drive/folders/1Cwo6-brPT5VlawXJ84VLN3U-dabiQjAb?usp=drive_link

wolfv:
OK, I've done things a little differently now and made a "bld" folder in `/opt/bld` with the following commands:


```bash
sudo mkdir /opt/bld
sudo chown <user>:staff /opt/bld
export CONDA_BLD_PATH=/opt/bld 



```

PR #250, opened 2023-10-31
feat: no-build-id aka static build dirs

### LINKED ISSUE #248: Support static build dirs?
One not too often used feature of conda-build is that you can actually force it to use a stable build directory which is very handy for integrating tools like `ccache` and `sccache` with conda-build / rattler-build.

Here
https://github.com/prefix-dev/rattler-build/blob/c1d62baf5db9b048efcbe9fc47bc9ed124c67af8/src/metadata.rs#L76-L78

rattler build always sets the build dir to a dynamic path.

In conda-build you can force the legacy behavior by passing ` --no-build-id` which forces the build dir to not append the timestamp

swarnimarun:
Adds build option akin to conda with `--no-build-id`. 

Tasks:
- [x] Add cli flag.
- [x] Add tests?

Notes:
- Discuss, if we should provide an ENV variable as well to setup the flag?
- Figure out, how to add tests for the newly added option.

Fixes #248

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -298,6 +298,7 @@ async fn run_build_from_args(args: BuildOpts, multi_progress: MultiProgress) -> 
                 .clone()
                 .unwrap_or(vec!["conda-forge".to_string()]);
 
+            let timestamp = chrono::Utc::now();
             let output = rattler_build::metadata::Output {
                 recipe,
                 build_configuration: BuildConfiguration {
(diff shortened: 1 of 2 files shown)


wolfv:
This looks great. I actually don't know how to add a test for this either – maybe we could have a unit test for `setup_build_dir`? I have some additional requests for that function (to use the timestamp from the build configuration) that we can get to later on.

swarnimarun:
Makes sense will look into if there is a way to write tests for it properly.

wolfv:
huh, weird, looks like an unrelated test is failing now? Happy to merge this once green – and test it out here at the reproducible builds summit :)

swarnimarun:
hmm... I guess not testing the entire path makes sense

PR #252, opened 2023-11-01
feat: add cdt function

### LINKED ISSUE #91: Implement `cdt` jinja function
Is not currently implemented and should be reasonably straightforward.

swarnimarun:
TODO:
- [x] Add cdt function, as per, https://github.com/conda/conda-build/blob/e2916a2a5c590382323de97da6ac2fe6c516ff89/conda_build/jinja_context.py#L540
- [x] Add basic tests
- [ ] ~~Add end to end tests?~~

Notes:
- Discuss the approach to providing config into the set_jinja function. (currently we can't avoid clone, there are ways around it but would need some rearchitecturing)
- Made changes to a couple places where the unwrap was quite unbearable for my OCD(unwrapping while parsing doesn't seem nice).
- Figure out how to add end to end tests.

Fixes #91

### DIFF
--- a/src/recipe/jinja.rs
+++ b/src/recipe/jinja.rs
@@ -97,13 +97,13 @@ fn set_jinja(config: &SelectorConfig) -> minijinja::Environment<'static> {
             let version = Version::from_str(version).map_err(|e| {
                 minijinja::Error::new(
                     minijinja::ErrorKind::CannotDeserialize,
-                    format!("Failed to deserialize `version`: {}", e.to_string()),
+                    format!("Failed to deserialize `version`: {}", e),
                 )
             })?;
             let version_spec = VersionSpec::from_str(spec).map_err(|e| {
                 minijinja::Error::new(
                     minijinja::ErrorKind::SyntaxError,
-                    format!("Bad syntax for `spec`: {}", e.to_string()),
+                    format!("Bad syntax for `spec`: {}", e),
                 )
             })?;
             Ok(version_spec.matches(&version))
--- a/src/recipe/jinja.rs
+++ b/src/recipe/jinja.rs
@@ -121,31 +121,29 @@ fn set_jinja(config: &SelectorConfig) -> minijinja::Environment<'static> {
     } = config.clone();
     env.add_function("cdt", move |package_name: String| {
         use rattler_conda_types::Arch;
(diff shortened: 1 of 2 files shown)


src/recipe/jinja.rs
```diff
@@ -94,15 +94,72 @@ fn set_jinja() -> minijinja::Environment<'static> {
...
+            let version = Version::from_str(version).map_err(|e| {
+                minijinja::Error::new(
+                    minijinja::ErrorKind::CannotDeserialize,
+                    format!("Failed to deserialize `version`: {}", e.to_string()),
+                )
+            })?;
+            let version_spec = VersionSpec::from_str(spec).map_err(|e| {
+                minijinja::Error::new(
+                    minijinja::ErrorKind::SyntaxError,
+                    format!("Bad syntax for `spec`: {}", e.to_string()),
```
swarnimarun, line 106:
```suggestion
                    minijinja::ErrorKind::CannotDeserialize,
                    format!("Failed to deserialize `version`: {}", e),
                )
            })?;
            let version_spec = VersionSpec::from_str(spec).map_err(|e| {
                minijinja::Error::new(
                    minijinja::ErrorKind::SyntaxError,
                    format!("Bad syntax for `spec`: {}", e),
```

wolfv:
This looks very good & thanks for removing some unwraps! :)

I need to check, but we might have to also do something to make sure that the "cdt" variables are added to the "used variables" to compute the variant from.

swarnimarun:
Will take a look to see if what else is needed.

wolfv:
I just wrote a little check:


```yaml
# yaml-language-server: $schema=https://raw.githubusercontent.com/prefix-dev/recipe-format/main/schema.json

package:
  name: cdt_config
  version: "1.0.0"

requirements:
  host:
    - ${{ cdt('mypkg') }}
```

And then, when adding the following `variant_config.yaml`:

```yaml
cdt_name:
  - blabla
cdt_arch:
  - x512
```

this should change the result of the `cdt` function.

You can test with `cargo run -- build --render-only --recipe ./recipe.yaml -m ./variant_config.yaml`

I just added the necessary bits to get the used variables from the variant config (pushed to your PR!).

wolfv:
Awesome stuff!

PR #253, opened 2023-11-02
feat: add tests in rust

swarnimarun:
use command `cargo t -p rust-tests` to run the end to end tests

TODO:
- [x] PoC for tests written in rust with similar simplicity to pytests
- [x] Port atleast 1 test
- [x] Port rest of the tests from pytest
- [x] Ensure all tests are working
- [ ] Use new rust tests in CI

### DIFF
--- a/rust-tests/src/lib.rs
+++ b/rust-tests/src/lib.rs
@@ -1,12 +1,13 @@
 #[cfg(test)]
 mod tests {
+    use itertools::Itertools;
     use rattler_package_streaming::read::extract_tar_bz2;
     use std::{
         collections::HashMap,
         ffi::OsStr,
         fs::File,
         path::{Path, PathBuf},
-        process::{Command, Output},
+        process::{Command, Output, Stdio},
     };
 
     enum RattlerBuild {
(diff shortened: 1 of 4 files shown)


wolfv (approved):

wolfv:
@swarnimarun is this ready for review & merge?

rust-tests/src/lib.rs
```diff
@@ -0,0 +1,334 @@
...
+                    let p = p.as_object().unwrap();
+                    assert!(c["_path"] == p["_path"]);
+                    assert!(c["path_type"] == p["path_type"]);
+                    if p["_path"]
+                        .as_str()
+                        .map(|s| s.contains("dist-info"))
+                        .unwrap_or_default()
+                    {
+                        // TODO: figure out why this is not the same b/w expected and cmp
+                        // assert!(c["sha256"] == p["sha256"]);
```
swarnimarun, line 309: everything else is working except, can't figure out why the sha256/hash is different here. hmmm...

swarnimarun:
There is a bug with temp_dir paths, so it's best to run `cargo t -p rust-tests -- --test-threads 1`, note single test thread, means sequential tests rather than parallel tests, pytest matches this behaviour by default afaik, not sure why this sometimes bugs some of the tests but it's a gotcha.

wolfv:
Good to know. Can we set this in the Cargo.toml file as default?

rust-tests/src/lib.rs
wolfv: I can have a look!

wolfv: I think the logic was "backwards" - the distinfo files are the ones that embed the build path and are therefore not deterministic.

wolfv:
I think the tests are not yet executed in CI, correct?

PR #263, opened 2023-11-07
feat(outputs): Sort the outputs before building

GrayJack:
- [X] Basic topological sorting of the outputs
- [ ] Adapt to also account for variants

### DIFF
--- a/src/recipe/parser/requirements.rs
+++ b/src/recipe/parser/requirements.rs
@@ -15,7 +15,7 @@ use crate::{
 };
 
 /// The requirements at build- and runtime are defined in the `requirements` section of the recipe.
-#[derive(Debug, Default, Clone, Serialize, Deserialize)]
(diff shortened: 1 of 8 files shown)


src/recipe/parser/requirements.rs
```diff
@@ -65,6 +65,11 @@ impl Requirements {
         self.run_constrained.as_slice()
     }
 
+    /// Get all requirements at build time in one iterator.
+    pub fn all_build_time(&self) -> impl Iterator<Item = &Dependency> {
+        self.build.iter().chain(self.host.iter())
```
wolfv, line 70: There is an extra requirement that would be needed for all packages that are `pin_subpackage(...)` - those can appear in the run_exports as well as any other requirement sections.

wolfv:
Closing this as superseded by #278

PR #269, opened 2023-11-09
feat: git source with host git

swarnimarun:
### TODO:
- [x] Create the POC for host git.
- [x] Get the code working on unix systems at least.
- [x] Add basic unit tests from git_source to host git source.
- [x] Refactor the code, remove redundant code, remove unwraps, improve error messages and reporting.
- [x] Remove feature git, and ensure CI runs work.
- [x] Handle lack of git or git-lfs gracefully.
- [x] Write E2E test.
- [x] Windows path issues (too vague to list, out of 3 tests, last one is broken on windows, perhaps 2/3 is a good thing?)...
- [x] Fix bugs on all platforms.
- [x] Look into why conda-build performs "mirror", and if we should consider mirrors as well.

### TODO in a separate PR:
- [ ] Figure out and ensure ssh auth works by default on all platforms with all configurations.
- [ ] Use mirrored clones over normal clones.
- [ ] Write tests for all aspects (clone from git path, non-standard revs(hash, tags, branches, etc), use of cache, etc.)


swarnimarun:
This has both unit and E2E tests, but I am doing some user testing of my own, to test out all the features, will create a separate PR with more E2E tests or add more cases to the present E2E test.

PR #271, opened 2023-11-09
Also upload binaries on non-release

pavelzw:
I think it would be nice to have the binaries from non-release builds, WDYT?


wolfv:
let's do it!

PR #278, opened 2023-11-13
sort outputs when computing variant [skip ci]

wolfv:


wolfv:
I noticed one bug where the `name` is still the `jinja` and not the evaluated jinja. This was apparent in the `xtensor` recipe.

wolfv:
Couldn't recreate the issue, so merging for now! :)

PR #296, opened 2023-11-16
feat: add lfs option to git source

### LINKED ISSUE #285: Make `lfs` requirement specific?
Github Actions (checkout action) has a extra config field to ask for lfs or not.

E.g. 

```yaml
source:
   git_url: https://blalba
   lfs: true  # (defaults to false)
```

swarnimarun:
Unit tests should already handle it, but should we add a lfs test repo for end-to-end test?

Fixes #285

### DIFF
--- a/src/source/host_git_source.rs
+++ b/src/source/host_git_source.rs
@@ -179,7 +179,7 @@ pub fn git_src(
         return Err(SourceError::GitErrorStr("failed to git reset"));
     }
 
-    if git_lfs_required(&cache_path) {
+    if source.lfs() || git_lfs_required(&cache_path) {
         // only do lfs pull if needed!
         git_lfs_pull()?;
     }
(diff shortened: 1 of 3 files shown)


src/source/host_git_source.rs
```diff
@@ -179,7 +179,7 @@ pub fn git_src(
         return Err(SourceError::GitErrorStr("failed to git reset"));
     }
 
-    if git_lfs_required(&cache_path) {
+    if source.lfs() || git_lfs_required(&cache_path) {
```
wolfv, line 182: Shouldn't this be a `&&`? I think the explicit setting should be required to enable LFS and we could even skip the lfs required check (or run it just to save some extra invocations if it's not actually required).

wolfv (changes requested):

wolfv:
LGTM except for the semantics of this option.

src/source/host_git_source.rs
swarnimarun: makes sense

wolfv (approved):

PR #300, opened 2023-11-16
fix: tests for windows

swarnimarun:


src/windows/env.rs
```diff
@@ -8,6 +8,7 @@ pub fn default_env_vars(prefix: &Path, target_platform: &Platform) -> HashMap<St
         Platform::Win64 => "amd64",
         // TODO: Is this correct?
         Platform::WinArm64 => "arm64",
+        Platform::NoArch => "noarch",
```
wolfv, line 11: not sure this is correct. I think we should rather use `host_platform` where we are calling this function (that variable is guaranteed to not be noarch).

PS. my other comment about using `.arch()` was also wrong, that would not work.

swarnimarun: ack. It's provided target_platform here, btw, https://github.com/prefix-dev/rattler-build/blob/6c8f500d5000b94953ddd7b64a779f94d8e9274b/src/env_vars.rs#L353

PR #340, opened 2023-11-22
fix: error if git rev and git depth are used together

swarnimarun:


wolfv:
LGTM as discussed on Discord, might be good to add a check also when fetching the git source, so that GitSrc created from other places (e.g. future Python bindings) also error out properly.

src/source/git_source.rs
```diff
@@ -49,11 +49,27 @@ pub fn git_src(
...
+        .success()
+    {
+        return Err(SourceError::GitErrorStr(
+            "`git` command not found in `PATH`",
+        ));
+    }
+
+    if (source.rev().is_empty() || source.rev().eq("HEAD"))
+        // depth == -1, fetches the entire git history
+        && source.depth().map(|s| s == -1).unwrap_or_default()
```
swarnimarun, line 67:
```suggestion
        && source.depth().map(|s| s != -1).unwrap_or_default()
```

wolfv (approved):

PR #358, opened 2023-11-26
Add use_gitignore to recipe

### LINKED ISSUE #357: Add `use-gitignore` option to recipes
It would be nice to have the ability to also copy things that are ignored by `.gitignore`. This is helpful if you build your packages using backends like `setuptools-scm`.

pavelzw:
Fixes #357


wolfv:
Looks good to me. Do you have time to add this to the docs as well?

Just as a note to self:

- [ ] update CEP
- [ ] update recipe-format schema

pavelzw:
I won't have time for creating docs before this evening. If the next release is planned before that, I would suggest splitting the PR into two parts.
Alternatively, feel free to commit to my branch, otherwise I will do it this evening.

Issue #307, opened 2023-11-17
Implement multiple tests and new test format


wolfv:
We are working on having a new test section that supports additional features:

- multiple, independent tests
- different test types that are more powerful vs. the current test types

The current ideas for the new test section are described here: https://github.com/conda-incubator/ceps/blob/18ce1875c4758347d83e913012ec48fb59e381df/cep-20.2.md#test-section

Additionally, we have some ideas for a new test type to test the existence of files. This is described in the following document: https://github.com/conda-incubator/ceps/blob/4a0a540c237c88eedc64be53f7bf87e44b50a087/cep-20.3.md

PR #359, opened 2023-11-27
feat: package content tests 

swarnimarun:
TODO:
- [x] Parsing
- [x] Basic implementation for tests
- [ ] site_packages scanning (not sure for which python installation)
- [ ] add e2e tests
- [ ] perhaps? provide cli flags to just run the package contents test or to skip it?

fix https://github.com/prefix-dev/rattler-build/issues/307

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -283,6 +283,10 @@ pub async fn run_build(
         )
         .await
         .into_diagnostic()?;
+
+        test::run_package_content_tests(output.recipe.test().package_content())
+            .await
+            .into_diagnostic()?;
     }
 
     if !tool_configuration.no_clean {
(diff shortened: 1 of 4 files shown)


wolfv:
Cool stuff!

wolfv:
For site-packages we could just look for `**/site-packages/...` for now (or for `noarch: python` it's actually in the root at `/site-packages/...`.

We could also extract the python version that is encoded in the run dependencies, but we can also fix that later.

wolfv:
Also I don't think we need a CLI flag for now. 

In a follow up I would like to split up the tests into multiple tests (a list of tests) and then run them each individually. This new test section would be a possible enum element of a "test case".

PR #361, opened 2023-11-27
Add docs for use_gitignore and lfs

pavelzw:


docs/recipe_file.md
```diff
@@ -184,16 +184,21 @@ source:
...
   git_rev: 1.1.4
   git_depth: -1 # Defaults to -1/not shallow
+  lfs: true # defaults to false
+  use_gitignore: true # defaults to false
 ```
 
-Note: `git_rev` may not be available within commit depth range, consider avoiding use of both simultaneously.  
+Note: `git_rev` may not be available within commit depth range, consider avoiding use of both simultaneously.
+
+If you specify `use_gitignore: true`, then only the files that are ignored by git will be copied to the work directory after the clone.
```
pavelzw, line 200: Does this use case even make sense? The only thing that I can think of where this makes a difference is when your `.gitignore` has `.git` in it and then it won't be copied... Otherwise all `.gitignore` files are not copied anyway since they are not tracked by git...

pavelzw: How about we only support `use_gitignore` for `PathSource`? @wolfv wdyt?

PR #363, opened 2023-11-27
fix: improve jinja err msg

### LINKED ISSUE #309: Give nice error messages if `if-condition` Jinja doesn't parse properly
```yaml
requirements:
  build:
    - ${{ compiler('cxx') }}
    - cmake
    - if: unix =123=32 213=123
      then: make
```

We try to extract "used variables" from the if condition and Jinja parsing into an AST fails. 
Our error message is not great like the others, though.

swarnimarun:
Before:
image

After:
image

Note: currently I just pass the entire recipe source to the function. I am considering forwarding it to all the functions, as otherwise, the string that will be rendered won't make much sense.

Fixes #309

### DIFF
--- a/src/variant_config.rs
+++ b/src/variant_config.rs
@@ -380,7 +380,7 @@ impl VariantConfig {
         for output in outputs.iter() {
             // for the topological sort we only take into account `pin_subpackage` expressions
             // in the recipe which are captured by the `used vars`
-            let used_vars = used_vars_from_expressions(output)?;
+            let used_vars = used_vars_from_expressions(output, recipe)?;
             let parsed_recipe = Recipe::from_node(output, selector_config.clone())
                 .map_err(|err| ParsingError::from_partial(recipe, err))?;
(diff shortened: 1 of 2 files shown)


wolfv:
much better!

wolfv (approved):

wolfv:
Should we delete the commented out error? Happy to merge.

src/variant_config.rs
```diff
@@ -755,9 +755,8 @@ pub enum VariantError {
 
     #[error("Found a cycle in the recipe outputs: {0}")]
     CycleInRecipeOutputs(String),
-
-    #[error("Could not parse a Jinja expression: {0}")]
-    JinjaParseError(#[from] minijinja::Error),
+    // #[error("Could not parse a Jinja expression: {0}")]
+    // JinjaParseError(#[from] minijinja::Error),
```
swarnimarun, line 759:
```suggestion
```

PR #367, opened 2023-11-29
Fix jinja string interp instructions

JarrettSJohnson:
Explains using the new jinja string interp syntax for `recipe.yaml` as described in https://github.com/conda-incubator/ceps/pull/54


wolfv:
Thanks!

PR #368, opened 2023-11-29
implement pin-compatible

wolfv:
- [ ] figure out if we only want the resolved packages from the "previous" environment or all of them
- [ ] do we want transitive dependencies in the compatible packages (that can be pinned?)


wolfv:
I double checked conda-build's code and it looks to me as if `pin_compatible` only makes sense in the `run` section of the recipe?

```py
def pin_compatible(
    m,
    package_name,
    lower_bound=None,
    upper_bound=None,
    min_pin="x.x.x.x.x.x",
    max_pin="x",
    permit_undefined_jinja=False,
    exact=False,
    bypass_env_check=False,
):
    """dynamically pin based on currently installed version.

    only mandatory input is package_name.
    upper_bound is the authoritative upper bound, if provided.  The lower bound is the the
        currently installed version.
    pin expressions are of the form 'x.x' - the number of pins is the number of x's separated
        by ``.``.
    """
    global cached_env_dependencies
    compatibility = ""

    # optimization: this is slow (requires solver), so better to bypass it
    # until the finalization stage.
    if not bypass_env_check and not permit_undefined_jinja:
        # this is the version split up into its component bits.
        # There are two cases considered here (so far):
        # 1. Good packages that follow semver style (if not philosophy).  For example, 1.2.3
        # 2. Evil packages that cram everything alongside a single major version.  For example, 9b
        key = (m.name(), HashableDict(m.config.variant))
        if key in cached_env_dependencies:
            pins = cached_env_dependencies[key]
        else:
            if m.is_cross and not m.build_is_host:
                pins, _, _ = get_env_dependencies(m, "host", m.config.variant)
            else:
                pins, _, _ = get_env_dependencies(m, "build", m.config.variant)
                if m.build_is_host:
                    host_pins, _, _ = get_env_dependencies(m, "host", m.config.variant)
                    pins.extend(host_pins)
            cached_env_dependencies[key] = pins
        versions = {p.split(" ")[0]: p.split(" ")[1:] for p in pins}
        if versions:
            if exact and versions.get(package_name):
                compatibility = " ".join(versions[package_name])
            else:
                version = lower_bound or versions.get(package_name)
                if version:
                    if hasattr(version, "__iter__") and not isinstance(version, str):
                        version = version[0]
                    else:
                        version = str(version)
                    if upper_bound:
                        if min_pin or lower_bound:
                            compatibility = ">=" + str(version) + ","
                        compatibility += f"<{upper_bound}"
                    else:
                        compatibility = apply_pin_expressions(version, min_pin, max_pin)

    if not compatibility and not permit_undefined_jinja and not bypass_env_check:
        check = re.compile(r"pin_compatible\s*\(\s*[" '"]{}[' '"]'.format(package_name))
        if check.search(m.extract_requirements_text()):
            raise RuntimeError(
                "Could not get compatibility information for {} package.  "
                "Is it one of your host dependencies?".format(package_name)
            )
    return (
        " ".join((package_name, compatibility))
        if compatibility is not None
        else package_name
    )
```

wolfv:
It turns out that `m.is_cross` is true when there is a `host` section in the recipe ... I think our logic is fine for the moment.

Issue #255, opened 2023-11-02
Keep only needed `unwrap` statements


tdejager:
While looking through the code I've noticed a whole bunch of `unwrap` statements. Would be good to get a close look into which of these could be replaced.

See:

```bash
$ rg unwrap -c

src/used_variables.rs:3
src/windows/link.rs:1
src/main.rs:10
src/windows/env.rs:2
src/recipe.rs:3
src/packaging.rs:11
src/post.rs:3
src/index.rs:25
src/variant_config.rs:14
src/unix/env.rs:2
src/linux/link.rs:1
src/linux/env.rs:8
src/metadata.rs:2
src/render/pin.rs:15
src/render/resolved_dependencies.rs:5
src/recipe/parser.rs:3
src/source/url_source.rs:6
src/recipe/stage1.rs:5
src/recipe/jinja.rs:3
src/recipe/error.rs:3
src/render/solver.rs:13
src/recipe/parser/build.rs:3
src/recipe/parser/about.rs:2
src/source/git_source.rs:13
src/source/mod.rs:16
src/recipe/parser/package.rs:2
src/macos/link.rs:1
src/recipe/parser/source.rs:1
src/test.rs:20
src/selectors.rs:19
src/env_vars.rs:7
```

swarnimarun:
I have reviewed most of the code as well recently, can say largely most of it can be removed, some more easily than others, perhaps anyone making a PR consider doing it one file at a time.

wolfv:
I am also happy to take on some of these, maybe during the next train ride or so. Will ping here which files I am tackling to avoid double work.

wolfv:
I cleaned a few. Just to note, some are also `unwrap_or_default` and similar (you might want to use `unwrap()` as search).

PR #370, opened 2023-11-29
fix: remove all the unwraps from production code

swarnimarun:
4 of the unwraps are now expected as they can't be fixed trivially, i.e. require major refactoring and are unreachable(to my knowledge) and tests have been untouched.

Fixes https://github.com/prefix-dev/rattler-build/issues/255

### DIFF
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -91,7 +91,12 @@ impl Directories {
 
         let build_dir = setup_build_dir(&output_dir, name, no_build_id, timestamp)
             .expect("Could not create build directory");
-        let recipe_dir = recipe_path.parent().unwrap().to_path_buf();
+        let recipe_dir = recipe_path
+            .parent()
+            .ok_or_else(|| {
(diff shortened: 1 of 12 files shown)


src/metadata.rs
```diff
@@ -240,7 +245,7 @@ impl Display for Output {
             "\nOutput: {}-{}-{}\n",
             self.name().as_normalized(),
             self.version(),
-            self.build_string()
+            self.build_string().unwrap_or("{build-string-not-set}")
```
swarnimarun, line 248: This is done because Rust has string handling issues, and if you try to println Display then it ignores `fmt::Result` which means a crash from println, and as people throw println in all sorts of places that can cause unexpected crashes, I can't think of a better solution without significant refactoring.

A proper refactor would have two sets of types, the first for parsing YAML to.
The second set is built from the first set with all the hashing, and finalization done, which are then just used for operations.
Hence failing to get finalized dependencies and setting hash would be impossible, and wouldn't require explicit code to handle.

wolfv (approved):
Excellent

PR #389, opened 2023-12-05
fix: add docs for package-contents test

swarnimarun:
I think we should improve the comments and explanations here in general, any suggestions or guidelines I should follow while doing that?


wolfv (approved):

wolfv:
We don't really have any guidelines. Improving docs is of course always welcome!

PR #400, opened 2023-12-06
refactor: add python object

baszalmstra:
Adds the `python` object and moves `entry_points` in there. This PR aligns with the CEP.

### DIFF
--- a/src/recipe/parser/build.rs
+++ b/src/recipe/parser/build.rs
@@ -31,9 +31,9 @@ pub struct Build {
     pub(super) script: Script,
     /// A noarch package runs on any platform. It can be either a python package or a generic package.
     pub(super) noarch: NoArchType,
-    /// For a Python noarch package to have executables it is necessary to specify the python entry points.
-    /// These contain the name of the executable and the module + function that should be executed.
-    pub(super) entry_points: Vec<EntryPoint>,
+    /// Python specific build configuration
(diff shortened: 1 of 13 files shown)


src/recipe/parser/build.rs
```diff
@@ -104,16 +104,16 @@ impl TryConvertNode<Build> for RenderedMappingNode {
...
                     build.noarch = value.try_convert(key_str)?;
                 }
-                "entry_points" => {
+                "python" => {
                     if let Some(NoArchKind::Generic) = build.noarch.kind() {
                         return Err(_partialerror!(
                             *key.span(),
                             ErrorKind::Other,
-                            label = "`entry_points` are only allowed for `python` noarch packages"
+                            label = "`python` is only allowed for `python` noarch packages"
```
wolfv, line 112: I think this is not true. Even for `entry_points` it was already wrong (we added entry_points to the `rerun` recipe and it's not noarch).

Some of the other options (`skip_pyc_compilation` are also in use for non-noarch packages I believe).

wolfv:
looks great just one little comment

src/recipe/parser/build.rs
baszalmstra: Ok! I removed it.

wolfv (approved):

PR #405, opened 2023-12-07
refactor: serde Dependency instead of string magic

baszalmstra:
Instead of serializing dependencies as magic strings (e.g. `__COMPILER c`) in the rendered recipe we render them using enum types instead. 

Turns:

```yaml
- __COMPILER c
- make
- perl
```

into 

```yaml
- compiler: c
- make
- perl
```


wolfv:
Nice!

PR #407, opened 2023-12-07
feat: error collect as vec

swarnimarun:
We now show most of the errors other than yaml parsing itself at once.
image


swarnimarun:
Currently each unit of MappingNode, needs to implement error collection, until then we will stop at the first error in each MappingNode.

swarnimarun:
I think I will create issues for it, we can slowly implement it for each mapping node.

wolfv (approved):

PR #408, opened 2023-12-08
Refactor CI

pavelzw:
This is an attempt to make the CI pipeline easier to read.

Notable changes:
- hardcode all non-changing things (like `PROJECT_NAME`) for easier readability
- use more github jinja instead of bash variables since they are evaluated directly when looking at the CI logs
- use rust cache for compiling
- don't strip debug information (they should already not be included when compiling with `--release`, right?)
- put `fmt`, `clippy` in the same workflow
- add `aarch64-unknown-linux-musl`
- switch from `x86_64-unknown-linux-gnu` to `x86_64-unknown-linux-musl` (as in pixi)
- move away from deprecated `actions-rs/*` actions


pavelzw:
Any ideas on why the aarch-unknown-linux-musl test is failing? Seems to be failing here:

https://github.com/prefix-dev/rattler-build/blob/d8631714a0bfd910f6cd925c195ba0ab0946dde6/src/source/git_source.rs#L292-L311

Maybe the `.git` directory is not mounted correctly in the cross container?

pavelzw:
Didn't manage to reproduce the failing test locally, let's keep aarch64 tests deactivated for now.

wolfv:
thanks!

Issue #410, opened 2023-12-09
rattler-build 0.6.0 breaks cross-python activation scripts


pavelzw:
```yml
# yaml-language-server: $schema=https://raw.githubusercontent.com/prefix-dev/recipe-format/main/schema.json

package:
  name: test-package
  version: '0.1.0'
build:
  script:
    - ls -lah
requirements:
  build:
    - if: build_platform != target_platform
      then:
        - python
        - cross-python_${{ target_platform }}
```

```yml
on:
  push

jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - uses: mamba-org/setup-micromamba@v1
        with:
          environment-name: test
          create-args: rattler-build=0.6.0
      - run: rattler-build build -r conda.recipe/recipe.yaml --target-platform=linux-aarch64 --no-test
        shell: bash -el {0}
```

I get the following error:

```

Resolving for environment specs:
 - python
 - cross-python_linux-aarch64
...

┌───────────────────────────────────────────┬────────────┬────────────────────┬─────────────┬────────────┐
│ Package                                   ┆ Version    ┆ Build              ┆ Channel     ┆ Size       │
╞═══════════════════════════════════════════╪════════════╪════════════════════╪═════════════╪════════════╡
│ cross-python_linux-aarch64                ┆ 3.12       ┆ 41_cpython         ┆ conda-forge ┆ 12.01 KiB  │
│ python                                    ┆ 3.12.0     ┆ hab00c5b_0_cpython ┆ conda-forge ┆ 30.64 MiB  │
│ xz                                        ┆ 5.2.6      ┆ h166bdaf_0         ┆ conda-forge ┆ 408.56 KiB │
│ readline                                  ┆ 8.2        ┆ h8228510_1         ┆ conda-forge ┆ 274.86 KiB │
│ libuuid                                   ┆ 2.38.1     ┆ h0b41bf4_0         ┆ conda-forge ┆ 32.81 KiB  │
│ libexpat                                  ┆ 2.5.0      ┆ hcb278e6_1         ┆ conda-forge ┆ 76.15 KiB  │
│ libzlib                                   ┆ 1.2.13     ┆ hd590300_5         ┆ conda-forge ┆ 60.14 KiB  │
│ libnsl                                    ┆ 2.0.1      ┆ hd590300_0         ┆ conda-forge ┆ 32.62 KiB  │
[...]

PR #411, opened 2023-12-12
use the proper platform when resolving

wolfv:


wolfv:
this fixes #410 and makes cross-compilation work properly (for the first time? 😱)

PR #413, opened 2023-12-12
Add auth_file option to read authentication information for repositories from a file

0xbe7a:
This PR adds the option to disable the usage of the system keychain for storing login information but instead to rely soly on the fallback auth store, which is a simple JSON File.

Depends on https://github.com/mamba-org/rattler/pull/435


0xbe7a:
@wolfv, I have customized the PR again and there is now only the option to specify a path to a file containing the authentication information. Should we leave the option hidden?

pavelzw:
If you keep the information hidden, please add a section for it in the docs.

pavelzw:
I also did this with setup-pixi but with my personal account.

wolfv:
great stuff! I can make such an account on prefix, of course. Any package that you think would work good?

0xbe7a:
Just any dummy package we can list as a build dependency.  Does not require any content. e.g. we used this recipe for conda-forge https://github.com/conda-forge/_python_rc-feedstock/blob/main/recipe/meta.yaml

wolfv:
I added you two to this (channel)[https://prefix.dev/channels/rattler-auth-test] and uploaded `_libgcc_mutex` (linux-64 only!).

0xbe7a:
I want to include a read-only key for this channel within the test so that everyone can run the full integration tests. With prefix-dev I can only include one unrestricted API key for my entire account and all channels, right?

wolfv:
yeah that's true. how about I add one as a secret to the repo for now and we do what you suggest once we have implemented this kind of scoped access on prefix?

0xbe7a:
👍 Then just add the API key of a user who has read access to the channel as a secret and environment variable, and I'll just skip the test if the environment variable is not set.

wolfv:
Added the following secret: `PREFIX_DEV_READ_ONLY_TOKEN`

0xbe7a:
@wolfv, I added an integration test to test authentication in a private prefix.dev channel and tested it locally and it works. The CI is currently failing as the secret is not exposed to outside contributors running.

I think you need to make an empty commit to this PR for the secrets to get exposed to the workflow

.github/workflows/end-to-end.yml
pavelzw, line 1: Can you add macos-latest too?

test/end-to-end/test_simple.py
```diff
@@ -169,3 +175,31 @@ def test_pkg_hash(rattler_build: RattlerBuild, recipes: Path, tmp_path: Path):
     pkg = get_package(tmp_path, "pkg_hash")
     expected_hash = variant_hash({"target_platform": host_subdir()})
     assert pkg.name.endswith(f"pkg_hash-1.0.0-{expected_hash}_my_pkg.tar.bz2")
+
+@pytest.mark.skipif("PREFIX_DEV_READ_ONLY_TOKEN" not in os.environ, reason="requires PREFIX_DEV_READ_ONLY_TOKEN")
```
pavelzw, line 179:
```suggestion
@pytest.mark.skipif(not os.env.get("PREFIX_DEV_READ_ONLY_TOKEN", ""), reason="requires PREFIX_DEV_READ_ONLY_TOKEN")
```

the test is currently still being executed, maybe because the env variable is the empty string?

0xbe7a:
The test currently fails on main on macos because the workflow was never executed with the secret exposed and therefore always skipped (see my last comment)

0xbe7a:
The test currently fails on main, but works on my fork (https://github.com/0xbe7a/rattler-build/actions/runs/7246636641/job/19738988493). Can you run the test again to make sure it's not a temporary issue, @wolfv?

PR #418, opened 2023-12-13
[🚧] new multi tests

wolfv:
- [x] decide on adding `pip_check: True / False` to `PythonTest`.
- [ ] Reuse previous environments, ignoring `pip`
- [x] Decide on serialization format, should we keep `run_test.py` or serialize to a JSON?
- [ ] For the Commands test, should we serialize (potential) if selectors so that people can have optional dependencies based on their test platform? E.g. 
```yaml
requirements:
  build:
     - if: build_platform == linux-64
       then:
         - arm-emulator
```

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -118,8 +118,8 @@ requirements:
   run_constrained:
     - xsimd >=8.0.3,<10
 
-test:
-  commands:
+tests:
+  - script:
     - if: unix or emscripten
       then:
         - test -d ${PREFIX}/include/xtensor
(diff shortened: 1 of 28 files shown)


wolfv:
This is how we are currently serializing the 3 different test types:

`PythonTest`:

- run_test.py (this contains the imports that are listed as `import bla...`, one per line)

`CommandsTest`:

- `test_time_dependencies.json that contains a dictionary with `run` and `build` keys
- `run_test.sh` / `run_test.bat`. If it's an arch'd package, either the `.sh` or `.bat`. If it is a `noarch` package we write both files.

`DownstreamTest`:

- `downstream_test.json`: `{ "downstream": "package_name" }`.

wolfv:
The Python tests are now serialized to a `python_test.json` file instead.

Issue #404, opened 2023-12-06
Version can't be determined from a tag? 


mariusvniekerk:
Given this minimal recipe

```yaml
# ./recipe/foo/recipe.yaml
context:
  version: ${{ env['GIT_DESCRIBE_TAG'] | replace('v', '') }}${{ 'dev' if env['GIT_DESCRIBE_NUMBER'] == 0 }}

package:
  version: "${{ version }}"
  name: foo
  
source:
  git: "../../"
```

rattler-build will fail with

```
o target platform specified, using current platform
Error:   × Parsing: missing field `version`
   ╭─[4:1]
 4 │     package:
 5 │ ╭─▶   version: "${{ version }}"
 6 │ │     name: foo
 7 │ │     
 8 │ ├─▶ source:
   · ╰──── here
 9 │       git: "../../"
   ╰────
  help: the field `version` is required for `package`
```

conda-build works for this (boa also fails)

swarnimarun:
This is largely cause we don't have these [docs/source/user-guide/environment-variables.rst](https://github.com/conda/conda-build/blob/ca9de6013cf7d6909f063247512546d351f7a757/docs/source/user-guide/environment-variables.rst#git-environment-variables) defined.

Created an issue for them, and will add the source env variables as they seem to be quite useful.


But otherwise you should be writing.
```jinja
${{ env.get('GIT_DESCRIBE_TAG') }}${{ 'dev' if env.get('GIT_DESCRIBE_NUMBER') == 0 }}
```

As mentioned in the docs. We don't have indexing like conda-build. Also note `get` will error out if the variable is not found at build time. Hence, if you are fine with empty values use `get_default` to explicitly as for it.

[docs for jinja env.get pin expr](https://prefix-dev.github.io/rattler-build/recipe_file.html?highlight=env.get#the-env-jinja-functions)

PR #423, opened 2023-12-15
feat: add git jinja functions & experimental flag

swarnimarun:
- This PR add `git` functions for getting `latest_tag_rev`, `latest_tag_hash` and latest `head_hash`.
This should simplify possible uses cases for automating git tag fetching.

- Futhermore, this PR adds `--experimental` flag to commonopts, aka both rebuild and build commands. And makes this feature experimental.
We provide an experimental feature error if the feature is used without experimental enabled. And experimental features can also be enabled with env flag, `RATTLER_BUILD_EXPERIMENTAL`. 

Aside:
The only issue with this approach is `--sort` requires git from 2018 that is 2.18 specifically, currently not sure if it worth implementing something manually. Do we want to support older git versions?

Fixes https://github.com/prefix-dev/rattler-build/issues/404


wolfv:
I think this is quite cool. Maybe we should ask @mariusvniekerk or @pavelzw what they think.

The idea for the usage is something like:

```
${{ git.latest_tag("https://github.com/my/repo") }} -> 0.8.0
${{ git.latest_tag_hash("https://github.com/my/repo") }} -> 1234rev
```

You could add these to the `context` or use them directly, and context interpolation should als o work, e.g.:


```
context:
   git_url: https://github.com/my/repo
   git_tag: ${{ git.latest_tag(git_url) }}
   short_version: ${{ git_tag.split('.')[0] }}

source:
  git: ${{ git_url }}
  tag: ${{ git_tag }}
```

This is a bit less magic and more "imperative" vs. what conda build is currently doing.

@swarnimarun regarding the naming of `latest_tag_hash`, shouldn't we call it `latest_rev` (or `ref` I always mix them up)?

pavelzw:
I think this sounds nice :) 
One thing to look out for is if you are only doing a shallow clone. (This is what `actions/checkout` does by default. There you specifically need to use `with: fetch-depth: 0`)

Another thing that conda-Build is implementing is GIT_DESCRIBE_NUMBER (the number of commits after the last tag) and GIT_DESCRIBE_HASH (this is the suffix of the output of `git describe`). Not sure about the relevance of this feature, but this is needed when trying to replicate what setuptools-scm is doing.

PR #425, opened 2023-12-18
fix: ensure zips without root are handled

### LINKED ISSUE #316: I don't think it unzips .zip files
When I point it at .tar.gz it unzips them into work folder, but with .zip files not so much 🤷

swarnimarun:
Zips don't need to have a root file, unlike normal tar even if they can be unzipped from tar extractor. So unconditional strip components for zip files it will break "most but not all" zip files. :3

Fixes #316


wolfv (approved):

wolfv:
Nice, thanks!

PR #429, opened 2023-12-18
Add quetz upload command

0xbe7a:


0xbe7a:
@wolfv I've only implemented Quetz uploading for now, but in a extensible way so that we can add additional integrations with ease later.

wolfv (approved):

PR #439, opened 2023-12-20
feat: artifactory upload

0xbe7a:
This also includes a small change to the upload CLI interface, as `package_name` is now a mandatory field and no longer an option.

### DIFF
--- a/src/upload.rs
+++ b/src/upload.rs
@@ -1,6 +1,7 @@
 use std::path::PathBuf;
 
-use miette::IntoDiagnostic;
+use miette::{Context, IntoDiagnostic};
+use rattler_conda_types::package::{IndexJson, PackageFile};
 use rattler_networking::{redact_known_secrets_from_error, Authentication, AuthenticationStorage};
 use reqwest::Method;
 use sha2::Digest;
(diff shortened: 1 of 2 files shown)


src/upload.rs
```diff
@@ -70,3 +71,78 @@ pub async fn upload_package_to_quetz(
...
+
+pub async fn upload_package_to_artifactory(
+    storage: &AuthenticationStorage,
+    username: Option<String>,
+    password: Option<String>,
+    package_file: PathBuf,
+    url: Url,
+    channel: String,
+) -> miette::Result<()> {
+    let package_dir = tempfile::tempdir().unwrap();
```
wolfv, line 83: Use `?` instead of `unwrap`?

wolfv (approved):

wolfv:
What do you think about making `package_file` `global` so that it can be set anywhere? I am going to merge this one and rebase my prefix upload PR. 

Could be nice to add the progress bars to all of the uploaders

pavelzw:
I'm not sure the `package_file` flag makes sense for `rattler-build build` since it can produce multiple packages when you specify variant files.

wolfv:
@pavelzw nope. But if we declare the option as `global` it will be attached to all of the subcommands (but not propagated up).

Issue #211, opened 2023-10-18
Write out only unique dependencies


wolfv:
Currently it can happen that we store the (exact) same MatchSpec more than once (e.g. once from run_exports, and the other time from the recipe). 
This actually happens in the `toml` test (we end up with `python` 2x in the run dependencies in `index.json`).

We should filter for unique matchspecs before writing the metadata.

hadim:
I gave it a try at https://github.com/prefix-dev/rattler-build/pull/441

PR #441, opened 2023-12-21
Fix deduplication issue in create_index_json function

hadim:
Fix https://github.com/prefix-dev/rattler-build/issues/211

It does not seem much more complicated than that.


wolfv:
Nice! Didn't know about this little function! 
Maybe add it to `constrains` as well?

hadim:
> Nice! Didn't know about this little function!

It's crazy the amount of little functions like that, that make rust super efficient (almost Python-like sometime).

> Maybe add it to constrains as well?

Done.

PR #442, opened 2023-12-21
[skip ci] use hand-rolled install-name-tool replacement in rust

wolfv:
Some experiments to replace calling `install_name_tool` with Rust code. 

I've tried this before, but this time it seems to work?! Needs more testing :)


wolfv:
superseded by #501

PR #447, opened 2023-12-25
fix extraction for zip files

### LINKED ISSUE #446: zip files are still sort of broken 2
Follow-up to #427 

When doing the same reproducer as in #427, we can see that `ls -lah` outputs the following

```
+ ls -lah
total 8800
drwxr-xr-x  13 pavel  staff   416B Dec 25 15:20 .
drwxr-xr-x   5 pavel  staff   160B Dec 25 15:20 ..
-rw-r--r--   1 pavel  staff   505B Dec 25 15:20 ._snowflake-jdbc-3.13.27.jar
-rw-r--r--   1 pavel  staff   495B Dec 25 15:20 ._snowflake-license.txt
```

It should be

```
+ ls -lah
total 8800
drwxr-xr-x  13 pavel  staff   416B Dec 25 15:20 .
drwxr-xr-x   5 pavel  staff   160B Dec 25 15:20 ..
-rw-r--r--   1 pavel  staff   505B Dec 25 15:20 snowflake-jdbc-3.13.27.jar
-rw-r--r--   1 pavel  staff   495B Dec 25 15:20 snowflake-license.txt
```

Tested on macos-arm64, macos-64, linux-64, they all behave the same in this regard.

0xbe7a:
Closes #446.

We now first extract the zip/tar files into a tempdir and then move all files to the target directory, deleting the root_dir if one exists.

### DIFF
--- a/rust-tests/src/lib.rs
+++ b/rust-tests/src/lib.rs
@@ -126,7 +126,7 @@ mod tests {
     }
 
     /// doesn't correctly handle spaces within argument of args escape all spaces
-    fn shx<'a>(src: impl AsRef<str>) -> Option<String> {
+    fn shx(src: impl AsRef<str>) -> Option<String> {
         let (prog, args) = src.as_ref().split_once(' ')?;
         Command::new(prog)
             .args(args.split(' '))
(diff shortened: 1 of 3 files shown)


wolfv (approved):
Awesome, thanks!

PR #449, opened 2024-01-01
feat: progressbars for downloading and extracting

swarnimarun:
TODO:
- [x] discuss prefix and labels.
- [x] Add tests.


wolfv:
I just tried it and didn't see a progress bar when downloading the source from a URL source. Is that working for you?

src/source/mod.rs
```diff
@@ -314,6 +360,68 @@ fn extract_zip(archive: &Path, target_directory: &Path) -> Result<(), SourceErro
...
+        let tempdir = tempfile::tempdir().unwrap();
+        let file_path = tempdir.path().join("test.zip");
+        let mut file = File::create(&file_path).unwrap();
+        _ = file.write_all(HELLOW_ZIP_FILE);
+
+        let res = extract_zip(file_path, tempdir.path(), multi_progress.clone());
+        assert!(term.contents().trim().starts_with(
+            "Extracting zip       [00:00:00] [━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━]"
+        ));
+        assert!(matches!(res.err(), None));
```
swarnimarun, line 403:
```suggestion
        assert!(res.err().is_none());
```

src/source/url_source.rs
```diff
@@ -118,17 +126,53 @@ pub(crate) async fn url_src(source: &UrlSource, cache_dir: &Path) -> Result<Path
...
+        let resp = client.head(source.url().as_str()).send().await?;
+        if resp.status().is_success() {
+            resp.headers() // Gives is the HeaderMap
+                .get(reqwest::header::CONTENT_LENGTH) // Gives us an Option containing the HeaderValue
+                .and_then(|ct_len| ct_len.to_str().ok()) // Unwraps the Option as &str
+                .and_then(|ct_len| ct_len.parse().ok()) // Parses the Option as u64
+                .unwrap_or(0) // Fallback to 0
+        } else {
+            // We return an Error if something goes wrong here
+            return Err(SourceError::UrlNotFile(source.url().clone()));
```
swarnimarun, line 140:
```suggestion
            resp.headers()
                .get(reqwest::header::CONTENT_LENGTH)
                .and_then(|ct_len| ct_len.to_str().ok())
                .and_then(|ct_len| ct_len.parse().ok())
                .unwrap_or(0)
        } else {
            return Err(SourceError::UrlNotFile(source.url().clone()));
```
redundant comments

src/source/mod.rs
wolfv: you can also just `unwrap` in the test and it should fail the test if there is an error.

Issue #433, opened 2023-12-19
improve shell completion DX


pavelzw:
https://github.com/prefix-dev/rattler-build/pull/426#issuecomment-1863059761

PR #451, opened 2024-01-01
fix: improve completions dx

swarnimarun:
TODO:
- [ ] add tests...

### DIFF
--- a/docs/cli_usage.md
+++ b/docs/cli_usage.md
@@ -3,11 +3,11 @@
 ## Shell Completions
 
 We support shell completions through clap-complete.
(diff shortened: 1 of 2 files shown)


pavelzw:
`rattler-build upload <TAB>` still looks a bit odd... 
image
`prefix` and `quetz` are in the same line

pavelzw:
Closes #433

pavelzw:
Also, `pixi completion` automatically detects your shell and outputs the corresponding shell completion. I don't really have strong opinions on whether we would want shell detection or not, I personally am fine with both 🤷🏻

wolfv (approved):

PR #452, opened 2024-01-02
[skip ci] fix: improve tests performance with custom harness

swarnimarun:
Cleaned up the code, and added proper multi-threading.

Locally this is around 50% faster(from ~30s to ~15s). And provides better overall control for running tests, hopefully multi-threading doesn't cause bugs here, from my testing I haven't had any issues with it locally.


wolfv:
Do you know what is going on here? Would be happy to merge when the tests pass.

swarnimarun:
I have not being able to reproduce the issue on any windows systems or servers, so I think this is very specific to the windows server for github actions. Will separately debug it on my own repo within the action itself, until then I am moving it to draft. Might generate some noisy commits here as well so disabled CI here.
for the moment.

PR #458, opened 2024-01-03
feat: add underlinking/overlinking checks

### LINKED ISSUE #527: Implement missing fields of `dynamic_linking`
A few fields are missing in the `dynamic_linking` section. Also depends on #458 and #75.

`binary_relocation` and `rpath_allowlist` are already handled.

```yaml
  # settings for shared libraries and executables
  dynamic_linking:
    # linux only, list of rpaths (was rpath)
    rpaths: [path] (defaults to ['lib/'])

    # wether to relocate binaries or not. If this is a list of paths, then
    # only the listed paths are relocated
    binary_relocation: bool (defaults to true) | [glob]

    # Allow linking against libraries that are not in the run requirements
    # (was `missing_dso_whitelist`)
    missing_dso_allowlist: [glob]

    # Allow runpath / rpath to point to these locations outside of the environment
    # (was `runpath_whitelist`)
    rpath_allowlist: [glob]

    # what to do when detecting overdepending
    overdepending_behavior: OneOf<"ignore" | "error"> # (defaults to "error")

    # what to do when detecting overlinking
    overlinking_behavior: OneOf<"ignore" | "error"> # (defaults to "error")
```

wolfv:
This is emulating some Python code to determine the "nature" of a package for over and underlinking checks.

edit: closes #527


wolfv:
This code is supposed to do the following:

- classify all the dependencies in the `host` environment (e.g. are they packages that contain only shared libraries?)
- for all packages that contain shared libraries, check if we are linking against them
- if we are linking against them, make sure that they are also included in the run dependencies (if they are not included in the run dependencies, it's an issue and we call it underlinking)
- if there are some packages included in the run dependencies, that we are _not_ actually linking against, then that is another issue that we are calling overlinking

So far I've implemented the same "classification" methods as are implemented in conda-build for the dependencies. 

Now we need to resolve all the shared libraries that are linked by any of the dynamic objects included in our package and then emit warnings / errors based on this check. 

One other step for this would be to add a `resolve` function to the dynamic object that resolves the shared library and takes into account the `rpath`.

wolfv:
uh oh: Failed to parse the ELF file: BadMagic(7955925836558921542)

orhun:
damn, any idea what happened? can you run those tests locally?

wolfv:
I pushed a fix for the test issue (I think).

orhun:
Thanks!

macos pipeline fails due to:

```
thread 'tests::test_underlinking_check' panicked at rust-tests/src/lib.rs:585:9:
assertion failed: output.contains(\"Linking check error: Underlinking against: libzlib\")
```

It is either because underlinking check does not work or `libzlib` is something else on macos...

orhun:
what a wild ride... ready for review!

orhun (approved):

PR #464, opened 2024-01-05
Add support for uploading to anaconda.org

### LINKED ISSUE #440: Implement `upload` command for Anaconda.org
Upload works for quetz, artifactory and prefix.dev but not for anaconda.org yet.

I think originally @travishathaway was interested in this.

0xbe7a:
- add uploading to anaconda.org
- use shared Package struct
- rename to ExtractedPackage

Closes #440

### DIFF
--- a/.github/workflows/end-to-end.yml
+++ b/.github/workflows/end-to-end.yml
@@ -40,3 +40,4 @@ jobs:
       - run: pixi run end_to_end_test
         env:
           PREFIX_DEV_READ_ONLY_TOKEN: ${{ secrets.PREFIX_DEV_READ_ONLY_TOKEN }}
+          ANACONDA_ORG_TEST_TOKEN: ${{ secrets.ANACONDA_ORG_TEST_TOKEN }}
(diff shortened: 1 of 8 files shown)


0xbe7a:
Should we add a `--force` option that replaces uploaded files in case of conflicts instead of failing @wolfv ? In conda-forge this is afaik never intended

wolfv:
Awesome! I do think that adding a `--force` option is good, because some people use that for testing. There should be other mechanisms to prevent that in `conda-forge` :)

0xbe7a:
I have added a `--force` option that deletes existing files in case of conflicts

0xbe7a:
We should probably add an end-to-end test to upload a test package to anaconda.org and maybe test the --force option as well.

wolfv:
Looks great, and I am very excited about this "feature-completeness"! 

Do you want to add the e2e test in this PR or do you want to make another PR?

0xbe7a:
I will include it in this PR

PR #465, opened 2024-01-06
add conda-forge integration

0xbe7a:
Until https://github.com/Anaconda-Platform/anaconda-client/pull/696 is merged, anaconda-client cannot be used to upload packages with conda-forge-ci-setup. This PR adds a conda-forge upload as a special case of the anaconda upload, which makes a copy request after the upload.

conda-forge-ci-setup also uses conda-build under the hood, so this allows testing of rattler-build for conda-forge packages without being blocked by conda-build. In the long run, we probably want to integrate rattler-build properly into conda-forge-ci-setup / fix existing problems of conda-build / anaconda-client when uploading packages built by rattler-build


0xbe7a:
Currently the channel / label is hardcoded (see https://github.com/conda-forge/polarify-feedstock/blob/rattler-build/.scripts/build_steps.sh#L83). Do you think we could we could read this information from the uploaded package itself / include it in the package @wolfv ? I'm not quite sure if the "label / channel" concept is anaconda.org or a general conda concept and how much of it should go into the package format

wolfv:
Isn't the channel / label stuff something that `conda-smithy` would handle? It's got to be somewhere passed in to the `recipe`. 

Instead of using `yg` you can also just use the YAML parser of `rattler-build`. The `.ci_config/...yaml` files are already evaluated so that they contain "pure YAML" and should be easy to parse (as long as you select the proper one).

At the end of the day, a label is just a channel. Personally not a huge fan of labels, so if we could only talk about channels I would probably prefer it but since this code is anaconda specific anyways, i don't mind.

0xbe7a:
I took a look at how conda-forge-ci-setup does it, and they parse the variant file and use the channel_targets field to determine the upload channel. Since the variant file is contained in the info/recipe folder of the package, we can determine the target channel directly from the package

wolfv:
great, will merge! 

Just a side note: I think at some point we should refactor all the upload methods to return proper errors because some people might want to use the methods in a library (or from Python, when we have the bindings). But for now, miette does a nice job :)

PR #467, opened 2024-01-09
Remove rpaths that point outside the prefix

### LINKED ISSUE #457: Remove rpaths that point outside of the prefix
conda-build removes rpath entries that point to e.g. the "build" prefix

```py
    for old in existing:
        if old.startswith("$ORIGIN"):
            new.append(old)
        elif old.startswith("/"):
            # Test if this absolute path is outside of prefix. That is fatal.
            rp = relpath(old, prefix)
            if rp.startswith(".." + os.sep):
                print(f"Warning: rpath {old} is outside prefix {prefix} (removing it)")
            else:
                rp = "$ORIGIN/" + relpath(old, origin)
                if rp not in new:
                    new.append(rp)
```

orhun:
closes #457

See https://github.com/conda/conda-build/blob/03f2a2af949abc2d5a7a20d86ac1d06af0b7bc17/conda_build/post.py#L622

e.g.

```
WARN rattler_build::linux::link: rpath ("/home/orhun/gh/rattler-build/output/bld/rattler-build_zlink_1704804443/build_env/lib") is outside of prefix ("/home/orhun/gh/rattler-build/output/bld/rattler-build_zlink_1704804443/host_env_placehold_.....") - removing it
```


src/linux/link.rs
```diff
@@ -100,15 +100,23 @@ impl SharedObject {
...
+                    &new_rpath,
                     self.path.parent().ok_or(RelinkError::NoParentDir)?,
                 )
                 .ok_or(RelinkError::PathDiffFailed)?;
-                tracing::info!("New relative path: $ORIGIN/{}", relative_path.display());
-                final_rpath.push(PathBuf::from(format!(
-                    "$ORIGIN/{}",
-                    relative_path.to_string_lossy()
-                )));
+                if relative_path.starts_with("..") {
```
wolfv, line 107: Does `Component::Parent` work with `starts_with`, too? For Path's.

orhun: Yes,

```rust
use std::path::Path;

fn main()  {
    let path = Path::new("../test");
    assert!(path.starts_with(".."));
    assert_eq!(Some(Path::new("..")), path.parent());
}
```

Do you prefer checking that instead?

baszalmstra:
Im wondering if we should also emit which file contains the rparh?

wolfv (approved):

orhun:
> Im wondering if we should also emit which file contains the rparh?
can you elaborate? @baszalmstra

baszalmstra:
If I understand it correctly a warning is emited if a binary file is found that contains an rpath outside of the build prefix. 

But the warning does not specify which file contained the rpath if Im not mistaken. I was thinking that that would be useful information.

orhun:
Ah I see what you mean, done in b98613dec574b7713c8954ba0130ad195c125eea

wolfv (approved):

baszalmstra (approved):

Issue #244, opened 2023-10-30
Support injecting variables from other files through jinja syntax


pavelzw:
Similar to #138

In conda-build, something like `load_file_data('pyproject.toml').get('project').get('version')` is possible. Would be nice to get something similar here as well.

wolfv:
I think my question here is wether this would be a local file or supposed to come from the source section of the recipe. 

I think it would be feasible if it's a local file / recipe. But if the sources need to be downloaded first to discover these missing variables it would create a dependency within the rendering / execution of the recipe that I wouldn't want.

pavelzw:
For my personal use case (where my sources are local too), a local file would be sufficient.

PR #468, opened 2024-01-09
feat: add `load_from_file` jinja function

swarnimarun:
This PR adds load_from_file function that either loads a map from yaml, json, or toml, or otherwise just loads the entire string source from the file itself.

This is supposed to be an experimental feature.

Note: Currently it doesn't check if the path is inside the output_dir. But it shouldn't be a concern as it happens during jinja to yaml build phase which means we can't have cyclic dependency.

Fixes https://github.com/prefix-dev/rattler-build/issues/244


wolfv:
Other PR is merged :)

swarnimarun:
rebased

wolfv:
Looks good! If you can write some docs on this, the `git` functions and the `--experimental` flag that would be appreciated! :)

swarnimarun:
finally working. :3

Issue #113, opened 2023-04-16
Implement `ignore_run_exports` and `ignore_run_exports_from`


wolfv:
We currently don't use the two values.

- [ ] `ignore_run_exports`: a list of run exports to ignore
- [ ] `ignore_run_exports_from`: a list of packages whose run_exports we should ignore

PR #472, opened 2024-01-10
fix: use ignore_run_exports for building env

swarnimarun:
- This PR uses `ignore_run_export` while building finalized run dependencies.

Closes https://github.com/prefix-dev/rattler-build/issues/113

Note:
`ignore_run_exports_from` is already implementated,

https://github.com/prefix-dev/rattler-build/blob/3e883a6d4d99484ee698ab57cea2e42aebbdecf1/src/render/resolved_dependencies.rs#L481-L487

### DIFF
--- a/src/render/resolved_dependencies.rs
+++ b/src/render/resolved_dependencies.rs
@@ -517,6 +517,23 @@ pub async fn resolve_dependencies(
         let mut cloned = Vec::new();
         for spec in specs {
             let spec = MatchSpec::from_str(spec)?;
+            let in_ignore_run_exports = |pkg| {
+                output
+                    .recipe
+                    .requirements()
+                    .ignore_run_exports()
+                    .by_name()
+                    .contains(pkg)
+            };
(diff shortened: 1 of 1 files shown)


wolfv:
I think there may be a misunderstanding (or I am reading the code wrong) :)

A recipe can _create_ it's own run-exports. Those are defined by the recipe.

However, when evaluating a recipe and solving for the packages, other packages can introduce `run_exports` which ultimately change the run dependencies of the package that is being built. Those are the ones we want to ignore, either by name of the package that the run exports are coming from or by name of the run export itself.


For example, the `libfoo` package migth define two run exports:

run_exports:
  - libfoo 1.23
  - foo *

Now, when another package depends on `libfoo` in the _host_ section, then the run exports will be attached to the final run dependencies:

```
host:
  - libfoo
run:
  - libfoo 1.23  # from run-exports
  - foo          # from run-exports
```

And those are the ones we may want to ignore.

PR #473, opened 2024-01-10
feat: add `rpath_allowlist` to the recipe

orhun:
This PR makes it possible to define a list of glob paths allowing rpaths that are outside of the environment.

```yml
build:
  dynamic_linking:
    rpath_allowlist: [""]
```

Todolist:

- [ ] Add tests
- [ ] Do we need an example for this? If so, how should we do that?


wolfv:
Looks like a great start! Would be nice to see an `insta` test on how the glob' serializes (because we use serde to store rendered recipes as part of the package).

orhun:
The `glob` serializes like the following:

```
         89 │+        dynamic_linking: Some(
         90 │+            DynamicLinking {
         91 │+                rpath_allowlist: [
         92 │+                    Glob {
         93 │+                        glob: "/usr/lib/**",
         94 │+                        re: "(?-u)^/usr/lib/.*$",
         95 │+                        opts: GlobOptions {
         96 │+                            case_insensitive: false,
         97 │+                            literal_separator: false,
         98 │+                            backslash_escape: true,
         99 │+                            empty_alternates: false,
        100 │+                        },
        101 │+                        tokens: Tokens(
        102 │+                            [
        103 │+                                Literal(
        104 │+                                    '/',
        105 │+                                ),
        106 │+                                Literal(
        107 │+                                    'u',
        108 │+                                ),
        109 │+                                Literal(
        110 │+                                    's',
        111 │+                                ),
        112 │+                                Literal(
        113 │+                                    'r',
        114 │+                                ),
        115 │+                                Literal(
        116 │+                                    '/',
        117 │+                                ),
        118 │+                                Literal(
        119 │+                                    'l',
        120 │+                                ),
        121 │+                                Literal(
        122 │+                                    'i',
        123 │+                                ),
        124 │+                                Literal(
        125 │+                                    'b',
        126 │+                                ),
        127 │+                                RecursiveSuffix,
        128 │+                            ],
        129 │+                        ),
        130 │+                    },
        131 │+                ],
        132 │+            },
        133 │+        ),
   90   134 │     },
```

Is that fine or should we go with raw string instead?

wolfv:
Wow, that is very verbose and brittle serialization ... i prefer a string, indeed!

orhun:
I made the changes in ec4b1e52c4002c986014359a45310455dc1df62a

I was going to pass around the `DynamicLinking` struct but it requires to be public in this case. (`pub mod build`)

Now I'm working on updating the relink test for Linux. After that I will check out the macOS part.

PR #481, opened 2024-01-12
fix: non deterministic git tests

swarnimarun:
Rewrote my older test which had issues, this fixes failing unit-tests.


swarnimarun:
and now it still fails in the CI, sec let me test this out on linux or otherwise in ssh'ed github runner. Failing doesn't make much sense as it works on mac and windows for me.

swarnimarun:
A quick ssh session later, apparently macos github image has git config preset...

swarnimarun:
working now @wolfv

PR #484, opened 2024-01-12
expose and store compression level and compression threads

### LINKED ISSUE #376: Support compression level setting for bz2 and zstd
It would be nice to have support for something like `--zstd-compression-level` or `--package-format conda:15`

wolfv:
fixes #376

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -61,9 +61,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.5"
+version = "0.6.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


pavelzw:
@wolfv is there a reason why `compression_threads` defaults to `1` and not the highest amount?

wolfv:
Not really except that I haven't tested it much yet. But I am fine with changing it to `num_cpu` (although that should be done earlier in the code because we want to store that in the recipe, too). I haven't yet verified that compression threads do not change the SHA256 of the final package

wolfv:
Yeah, we would allow to use a different thread count, but not store it in the recipe. At least that's the idea ...

wolfv:
But if we verify that the thread count doesn't change the result we can also skip storing it.

PR #489, opened 2024-01-12
Use rattler index crate instead of local index functionality

BenjaminLowry:
In https://github.com/mamba-org/rattler/pull/436 I copied the index logic in this repo to rattler and created a new crate for it so that I could write python bindings and thus have access to index functionality via py-rattler. This PR replaces the local index crate with rattler's since that is where this indexing logic will live now.


BenjaminLowry:
@wolfv Hey! Here is the PR for removing the index functionality from rattler-build now that it's in rattler, let me know if I'm missing anything

wolfv (approved):
Looks good, thanks!

Issue #102, opened 2023-04-15
Implement `patchelf` and `install_name_tool` functionality with goblin in Rust


wolfv:
This would be very nice, for cross-platform builds.

wolfv:
Another option could be the `object` crate: https://docs.rs/object/latest/object/

wolfv:
cc @matthiasbeyer you might want to check out the `object` crate as well :)

wolfv:
I asked a question on the `object` repo: https://github.com/gimli-rs/object/issues/148#issuecomment-1787180928

wolfv:
Another good reference for `install_name_tool` (MachO) is from Homebrew, written in Ruby: https://github.com/Homebrew/ruby-macho/blob/master/lib/macho/macho_file.rb

orhun:
I think this can be closed now.

PR #494, opened 2024-01-15
feat: store system information in the rendered recipe

### LINKED ISSUE #485: Store "system" information in the rendered recipe
We should store the `rattler-build` version in the rendered recipe (so we could potentially grab that version for the rebuilding of a given package/recipe).

We might also want to store some additional stuff like:

- version of `patchelf`
- version of `install_name_tool`
- version of `git`? Although that one probably (hopefully) doesn't make a difference.

orhun:
closes #485

### DIFF
--- a/src/recipe/parser/system_info.rs
+++ b/src/recipe/parser/system_info.rs
@@ -10,9 +10,6 @@ pub struct SystemInfo {
     /// `patchelf` version (Linux).
     #[serde(skip_serializing_if = "Option::is_none")]
     patchelf_version: Option<String>,
-    /// `install_name_tool` version (macOS).
-    #[serde(skip_serializing_if = "Option::is_none")]
-    install_name_tool_version: Option<String>,
 }
 
 impl SystemInfo {
@@ -24,12 +21,7 @@ impl SystemInfo {
             ..Default::default()
         };
         system_info.git_version = Self::run_os_command("git", "version");
-        if cfg!(target_os = "linux") {
-            system_info.patchelf_version = Self::run_os_command("patchelf", "--version");
-        } else if cfg!(target_os = "macos") {
-            system_info.install_name_tool_version =
-                Self::run_os_command("install_name_tool", "--version");
-        }
+        system_info.patchelf_version = Self::run_os_command("patchelf", "--version");
         system_info
     }
 
@@ -40,7 +32,6 @@ impl SystemInfo {
             rattler_build_version: String::from("test"),
             git_version: Some(String::from("2.43.0")),
             patchelf_version: Some(String::from("0.18.0")),
-            install_name_tool_version: Some(String::from("0.0.0")),
         }
     }
 
--- a/src/recipe/parser/system_info.rs
+++ b/src/recipe/parser/system_info.rs
@@ -8,8 +8,10 @@ pub struct SystemInfo {
     /// Git version.
     git_version: Option<String>,
     /// `patchelf` version (Linux).
+    #[serde(skip_serializing_if = "Option::is_none")]
     patchelf_version: Option<String>,
     /// `install_name_tool` version (macOS).
+    #[serde(skip_serializing_if = "Option::is_none")]
     install_name_tool_version: Option<String>,
 }
 
--- a/src/recipe/parser/system_info.rs
+++ b/src/recipe/parser/system_info.rs
@@ -1,0 +1,71 @@
+use serde::{Deserialize, Serialize};
+
+/// The system information for the rendered recipe.
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+pub struct SystemInfo {
+    /// `rattler-build` version.
+    rattler_build_version: String,
+    /// Git version.
+    git_version: Option<String>,
+    /// `patchelf` version (Linux).
+    patchelf_version: Option<String>,
+    /// `install_name_tool` version (macOS).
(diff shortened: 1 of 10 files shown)


src/recipe/parser/system_info.rs
```diff
@@ -0,0 +1,71 @@
...
+
+/// The system information for the rendered recipe.
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+pub struct SystemInfo {
+    /// `rattler-build` version.
+    rattler_build_version: String,
+    /// Git version.
+    git_version: Option<String>,
+    /// `patchelf` version (Linux).
+    patchelf_version: Option<String>,
```
wolfv, line 11: I would prefer to skip serialize these fields if they are `None`.

src/recipe/parser/system_info.rs
```diff
@@ -0,0 +1,71 @@
...
+impl SystemInfo {
+    /// Collects the system information.
+    #[cfg(not(test))]
+    pub fn new() -> Self {
+        let mut system_info = SystemInfo {
+            rattler_build_version: env!("CARGO_PKG_VERSION").to_string(),
+            ..Default::default()
+        };
+        system_info.git_version = Self::run_os_command("git", "version");
+        if cfg!(target_os = "linux") {
```
wolfv, line 25: This is not correct, one could have `patchelf` on macOS as well: https://prefix.dev/channels/conda-forge/packages/patchelf

We could check the `target_platform` of the `output` to figure out wether we are compiling for macOS or Linux and based on that information only log the relevant one.

src/recipe/parser/system_info.rs
```diff
@@ -0,0 +1,71 @@
...
+            system_info.patchelf_version = Self::run_os_command("patchelf", "--version");
+        } else if cfg!(target_os = "macos") {
+            system_info.install_name_tool_version =
```
wolfv, line 28: I just checked and `--version`, `-v` or `-V` flags are not available for `install_name_tool` ... will check further if there is a way

wolfv (changes requested):

src/recipe/parser/system_info.rs
orhun: Ah, good point!

> We could check the target_platform of the output to figure out wether we are compiling for macOS or Linux and based on that information only log the relevant one.

Any idea how to access that information inside `Recipe`? 👀

src/recipe/parser/system_info.rs
wolfv: Not even the manpage shows a version ... https://www.unix.com/man-page/osx/1/install_name_tool/

wolfv: Maybe we just have to accept this for now and store only the version. The #102 will remove the dependency anyways...

orhun: RIP.

swarnimarun: We can always use [llvm-install-name-tool](https://llvm.org/docs/CommandGuide/llvm-install-name-tool.html#llvm-install-name-tool-llvm-tool-for-manipulating-install-names-and-rpaths). They implement all the cli quality of life things. And are essentially a drop-in replacement.

src/recipe/parser/system_info.rs
orhun: Now that we will remove `install_name_tool`, does that mean we can store patchelf version for both platforms? What do you prefer here?

src/recipe/parser/system_info.rs
orhun: and it has `--version, -V` flag!

wolfv: Good point @swarnimarun. I hope that we can complete https://github.com/prefix-dev/rattler-build/pull/442 soon and then we don't rely on `install_name_tool` at all anymore.

orhun: Using `llvm-install-name-tool` is a good idea but I'm not sure we can *just* switch to it. It might be better to wait for #442 as Wolf said.

I removed that field in 35ad61ad687dccafa92723aa266d2890ba81e73b

wolfv:
Hmm, so I think we should actually do this a little different (after I've done some reflecting). Here are my thoughts:

In our code, whenever we call an external tool, it should go through the "system" object. For example:

```
system.git(args, ...);
```

When we invoke this, we'll store the version that was used. The storing happens on an interior mutable type or hashmap or so, and if we already have a version (because we're running `rattler-build rebuild ...` we do _not_ overwrite the value that is already in there but rather give the user a warning, like:

```
... rebuild: using patchelf 0.19 instead of 0.18 as in the original build ...
```

orhun:
That makes sense! I will try to implement that.

orhun:
I have a quick untested/PoC of this new approach here: https://github.com/orhun/rattler-build/commit/349e3df9a4b1a42415124581896be2b6617dabab

I'm guessing storing the tool/version for each command invocation means globally passing around an object so I went ahead with that. The problem with that approach is that we need to use the intermediate `SystemInfo::call_tool` function when we want to construct `std::process::Command` before the actual execution.

We can clarify how this can be handled differently/better. I would hold pushing to this PR until we land on a good enough solution we can agree 🐻 

Let me know what you think!

wolfv:
Closed because we merged #587

PR #501, opened 2024-01-17
implement `patchelf` and `install_name_tool` in Rust

wolfv:


src/linux/link.rs
```diff
@@ -168,11 +172,141 @@ fn call_patchelf(elf_path: &Path, new_rpath: &[PathBuf]) -> Result<(), RelinkErr
     }
 }
 
+fn ctx(object: &Elf) -> goblin::container::Ctx {
```
orhun, line 175: It is not clear why this is needed at first glance, let's document this function!

src/linux/link.rs
```diff
@@ -168,11 +172,141 @@ fn call_patchelf(elf_path: &Path, new_rpath: &[PathBuf]) -> Result<(), RelinkErr
...
+
+    let object = goblin::elf::Elf::parse(&data)?;
+
+    let dynamic = match object.dynamic.as_ref() {
+        Some(dynamic) => dynamic,
+        None => {
+            tracing::debug!("{} is not dynamically linked", elf_path.display());
+            return Ok(());
+        }
+    };
```
orhun, line 215: We can probably use the already existing `SharedObject` struct for this.

src/linux/link.rs
```diff
@@ -168,11 +172,141 @@ fn call_patchelf(elf_path: &Path, new_rpath: &[PathBuf]) -> Result<(), RelinkErr
...
+
+    let overwrite_strtab =
+        |data_mut: &mut MmapMut, offset: usize, new_value: &str| -> Result<(), RelinkError> {
+            let new_value = new_value.as_bytes();
+            let old_value = dynstrtab
+                .get_at(offset)
+                .ok_or(RelinkError::PatchElfFailed)?;
+
+            if new_value.len() > old_value.len() {
+                panic!("new value is longer than old value");
```
orhun, line 243: Return an error instead?

src/linux/link.rs
```diff
@@ -168,11 +172,141 @@ fn call_patchelf(elf_path: &Path, new_rpath: &[PathBuf]) -> Result<(), RelinkErr
...
+    if needs_rewrite {
+        // now we need to write the new dynamic section
+        let header_offset = object
+            .program_headers
+            .iter()
+            .find(|header| header.p_type == goblin::elf::program_header::PT_DYNAMIC)
+            .map(|header| header.p_offset)
+            .ok_or(RelinkError::PatchElfFailed)? as usize;
+
+        let mut offset = header_offset;
```
orhun, line 293:
```suggestion
        // now we need to write the new dynamic section
        let mut offset = object
            .program_headers
            .iter()
            .find(|header| header.p_type == goblin::elf::program_header::PT_DYNAMIC)
            .map(|header| header.p_offset)
            .ok_or(RelinkError::PatchElfFailed)? as usize;

```

wolfv:
Hmmm, I am actually wondering right now if we need to take into account the endianness when writing the string?! E.g. should we use `pwrite<&str>` instead of just dumping the bytes?

orhun:
Made some formatting/ergonomic/docs changes.

src/linux/link.rs
```diff
@@ -241,10 +242,11 @@ fn relink(elf_path: &Path, new_rpath: &[PathBuf]) -> Result<(), RelinkError> {
             let new_value = new_value.as_bytes();
             let old_value = dynstrtab
                 .get_at(offset)
-                .ok_or(RelinkError::PatchElfFailed)?;
+                .ok_or(RelinkError::BuiltinPatcherFailed)?;
 
             if new_value.len() > old_value.len() {
-                panic!("new value is longer than old value");
+                tracing::error!("new value is longer than old value");
+                return Err(RelinkError::BuiltinPatcherFailed);
```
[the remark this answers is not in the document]
orhun: Done in ec4fcad13b2048161478469eb7a2131132e2214f

src/linux/link.rs
```diff
@@ -317,6 +317,7 @@ fn relink(elf_path: &Path, new_rpath: &[PathBuf]) -> Result<(), RelinkError> {
 }
 
 #[cfg(test)]
+#[cfg(target_os = "linux")]
```
wolfv, line 320: This isn't necessary for the `builtin` test. Part of the point is that it will make cross-compilation (a little) easier! :)

orhun (approved):
LGTM

wolfv:
I need to adjust one error type in the macos version of the relinking, and then we can merge :rocket: :)

orhun:
Neat. 🐑 it!

PR #507, opened 2024-01-18
add progress bar for .tar.bz2

wolfv:
@pavelzw a little prototype for you :) 

For `.conda` packages it's probably a bit more tricky. We probably don't care about a progress bar for the `info/*` files. Maybe it needs an additional argument to the package conda function or so.

### DIFF
--- a/src/packaging.rs
+++ b/src/packaging.rs
@@ -1,9 +1,10 @@
 use content_inspector::ContentType;
 use fs_err as fs;
 use fs_err::File;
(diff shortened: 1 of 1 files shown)


wolfv:
Screenshot 2024-01-18 at 20 22 10

wolfv:
While it looks cool it's actually wrong (because it measures how many compressed bytes have been written and compares that to the number of uncompressed bytes). Instead, we should wrap the reader file object. For that we'll have to extend rattler a little bit ... but should be nice & doable!

PR #508, opened 2024-01-19
add progress bar for package write

### LINKED ISSUE #496: Add progress bar for large compressions
For really large packages (1GB), when setting `--package-format conda:max` it can happen that rattler-build gets stuck for quite a long time at

```
Creating target folder "/Users/pavel/projects/my-package/output/linux-64"
```

It would be nice to have a progress bar here.

swarnimarun:
Testing a bit large-ish packages seems stuck, this PR provides some feedback with progress bar for users.

Fixes #496

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -317,6 +317,7 @@ pub async fn run_build(
         &directories.host_prefix,
         &directories.output_dir,
         &output.build_configuration.packaging_settings,
+        &tool_configuration,
     )
     .into_diagnostic()?;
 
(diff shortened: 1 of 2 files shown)


wolfv:
@swarnimarun nice – your solution is a bit slimmer than mine (#507) :D 

I think we need to ideally extend `rattler` though, and wrap the `Read` of the file (so that we know how many bytes have been processed out of the total number of bytes).

And you are setting the progress bar to `1` – so what does that do when writing? Will have to test this out a bit.

wolfv:
@swarnimarun pls have a look at my PR (#507) where I did that. But we need to change rattler (because the reading is done there). I was thinking to extend the functions with a variant that takes a progress bar or some trait impl (`compress ... _with_progress`)

PR #532, opened 2024-01-24
Add rpath test

pavelzw:


wolfv:
I rebased & pushed.

wolfv:
Hmm, looks like we need to `brew install patchelf` (or `pixi global install patchelf`). 

At some point I want `rattler-build` to manage it's external tool dependencies by itself (e.g. in a `tools` folder somewhere on the system where patchelf, install_name_tool, git, etc. are installed if needed.

wolfv:
I took over! :) hope it's ok

wolfv:
hmm, i might have broken something on main :D will try to revert my last commit

PR #538, opened 2024-01-24
properly set stdout and stderr capture  in rust-tests

0xbe7a:
During testing, the Rust standard library normally replaces the implementation of `println!`, `eprintln!` etc. to capture their output instead of printing directly to stdout / stderr. This gives you the nice default behavior of printing only the output for failed tests and muting the output for all other tests. You can also print everything by using --nocapture. This PR now captures the output of rattler-build instead of just inheriting stdout and stderr and giving the user the same behavior as the normal println!


wolfv:
Kinda funny because just yesterday I added os_pipe (https://github.com/prefix-dev/rattler-build/pull/534) – but `duct` also looks good to me. Since it's only in the tests, I don't care that much anyways :)

PR #540, opened 2024-01-25
add `GlobVec`, `always_copy_files` and `always_include_files`

wolfv:
cc @orhun you also did some work on glob's recently.
This introduces a new type that renders them as regular strings, but also eagerly parses them so that we can emit errors right away.

### DIFF
--- a/src/recipe/parser/glob_vec.rs
+++ b/src/recipe/parser/glob_vec.rs
@@ -1,0 +1,140 @@
+use globset::{Glob, GlobSet};
+
+use serde::ser::SerializeSeq;
+use serde::{Deserialize, Serialize};
+
+use crate::_partialerror;
+use crate::recipe::custom_yaml::{HasSpan, RenderedNode, RenderedSequenceNode, TryConvertNode};
+use crate::recipe::error::{ErrorKind, PartialParsingError};
+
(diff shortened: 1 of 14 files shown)


src/recipe/parser/glob_vec.rs
```diff
@@ -0,0 +1,140 @@
...
+        }
+    }
+}
+
+impl GlobVec {
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
+    pub fn globset(&self) -> Option<&GlobSet> {
```
orhun, line 54: Should we also add a getter method for getting the underlying Vector of `GlobMatcher`s? Or is `GlobSet` enough to perform matches?

orhun (commented):
I like this! Just had one comment since I mainly used `GlobMatcher`s.

src/recipe/parser/glob_vec.rs
wolfv: Let's do that when we need it? Maybe `globset` is enough -- that depends I'd say.

orhun: Alright! I will switch to using `GlobVec` after this PR is merged and see if that's necessary.

orhun (approved):

PR #541, opened 2024-01-25
Fix version check and create issue on failure

pavelzw:


wolfv:
ready?!

.github/workflows/rust.yml
```diff
@@ -193,3 +195,13 @@ jobs:
             ${{ steps.package.outputs.pkg-path }}
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Create issue on failure
```
pavelzw, line 199: Haven't checked this one but should be fine:tm:

PR #545, opened 2024-01-26
feat: add `merge_build_and_host_envs` section

### LINKED ISSUE #518: Implement `merge_build_and_host_envs`
Was: `merge_build_host` in `conda-build`.

This option would just combine the specs from build and host env and create a joint "host" environment.

This is used in many `R` recipes on Windows: https://github.com/search?q=org%3Aconda-forge+merge_build_host%3A+true+language%3AYAML&type=code&l=YAML

swarnimarun:
- Added the `merge_build_and_host_envs` section as named in `schema.json`.
- It merges build into host env. But leaves the build env untouched. 
- Run dependencies are built from new host env.
- Modified the test to use `merge_build_and_host_envs`

Fixes #518


src/recipe/parser/build.rs
```diff
@@ -55,9 +55,17 @@ pub struct Build {
     /// Setting to control wether to always include a file (even if it is already present in the host env)
     #[serde(default, skip_serializing_if = "GlobVec::is_empty")]
     pub(super) always_include_files: GlobVec,
+    /// Merge the build and host envs
+    #[serde(default)]
```
wolfv, line 59: Would be cool to skip rendering this if it's false (the default).

I think I added a `is_false` function somewhere. We could move it to a `serde_utils.rs` file.

swarnimarun: We can use `std::ops::Not::not`

src/render/resolved_dependencies.rs
```diff
@@ -594,6 +596,31 @@ pub async fn resolve_dependencies(
         None
     };
 
+    let host_env = if merge_build_host {
```
wolfv, line 599: I think this is the wrong place. The merging should happen of the `matchspecs`, and before the environment is resolved. 

E.g. if you have

```
build:
  - compiler
host:
  - bla
```

It should just insert `compiler` into the `host` env (before resolving & installing) and then resolve everything at once into the same environment. We shouldn't have to touch the `ResolvedDependencies` as they should be fine (at that stage).

```
    let match_specs = specs.iter().map(|s| s.spec().clone()).collect::<Vec<_>>();
    // insert build match specs here!
    let host_env = if !match_specs.is_empty() {
```

wolfv (changes requested):
There is a little logic bug on where `host` and `build` deps are merged. It needs to happen at the `matchspec` level :)

src/render/resolved_dependencies.rs
swarnimarun: oh, that makes sense. Yeah pretty easy to change.

src/render/resolved_dependencies.rs
```diff
@@ -553,7 +553,16 @@ pub async fn resolve_dependencies(
         }
     }
 
-    let match_specs = specs.iter().map(|s| s.spec().clone()).collect::<Vec<_>>();
+    let match_specs = specs
```
wolfv, line 556: this looks fine but should only happen when the `merge_build_host` is true. 
Also in that case, the build env does not need to be created.

wolfv (changes requested):
still some logic fixes necessary

wolfv:
It also still creates the build env ... We want to merge build and host, and therefore don't need to create the build env at all.

src/render/resolved_dependencies.rs
```diff
@@ -663,26 +674,30 @@ pub async fn resolve_dependencies(
...
-                    run_specs
-                        .depends
-                        .extend(clone_specs(name, "build", &rex.strong)?);
-                    run_specs.constrains.extend(clone_specs(
-                        name,
-                        "build",
-                        &rex.strong_constrains,
-                    )?);
+    // if already merged don't redo
+    if !merge_build_host {
```
wolfv, line 678: this isnt' reallt necessary anymore, right? Since the build env is None now.

wolfv (approved):

PR #546, opened 2024-01-26
feat: add variant section

### LINKED ISSUE #521: Implement `variant` section
The `variant` section contains two entries: `use_keys` and `ignore_keys`.

Was: `force_ignore_keys` and `force_use_keys`.

These are used to forcibly add new variants (`use_keys`) or ignore a key from the hash computation. We would need to look up the exact semantics used in `conda-build` (does this _only_ affect the hash computation?).

```yaml
build:
  variant:
    # Keys to forcibly use for the variant computation (even if they are not in the dependencies)
    use_keys: [string]

    # Keys to forcibly ignore for the variant computation (even if they are in the dependencies)
    ignore_keys: [string]
```

https://github.com/search?q=org%3Aconda-forge+force_ignore_keys+language%3AYAML&type=code
https://github.com/search?q=org%3Aconda-forge+force_use_keys+language%3AYAML&type=code

swarnimarun:
- Adds a schema compliant `variant` section.
- Handles variant key skipping.
- use_keys forces keys into used_vars
- ignore_keys removed keys from used_vars
- Added test to verify used_variables are correctly modified.

Fixes #521


src/recipe/parser/build.rs
```diff
@@ -55,9 +101,16 @@ pub struct Build {
     /// Setting to control wether to always include a file (even if it is already present in the host env)
     #[serde(default, skip_serializing_if = "GlobVec::is_empty")]
     pub(super) always_include_files: GlobVec,
+    /// Variant ignore and use keys
+    #[serde(default, skip_serializing_if = "Option::is_none")]
```
wolfv, line 105: I prefer to have an `is_default` on the `VariantKeyUsage` and get rid of the `Option`.

swarnimarun: We can use `std::ops::Not::not`

wolfv: You didn't push any changes if I see correctly.

PR #567, opened 2024-01-31
Fix release binaries

pavelzw:
testing it out at https://github.com/pavelzw/rattler-build/actions/runs/7724654034/job/21057223260


pavelzw:
Seems to work https://github.com/pavelzw/rattler-build/releases/tag/v0.8.0

PR #569, opened 2024-01-31
skip over any Null values in mappings

wolfv:
This is an extension to the fix for the `env` mapping and skips over any `Null` values in `RenderedMapppings`. 

That means, keys are ignored if the value is `Null`. I do think that's the desired behavior and means we can write stuff like 

```yaml
build:
  number: ${{ 10 if false }}
```

without error.


wolfv:
Note: even when the value is Null, duplicate keys should not be allowed (see #570)

PR #592, opened 2024-02-05
feat: experimental indent when logging to terminal

wolfv:
TODO:

- [x] refactor `run_in_env` so that we can use it both in tests and build execution
- [x] Add more `span!` (e.g. when running the tests)
- [x] make interpreter selectable while we are at it :)
- [x] Test `rebuild` and `test` subcommands
- [x] Add "summary" function after the build(s) is done

Still needs fixing:

- [x] Summary is nice, but needs some more tweaks:
  - Add a Github subscriber that prints warning as `::warning::{message}` and error as `::error::{message}`
  - Make the summary have proper tables

### DIFF
--- a/src/render/solver.rs
+++ b/src/render/solver.rs
@@ -2,7 +2,7 @@ use anyhow::Context;
 use comfy_table::Table;
 use futures::{stream, stream::FuturesUnordered, FutureExt, StreamExt, TryFutureExt, TryStreamExt};
 
-use indicatif::{style::TemplateError, HumanBytes, ProgressBar, ProgressState, ProgressStyle};
+use indicatif::{style::TemplateError, HumanBytes, ProgressBar};
 use rattler::{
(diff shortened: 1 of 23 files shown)


wolfv:
It's starting to look cool: https://asciinema.org/a/640577

src/render/solver.rs
```diff
@@ -134,7 +133,7 @@ pub async fn create_environment(
                     platform,
                     &repodata_cache,
                     download_client.clone(),
-                    tool_configuration.multi_progress_indicator.clone(),
+                    tool_configuration.fancy_log_handler.clone(),
```
schuylermartin45, line 136: Amazing name lol

schuylermartin45 (approved):
Pretty straight forward, looks good.

pavelzw:
The "install yq with chocolatey" ci step seems to fail quite often... We could do `prefix-dev/setup-pixi` and afterwards `pixi global install yq`...

PR #605, opened 2024-02-07
Resolve rpath

wolfv:
- [ ] Linux implementation of the same function

### DIFF
--- a/src/macos/link.rs
+++ b/src/macos/link.rs
@@ -107,7 +107,7 @@ impl Dylib {
             encoded_prefix.join(self.path.strip_prefix(prefix).expect("dylib not in prefix"));
         if let Ok(rpath_without_loader) = rpath.strip_prefix("@loader_path") {
             if let Some(library_parent) = self_path.parent() {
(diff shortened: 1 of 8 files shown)


src/macos/link.rs
```diff
@@ -144,6 +160,17 @@ impl Dylib {
         let mut final_rpaths = Vec::new();
 
         for rpath in &self.rpaths {
+            if rpath.starts_with("@loader_path") {
+                let resolved = self.resolve_rpath(rpath, prefix, encoded_prefix);
+                if resolved.starts_with(prefix) {
+                    final_rpaths.push(rpath.clone());
+                } else if rpath_allowlist.map(|g| g.is_match(rpath)).unwrap_or(false) {
+                    tracing::warn!("Rpath not in prefix: {} – removing it", rpath.display());
+                }
```
orhun, line 169: Don't we have the same functionality at line 192-195 already?

wolfv: Yeah the difference is that this one is a relative one. We would need to determine if users should give us the actually encoded rpaths in the allowlist or wether they should give us the resolved path ... or wether both work.

PR #612, opened 2024-02-09
More docs

wolfv:
@pavelzw wdyt?

### DIFF
--- a/docs/variants.md
+++ b/docs/variants.md
@@ -178,7 +178,7 @@ build:
       # use cuda from the variant config, e.g. to build multiple CUDA variants
       - cuda
     # this will down-prioritize the cuda variant versus other variants of the package
-    down_prioritize_variant: ${{ -1 if cuda else 0 }}
+    down_prioritize_variant: ${{ 1 if cuda else 0 }}
 ```
 
 ### Mutex packages
@@ -205,10 +205,10 @@ package:
   version: 1.0
 
 build:
-  string: ${{ blas_variant }}
+  string: ${{ blas_variant }}${{ hash }}_${{ build_number }}
   variant_config:
     # make sure that `openblas` is preferred over `mkl`
-    down_prioritize_variant: ${{ -1 if blas_variant == "mkl" else 0}}
+    down_prioritize_variant: ${{ 1 if blas_variant == "mkl" else 0 }}
 ```
 
 This will create two package: `blas_mutex-1.0-openblas` and `blas_mutex-1.0-mkl`.
@@ -226,7 +226,8 @@ package:
 requirements:
   # any package depending on openblas should also depend on the correct blas_mutex package
   run_export:
-    - blas_mutex * openblas
+    # Add a run export on _any_ version of the blas_mutex package whose build string starts with "openblas"
+    - blas_mutex * openblas*
 ```
 
 And then the recipe of a package that wants to build two variants, one for `openblas` and one for `mkl` could look like this:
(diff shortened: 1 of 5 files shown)


docs/variants.md
```diff
@@ -154,3 +154,93 @@ requirements:
...
+
+And then the `recipe.yaml` for the `mutex` package could look like this:
+
+```yaml title="recipe.yaml" hl_lines="6 9"
+package:
+  name: blas_mutex
+  version: 1.0
+
+build:
+  string: ${{ blas_variant }}
```
pavelzw, line 208: How about using `${{ blas_variant }}${{ hash }}_${{ build_number }}` to not encourage users to create packages without hashes in the build string?

docs/variants.md
```diff
@@ -154,3 +154,93 @@ requirements:
...
+  variant_config:
+    # make sure that `openblas` is preferred over `mkl`
+    down_prioritize_variant: ${{ -1 if blas_variant == "mkl" else 0}}
```
pavelzw, line 211:
```suggestion
    down_prioritize_variant: ${{ -1 if blas_variant == "mkl" else 0 }}
```

docs/variants.md
```diff
@@ -154,3 +154,93 @@ requirements:
...
+
+```yaml title="recipe.yaml" hl_lines="2 8"
+package:
+  name: openblas
+  version: 1.0
+
+requirements:
+  # any package depending on openblas should also depend on the correct blas_mutex package
+  run_export:
+    - blas_mutex * openblas
```
pavelzw, line 229: This looks like a multiplication 😅 is `blas_mutex *openblas` also okay?

docs/variants.md
```diff
@@ -154,3 +154,93 @@ requirements:
...
+  name: fastnumerics
+  version: 1.0
+
+requirements:
+  host:
+    # build against both openblas and mkl
+    - ${{ blas_variant }}
+  run:
+    # implicitly adds the correct blas_mutex package through run exports
+    # - blas_mutex * ${{ blas_variant }}
```
pavelzw, line 245:
```suggestion
    # - blas_mutex *${{ blas_variant }}
```

See above

pavelzw (approved):
Awesome! Some minor comments

docs/variants.md
```diff
@@ -154,3 +154,93 @@ requirements:
...
+The `down_prioritize_variant` option allows you to specify a variant that should be _down-prioritized_. For example:
+
+```yaml title="recipe.yaml" hl_lines="7"
+build:
+  variant_config:
+    use_keys:
+      # use cuda from the variant config, e.g. to build multiple CUDA variants
+      - cuda
+    # this will down-prioritize the cuda variant versus other variants of the package
+    down_prioritize_variant: ${{ -1 if cuda else 0 }}
```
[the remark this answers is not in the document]
wolfv: Yeah right now it doesn't actually matter because we take the absolute value. Maybe it should be positive values only?

docs/variants.md
wolfv: The `*` is for the version.

docs/variants.md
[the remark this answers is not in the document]
pavelzw: I am in favor of positive values only 👍🏻

docs/variants.md
pavelzw: Ah so the default way of specifying dependencies is `- package-name version build_str`?

wolfv: yep!

PR #633, opened 2024-02-13
docs: Add github action to documentation

pavelzw:
I'm wondering whether we should symlink `index.md` and `README.md`...


wolfv:
Yeah if symlinking is possible, that would be nice :)

Issue #399, opened 2023-12-06
improve partialerror messages/labels


swarnimarun:
All the below locations don't provide any meaningful labels for diagnostics produced by miette.

Is very easy to fix, eg https://github.com/prefix-dev/rattler-build/pull/398

```sh
❯ rg "_partialerror\!\((.*)\)" --glob=\!src/recipe/error.rs
src/variant_config.rs
54:            .ok_or_else(|| _partialerror!(*self.span(), ErrorKind::ExpectedMapping,))
669:            .ok_or_else(|| _partialerror!(*self.span(), ErrorKind::ExpectedMapping))

src/recipe/parser.rs
85:            .ok_or_else(|| _partialerror!(*root_node.span(), ErrorKind::ExpectedMapping,))?;

src/recipe/custom_yaml.rs
1072:        Url::parse(self.as_str()).map_err(|err| _partialerror!(*self.span(), ErrorKind::from(err)))

src/recipe/parser/test.rs
76:                Err(_partialerror!(*self.span(), ErrorKind::ExpectedMapping,))?
157:                Err(_partialerror!(*self.span(), ErrorKind::ExpectedMapping,))?

src/recipe/parser/package.rs
38:            .ok_or_else(|| _partialerror!(*self.span(), ErrorKind::ExpectedMapping,))
110:            .ok_or_else(|| _partialerror!(*self.span(), ErrorKind::ExpectedMapping,))
155:            .ok_or_else(|| _partialerror!(*self.span(), ErrorKind::ExpectedScalar))
163:            .map_err(|err| _partialerror!(*self.span(), ErrorKind::from(err),))

src/recipe/parser/build.rs
113:            .ok_or_else(|| _partialerror!(*self.span(), ErrorKind::ExpectedMapping))
225:            .ok_or_else(|| _partialerror!(*self.span(), ErrorKind::ExpectedMapping))
409:            .ok_or_else(|| _partialerror!(*self.span(), ErrorKind::ExpectedScalar,))?
435:            .ok_or_else(|| _partialerror!(*self.span(), ErrorKind::ExpectedScalar))
443:            .map_err(|err| _partialerror!(*self.span(), ErrorKind::EntryPointParsing(err),))

src/recipe/parser/about.rs
91:            .ok_or_else(|| _partialerror!(*self.span(), ErrorKind::ExpectedMapping,))
167:            .ok_or_else(|| _partialerror!(*self.span(), ErrorKind::ExpectedScalar,))
[...]

PR #634, opened 2024-02-13
fix: labels and help messages for partial errors for clarity

swarnimarun:
Some of the partial errors could use labels, this pr adds some of them in cases where I thought they would improve clarity especially in edge cases where spans and error messages aren't clear enough.

fixes https://github.com/prefix-dev/rattler-build/issues/399


wolfv:
I've opened #805

PR #641, opened 2024-02-17
refactor `script` execution during build

wolfv:


wolfv:
It's almost good - the remaining problem is that the test script files (`conda_buidl.bat`) is written to the package cache. Will fix that tomorrow.

PR #677, opened 2024-02-26
feat: support scp-style SSH urls for Git source

### LINKED ISSUE #672: Make sure that `git@github.com:blabla` urls' work in git source
The SSH style URLs don't work because we don't parse them correctly. We might have to introduce a third kind of URL (URL, SSH_URL and Path) when parsign a git source.

orhun:
We now support:

```yml
source:
  git: git@github.com:killercup/cargo-edit.git
```

Todo list:

- [ ] Update documentation about this new key
- [ ] Update one of the examples to use this

closes #672

### DIFF
--- a/src/recipe/parser/snapshots/rattler_build__recipe__parser__source__tests__ssh_serialization.snap
+++ b/src/recipe/parser/snapshots/rattler_build__recipe__parser__source__tests__ssh_serialization.snap
@@ -1,0 +1,5 @@
+---
+source: src/recipe/parser/source.rs
+expression: yaml
+---
+git: git@github.com:prefix-dev/rattler-build.git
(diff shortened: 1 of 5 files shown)


baszalmstra:
Nice! We also need to add this to the https://github.com/prefix-dev/recipe-format

wolfv:
I think we don't need to update examples or even docs. But it would be good to use this in one of the end-to-end tests or another test. And the lint needs fixing. Then I am happy to merge!

orhun:
Looks like there is a problem with using SSH URLs in the CI:

```
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.
```

orhun:
We need to add an SSH key to clone over SSH 💀 

<https://stackoverflow.com/questions/60222741/github-actions-and-git-clone-issue>

wolfv:
Hmm, well, if there is no quick way of testing that it might not matter that much. We could add a test that just makes sure that we can parse these.

wolfv (approved):

PR #679, opened 2024-02-27
chore: add git-cliff config

orhun:
This PR adds the `git-cliff` config (the format is the same as `pixi`)

P.S. Highlights is empty on purpose since we want to write a summary there.


orhun:
Usage:

```bash
pixi global install git-cliff
git cliff --unreleased
```

PR #723, opened 2024-03-22
fix: do not take python variant when noarch:python

nichmor:
When having noarch: python in recipe 
```
requirements:
  host:

    - python
    - pip
    - setuptools
  run:
    - pip
```

and also variants.yaml 
```
python:
  # part of a zip_keys: python, python_impl, numpy
  - 3.8.* *_cpython
  - 3.9.* *_cpython

python_impl:
  # part of a zip_keys: python, python_impl, numpy
  - cpython
  - cpython
numpy:
  # part of a zip_keys: python, python_impl, numpy
  - 1.22
  - 1.22
```

rattler-build should render only one recipe and don't take into account python variants ( like how conda render do )

### DIFF
--- a/src/snapshots/rattler_build__variant_config__tests__python_is_not_used_as_variant_when_noarch.snap
+++ b/src/snapshots/rattler_build__variant_config__tests__python_is_not_used_as_variant_when_noarch.snap
@@ -1,0 +1,5 @@
+---
+source: src/variant_config.rs
+expression: used_variables_all
+---
+- target_platform: noarch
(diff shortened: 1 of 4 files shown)


wolfv:
Cool! Can you add a test of some sort? Even if it's just a snapshot test of the rendering or something like that. 

I think the `variant_config.rs` file deserves to be refactored some day but I am OK with merging this for now :)

nichmor:
> Cool! Can you add a test of some sort? Even if it's just a snapshot test of the rendering or something like that.
> 
> I think the `variant_config.rs` file deserves to be refactored some day but I am OK with merging this for now :)

I've added the test for it

orhun (approved):

wolfv (approved):

PR #724, opened 2024-03-22
chore: remove unused dependencies

wolfv:


Cargo.toml
```diff
@@ -81,7 +80,6 @@ spdx = "0.10.4"
 fs_extra = "1.3.0"
 ignore = "0.4.22"
 globset = "0.4.14"
-tracing-indicatif = "0.3.6"
```
nichmor, line 84: based on tests failure I think we need indicatif

wolfv: thanks!

PR #728, opened 2024-03-23
fix: sorting topological on empty output

nichmor:


.github/workflows/rust.yml
```diff
@@ -47,6 +47,10 @@ jobs:
         run: cargo build --verbose
       - name: Run tests
         run: cargo test --all-features --verbose -- --nocapture
+      - name: Test run of rendering a recipe
+        run: |
+          cd examples/rich
+          cargo run -- build --render-only --output-dir output
```
orhun, line 53: We can probably add this as a E2E test in `rust-tests/`

nichmor: Yeah I agree - It was just faster for me to validate the error.
Thanks for mentioning it

orhun (approved):

Issue #716, opened 2024-03-18
rattler-build error when using --target-platform linux-64 on osx-arm64


nichmor:
Having this `recipe.yaml` in a `recipe` folder
```
# yaml-language-server: $schema=https://raw.githubusercontent.com/prefix-dev/recipe-format/main/schema.json

context:
  version: "13.4.2"


package:
  name: "boltons"
  version: ${{ version }}

source:
  url: https://github.com/mahmoud/boltons/archive/${{ version }}.tar.gz
  sha256: 1843160500f560802b5240adf8073e35112d17a4ca02581b5c45bb571ded4df7

build:
  script: 
    - python -m pip install . --no-deps -vv

requirements:
  host:
    - pip
    - python >=3.7
    - setuptools
  run:
    - python >=3.7

```

Running `rattler-build build --render-only --recipe recipe --target-platform linux-64`

will output this error:

```
thread 'tokio-runtime-worker' panicked at /Users/runner/miniforge3/conda-bld/rattler-build_1709722974673/_build_env/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.36.0/src/runtime/time/entry.rs:556:9:
A Tokio 1.x context was found, but it is being shutdown.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error:   × Failed to resolve dependencies: failed to link 'share/terminfo/N/NCR260VT300WPP'
  ├─▶ failed to link 'share/terminfo/N/NCR260VT300WPP'
  ├─▶ could not open source file
  ╰─▶ Invalid argument (os error 22)
```

and running `rattler-build build --render-only --recipe recipe --target-platform win-64` this:

```
thread 'tokio-runtime-worker' panicked at /Users/runner/miniforge3/conda-bld/rattler-build_1709722974673/_build_env/.cargo/git/checkouts/rattler-3fdeff0ba0214908/82dc20f/crates/rattler/src/install/entry_point.rs:18:14:
unsupported platform
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at /Users/runner/miniforge3/conda-bld/rattler-build_1709722974673/_build_env/.cargo/git/checkouts/rattler-3fdeff0ba0214908/82dc20f/crates/rattler/src/install/driver.rs:139:13:
[...]

PR #729, opened 2024-03-24
feat: --render-only will make a dry-run and will not install resolved packages

nichmor:
fixes this issue: #716

### DIFF
--- a/rust-tests/src/lib.rs
+++ b/rust-tests/src/lib.rs
@@ -616,4 +616,24 @@ mod tests {
         let output = String::from_utf8(rattler_build.stdout).unwrap();
         assert!(output.contains("it is included in the allow list. Skipping..."));
     }
+
+    #[test]
+    #[cfg(target_os = "macos")]
+    fn render_only_dont_install_pkgs() {
+        let tmp = tmp("test_render_only");
+
+        let rattler_build = rattler().with_args([
+            "build",
+            "--recipe",
+            recipes().join("flask").to_str().unwrap(),
+            "--render-only",
+            "--output-dir",
+            tmp.as_dir().to_str().unwrap(),
+            "--target-platform",
+            "win-64"
+
+        ]);
+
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -589,10 +591,10 @@ pub fn sort_build_outputs_topologically(
         .iter()
         .map(|idx| &outputs[idx.index()])
         .for_each(|output| {
-            tracing::info!(
-                "Ordered output: {:?}",
-                output.outputs[0].name().as_normalized()
-            );
+            // tracing::info!(
```
nichmor, line 594: will be uncommented when https://github.com/prefix-dev/rattler-build/pull/728
is merged

rust-tests/src/lib.rs
```diff
@@ -629,4 +629,24 @@ mod tests {
 
         assert!(rattler_build_render_only.status.success());
     }
+
+    #[test]
+    #[cfg(target_os = "macos")]
```
orhun, line 634: Only macos? :)

src/render/solver.rs
```diff
@@ -195,14 +195,18 @@ pub async fn create_environment(
...
+        install_packages(
+            &required_packages,
+            target_platform,
+            target_prefix,
+            &cache_dir,
+            tool_configuration,
+        )
+        .await?;
+    } else {
+        tracing::info!("skipping installation when --render-only is used",);
```
orhun, line 208: nit:
```suggestion
        tracing::info!("skipping installation when --render-only is used");
```

rust-tests/src/lib.rs
nichmor: currently this issue can only be reproduced on macos when using --target-platform win-64 ( because macos has insensitive file system ) - so I've decided to keep this test to this only scope

orhun (approved):
LGTM

PR #735, opened 2024-03-26
feat: support reading recipe from stdin

### LINKED ISSUE #734: Ability to pipe a recipe file into `rattler-build build` command
It would be nice if I could pipe a recipe file to the `rattler-build build` command so I could avoid writing to disk.

For example, using the [recipe conversion tool](https://github.com/conda-incubator/conda-recipe-manager/blob/main/conda_recipe_manager/commands/convert.py) in a CI environment, I could do something like:

```sh
convert meta.yaml | rattler-build build --render-only
```

This might also be useful for other/future recipe generation tools.

orhun:
closes #734

This is not a bulletproof implementation but it does the job:

```sh
cat examples/cargo-edit/recipe.yaml | rattler-build build --render-only
```

Couple of things:

- We heavily depend on recipe path during our build process and it is also required for the _reload_ functionality in TUI - that's why I decided to create a temp dir along with a `recipe.yaml` when stdin is used.
- In the cases where the build requires more than `recipe.yaml` this will fail. But that's fine because it is intended to pass only one file through stdin.

Either way, I think it is a neat feature! Let me know what you think.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -130,7 +130,7 @@ pub fn get_tool_config(
 /// Returns the output for the build.
 pub async fn get_build_output(
     args: &BuildOpts,
-    recipe_path: PathBuf,
+    recipe_path: &Path,
     tool_config: &Configuration,
 ) -> miette::Result<Vec<Output>> {
     let output_dir = args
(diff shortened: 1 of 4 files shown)


schuylermartin45 (commented):
Wow that was a fast turn-around! As a C/C++ guy currently learning Rust, this looks good but I don't feel confident enough to approve this right now.

I also am unclear if taking recipes from `STDIN` might complicate any future project goals/have unintended consequences.

Either way, I appreciate the effort and turn-around on this! 🚀

wolfv (approved):

PR #746, opened 2024-03-29
fix: take into consideration non-canonical variant also

nichmor:
when having in `recipe` requirements:
`-r-base`

and in `variant_config` 
```
r_base:
"4.2"
```

it will not be take into consideration as variant.

In  [ conda_build_config ](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/main/recipe/conda_build_config.yaml#L812)
they store it as underscore so my idea is to check both r_base and r-base if it's in used vars


wolfv:
As discussed on Discord, we might want to make a wrapper-struct instead for the variants hash-table that normalizes the keys on insertion / retrieval.

wolfv (approved):

wolfv (approved):

PR #750, opened 2024-04-02
style: improve spdx license error

### LINKED ISSUE #675: Unrecognized spdx - add hint about `LicenseRef`
We could add some help text when parsing of a SPDX fails. If it fails because it's not recognized, we should hint at the list of spdx (web URL) and print something about `LicenseRef-Custom` for cuomst licenses.

orhun:
```
Error:   × Parsing: failed to parse SPDX license: invalid character(s)
  │ See <https://spdx.org/licenses> for the list of valid licenses.
  │ Use 'LicenseRef-<MyLicense>' if you are using a custom license.
```

closes #675


wolfv:
LGTM! Some outdated snapshots I think.

PR #758, opened 2024-04-03
feat: add --with-solve option

nichmor:


src/opt.rs
```diff
@@ -217,6 +217,10 @@ pub struct BuildOpts {
     #[arg(long)]
     pub render_only: bool,
 
+    /// Render the recipe files with solving requirements.
+    #[arg(long, requires("render_only"))]
```
wolfv, line 221: nice, didn't know about the `requires`.

nichmor: I was super surprised to see that it's present out-of-the-box in clapy. 
In Click you need to write your own subclass to implement it :(

wolfv (approved):

PR #765, opened 2024-04-04
feat: implement faster skip-existing

wolfv:


src/opt.rs
```diff
@@ -259,8 +260,8 @@ pub struct BuildOpts {
     pub tui: bool,
 
     /// Wether to skip packages that already exist in any channel
-    #[arg(long, default_value = "false")]
-    pub skip_existing: bool,
+    #[arg(long)]
```
nichmor, line 263: can we do something like this here:
`#[clap(value_enum, default_value_t=SkipExisting::None)]`

so we can skip using unwrap_or here ? `args.skip_existing.unwrap_or(SkipExisting::None),`

wdyt?

nichmor (approved):

src/opt.rs
wolfv: I just made it the way I wanted it to be:

```
nothing -> None
--skip-existing -> skip local builds
--skip-existing=[local, all, none] -> use hte option
```

:)

nichmor: perfect!

PR #768, opened 2024-04-04
feat: set environment variables for package tests

### LINKED ISSUE #764: `PKG_VERSION` not set during test
Just saw that this failed 

```
    - test "$(cppcheck --version)" == "Cppcheck $PKG_VERSION"
```

orhun:
fixes #764

will follow up with an example

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -11,7 +11,7 @@ use crate::metadata::Output;
 use crate::package_test::TestConfiguration;
 use crate::recipe::parser::TestType;
 use crate::render::solver::load_repodatas;
-use crate::{package_test, tool_configuration};
+use crate::{env_vars, package_test, tool_configuration};
 
 /// Check if the build should be skipped because it already exists in any of the channels
 pub async fn skip_existing(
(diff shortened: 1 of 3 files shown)


wolfv:
this is not going to work for "standalone" tests though (e.g. when runnign with `rattler-build test -p ./mypkg.tar.bz2`).

We need to extract the info either from the filename or from the info.json.

orhun:
> We need to extract the info either from the filename or from the index.json.

`env_vars::vars` currently requires and `Output` to construct the the environment variables.

Is there an easy way (helper function?) of getting an `Output` from a package file? How can we achieve that?

wolfv:
No there is no easy way, but for the few variables we can just make a separate function

- PKG_VERSION
- PKG_NAME
- PKG_BUILD_STRING
- PKG_BUILD_NUMBER

are the ones that come to mind. They are readily available in the `index.json` and we are already parsing that file at teh beginning of the test execution IIRC.

orhun:
Alright that makes sense, will add a function to add the available variables.

PR #806, opened 2024-04-17
use env_clear

wolfv:
cc @DerThorsten maybe that would help with the issues you had regarding deactivation. This should start a more isolated shell for building & testing.

One can still inject environment variables "from the outside" by using the `env` part of the script object, and the `${{ env.get(...) }}` to grab them from the current process.


baszalmstra:
Nice! This makes sure things dont accidentally leak into the build environment!

PR #822, opened 2024-04-23
fix: correctly identify tarballs (#821)

### LINKED ISSUE #821: Faulty tarball detection
So the issue is painfully simple.

Rattler extracts tarballs if and only if their source [url](https://github.com/prefix-dev/rattler-build/blob/31017605d1e5a3acf9881668cbe998deded33477/src/source/mod.rs#L173-L179)/[path](https://github.com/prefix-dev/rattler-build/blob/31017605d1e5a3acf9881668cbe998deded33477/src/source/mod.rs#L233-L239) string contains ".tar".

This is obviously an issue in case of extensions like `.tgz`, `.txz`, `.tbz2`, `.tzst` etc.
Rattler even already knows how to detect compression method for such files - [src/source/extract.rs#L22-L41](https://github.com/prefix-dev/rattler-build/blob/31017605d1e5a3acf9881668cbe998deded33477/src/source/extract.rs#L22-L41). It's just it never gets to this point :(

michalsieron:
This is an initial proposal how to fix #821 with relatively low effort.
Maybe a better idea would be to somehow avoid duplicating strings between this new `is_tarball` function and `ext_to_compression`, but I don't have enough Rust knowledge, and I am completely new to rattler codebase.

### DIFF
--- a/src/source/extract.rs
+++ b/src/source/extract.rs
@@ -19,6 +19,40 @@ enum TarCompression<'a> {
     Lzop,
 }
 
+/// Checks whether file has known tarball extension
+pub fn is_tarball(file_name: &str) -> bool {
+    [
+        // Gzip
+        ".tar.gz",
+        ".tgz",
+        ".taz",
+        // Bzip2
+        ".tar.bz2",
+        ".tbz",
+        ".tbz2",
+        ".tz2",
+        // Xz2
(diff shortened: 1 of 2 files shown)


wolfv:
Awesome, thanks!

PR #827, opened 2024-04-27
feat: make binary smaller size

nichmor:
based on suggestions from: 
https://github.com/johnthagen/min-sized-rust
https://arusahni.net/blog/2020/03/optimizing-rust-binary-size.html
and pixi workflow

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -91,18 +91,18 @@ jobs:
 
       # Enable Link Time Optimization (LTO) for our release builds. This increases link time but drastically reduces
       # binary size.
-      CARGO_PROFILE_RELEASE_LTO: ${{ 'true' }}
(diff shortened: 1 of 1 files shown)


baszalmstra:
Nice! Curious about the difference!

nichmor:
> Nice! Curious about the difference!

comparing to latest build from main using these optimisations

we are saving **5.9mb** for x86_64-unknown-linux-musl

wolfv:
That's great!

wolfv:
I think we should only use these on release though, because usually they make the pipeline slower to run, right?

nichmor:
> I think we should only use these on release though, because usually they make the pipeline slower to run, right?

yes! I removed the if-guard to test it on my PR only

nichmor:
> I think we should only use these on release though, because usually they make the pipeline slower to run, right?

I've added back git.ref guard

PR #857, opened 2024-05-24
feat: clean env more agressively and run bash with --norc --noprofile

wolfv:
I had issues when running the build process from within a `pixi run`.

This fixes that, but it's also pretty aggressive. However, we do copy a lot of environment variables from the running system into the `build_env.sh` script anyways so probably this is fine.


wolfv:
Hmm, unfortunately, this gives me some headaches. On Windows we definitely need some env vars, and I see pretty weird failures inside `pytest`.

Windows reference: https://github.com/rust-lang/rust/issues/114737

PR #875, opened 2024-05-27
refactor: how dependencies are stored

baszalmstra:
Fixes 863
Fixes 865

I think we should also update the CEP.

### DIFF
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -633,7 +633,7 @@ mod test {
     use rattler_digest::{parse_digest_from_hex, Md5, Sha256};
     use url::Url;
 
-    use crate::render::resolved_dependencies::{self, DependencyInfo};
+    use crate::render::resolved_dependencies::{self, SourceDependency};
 
     use super::{Directories, Output};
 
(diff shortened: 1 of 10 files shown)


wolfv:
Does this enable using the `${{ compiler('c') }}` function e.g. in 

```
build:
  ignore_run_exports:
    from_package:
      - ${{ compiler('c') }}
```

I also just double checked and saw that someone uses the jinja function in the `test` section on conda-forge (tricky, especially when cross-testing ...).

However, I think we can ignore this case in the tests (at least for now) and ask users to use `compiler-c` "meta-package" on conda-forge in the tests. 

I didn't see any other uses except for in the requirements sections or ignore_run_export section on conda-forge so far.

baszalmstra:
> Does this enable using the ${{ compiler('c') }} function e.g. in

It does not, it only modifies the serialization behavior in the rendered recipe I think.

baszalmstra:
I was modifying the CEP to reflect these changes but it raises a question. This is the example that is in the CEP but modified to reflect the changes from this PR:

```yaml
- compile: c  # language specified in ${{ compiler(`c`) }}
  spec: clang_osx-arm64
- source: make >=1.3
- pin_compatible: quarto >=1.4.550,<1.5
- pin_subpackage: some-pkg-a >=1.0.0,<2
- run_export: zlib
  spec: libzlib >=1.3.1,<1.4.0a0
  from: host # environment that exported this dependency
- variant: setuptools # variant coming from the variant configuration
  spec: setuptools 69.*
```

Do I understand `variant` currently though? Previously the example for variant was:

```yaml
- source: variant
  spec: setuptools 69.*
  variant: "69" # variant coming from the variant configuration 
```

In this case the variant field is the `key` of the variant instead of the value which I assumed was the case. @wolfv Can you clarify?

wolfv:
Hmm, yeah, I don't think it matters much either way. Since we also store the variant configuration, it should be pretty easy to map it back by key.

PR #882, opened 2024-05-27
feat: remove hash_input from rendered recipe

### LINKED ISSUE #867: get rid of `hash_input` in rendered recipe

baszalmstra:
Fixes #867

Changes:

* Removes `hash_input` from rendered recipe.
* Renames `hash.hash_prefix` to `hash.prefix`.

Modification to the CEP: [conda/ceps@`0cdf2f1` (#74)](https://github.com/conda/ceps/pull/74/commits/0cdf2f11426064e25b3c59fda9387776648540fc)

### DIFF
--- a/src/hash.rs
+++ b/src/hash.rs
@@ -84,16 +84,42 @@ pub struct HashInfo {
     /// The hash (first 7 letters of the sha1sum)
     pub hash: String,
 
-    /// The exact input that was used to compute the hash
-    pub hash_input: String,
-
     /// The hash prefix (e.g. `py38` or `np111`)
-    pub hash_prefix: String,
+    #[serde(default, skip_serializing_if = "String::is_empty")]
+    pub prefix: String,
+}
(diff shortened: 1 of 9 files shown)


src/hash.rs
```diff
@@ -84,16 +84,42 @@ pub struct HashInfo {
...
+    pub prefix: String,
+}
+
+/// Represents the input to compute the hash
+pub struct HashInput(String);
+
+impl HashInput {
+    /// Create a new hash input from a variant
+    pub fn from_variant(variant: &BTreeMap<String, String>) -> Self {
+        let mut buf = Vec::new();
```
nichmor, line 98: do you think it's worth to make vec with capacity of BtreeMAP keys?

baszalmstra: I doubt that will increase performance significantly, and its also not a critical path. Id opt for simplicity instead.

PR #902, opened 2024-06-03
feat: make `jinja` filters and tests explicit, rename `cmp` to `match`

wolfv:
Also adds the `split` filter which is pretty useful for versions. 

```
${{ "1.2.3" | split(".") | ... | join(".") }}
```


ruben-arts (approved):
Thanks for fixing that!

PR #912, opened 2024-06-06
docs: Improve documentation `selectors` and `installation`

ruben-arts:
This PR slightly improves the documentation some more. 

Next steps would be making a good "getting started"


docs/selectors.md
```diff
@@ -61,41 +61,94 @@ if: linux and (aarch64 or ppc64le)) ... # true for linux ppc64le or linux-aarch6
...
+    - if: python == "3.8" # (1)!
+      then: mydep
+      else: otherdep
+```
+
+1. This will only add `mydep` when the Python version is 3.8. This comparison is a string comparison, so it is important
+   to
+   use the correct string representation of the variant.
+
 ### The `cmp` function
```
wolfv, line 112: Ah, this function was renamed to `match` actually.

PR #914, opened 2024-06-07
fix: add test for channel specific matchspec 

### LINKED ISSUE #706: Can't force a dependency to come from a particular channel
In conda-build / boa you can force a dependency to come from a particular channel by specifying it with the matchspec

```
my-special-channel::packagename =1.5.4
```

I tried doing the same with rattler-build and it does not seem to work

Ideally we should be able to specify a dependency in the expanded form without needing to resort to the esoteric matchspec stuff

```yaml
 - if: unix
   then:
      name: packagename
      version: 1.5.4
      channel: my-special-channel

```

wolfv:
fixes #706


wolfv:
This actually just adds a test, I think it was fixed by the repodata gateway.

PR #920, opened 2024-06-11
fix: preserve permissions when streaming conda or tarbz2

nichmor:


src/rebuild.rs
```diff
@@ -55,8 +57,7 @@ fn folder_from_conda(
                     std::fs::create_dir_all(parent_folder)?;
                 }
             }
-            let mut dest_file = std::fs::File::create(dest_file)?;
```
nichmor, line 58: we had 2 options of fixing this:
1) either we continue using io::copy but set separtly permisions
2) or we are using entry.unpack, which will do it because of set_preserve_permissions(true);

I've decided to use 2'nd approach :)

wolfv: Looks great! Didn't know about the existence of entry.unpack 👍

PR #928, opened 2024-06-17
Adds commonly used go compilers to default compiler logic

schuylermartin45:
As of `0.17.0` (with the introduction of the checks for `compiler()` function usage), the integration tests for `conda-recipe-manager` showed some failures with `go` projects in `conda-forge`:

```
 ╭─ Finding outputs from recipe
 │
Error:   × Failed to parse recipe

Error:   × Parsing: failed to render Jinja expression: undefined value: No compiler found for language: go-nocgo
  │ You should add `go-nocgo_compiler` to your variant config file. (in <string>:1)
    ╭─[22:7]
 21 │   build:
 22 │     - ${{ compiler('go-nocgo') }}
    ·       ─────────────┬─────────────
    ·                    ╰─┤ undefined value: No compiler found for language: go-nocgo
    ·                      │ You should add `go-nocgo_compiler` to your variant config file.
 23 │     - if: unix
    ╰────
```

This PR attempts to address this issue by adding default compilers for golang.

I verified this locally. I am able to dry-run build recipe files that use `compiler('go')` and `compiler('go-nocgo')` with these edits.

### DIFF
--- a/src/recipe/jinja.rs
+++ b/src/recipe/jinja.rs
@@ -188,36 +188,34 @@ fn jinja_pin_function(
 }
 
 fn default_compiler(platform: Platform, language: &str) -> Option<String> {
-    if platform.is_windows() {
-        match language {
-            "c" => Some("vs2017"),
-            "cxx" => Some("vs2017"),
-            "fortran" => Some("gfortran"),
-            "rust" => Some("rust"),
-            "go" => Some("go"),
-            "go-nocgo" => Some("go-nocgo"),
-            _ => None,
-        }
-    } else if platform.is_osx() {
-        match language {
-            "c" => Some("clang"),
-            "cxx" => Some("clangxx"),
-            "fortran" => Some("gfortran"),
-            "rust" => Some("rust"),
-            "go" => Some("go"),
-            "go-nocgo" => Some("go-nocgo"),
-            _ => None,
-        }
-    } else {
-        match language {
-            "c" => Some("gcc"),
-            "cxx" => Some("gxx"),
-            "fortran" => Some("gfortran"),
-            "rust" => Some("rust"),
-            "go" => Some("go"),
-            "go-nocgo" => Some("go-nocgo"),
-            _ => None,
-        }
+    match language {
+        // Platform agnostic compilers
+        "fortran" => Some("gfortran"),
+        "rust" => Some("rust"),
+        "go" => Some("go"),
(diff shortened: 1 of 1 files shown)


baszalmstra:
Seems like there is a lotnof repetition in this code per platform type. Perhaps we could refactor that a little. Not perse required for this PR! But if you wanna give that a go, that might be nice!

schuylermartin45:
> Seems like there is a lotnof repetition in this code per platform type. Perhaps we could refactor that a little. Not perse required for this PR! But if you wanna give that a go, that might be nice!

Agreed. Thought about that, but didn't want to be too aggressive in my first rattler PR :P

schuylermartin45:
@baszalmstra I'll give it a quick shot.

baszalmstra (approved):
Thanks! Appreciate the cleanup!

wolfv:
I am also a big fan of the cleanup, but I am unsure with regards to adding more defaults. I think `conda-build` does not ship these by default, does it?

schuylermartin45:
> I am also a big fan of the cleanup, but I am unsure with regards to adding more defaults. I think `conda-build` does not ship these by default, does it?

To be fair, `conda-build` doesn't support `rust` as a default either: https://github.com/conda/conda-build/blob/4457362847b002ed4f6ce5c77a3f05f9325b449f/conda_build/variants.py#L61

I would argue both `rust` and `go` are popular enough to warrant inclusion, and, as far as I know, there aren't many (if any) mainstream alternative compilers for these languages.

At the end of the day though, all I care about it is that recipes that use `compiler('go')` pass the dry-run build, so if there's another way to address this issue, we can go another route.

schuylermartin45:
> Regarding `rust` - point taken!
> 
> I think we can do either:
> 
> * add go compilers
> * add a default "compiler-name" forward behavior, so that `rust => rust` and `go => go` etc.

I would say let's add `go` for now and see how many more languages show up. If it's a lot, the forwarding idea is probably worth it as this list will be a pain to maintain. If it isn't a lot, then we provide some amount of validation/typo-protection to the user.

As a quick search in the 13k recipes in `conda-recipe-manager-test-data`, I'm seeing about 44 instances of `compiler("rust")` and about 37 instances of `compiler("go")/compiler("go-nocgo")`. Anything that isn't already covered looks like a one-off and there aren't all that many one-offs.

PR #946, opened 2024-06-25
fix: fix a serialization issue where a single-line script and cwd are set

wolfv:
I overlooked a condition where we have a simple (single-line) script and a `cwd` (as injected by the test serialization. 

Under these conditions, we did not render out the `cwd` and thus the test execution stopped working.

This PR has the fix.


schuylermartin45 (approved):
LGTM!

PR #953, opened 2024-06-28
feat: improvements to the R generator

wolfv:
- A build-string in R packages is replaced by just a number (e.g. 3.4-5 becomes 3.4.5)
- I added a few more licenses
- A better list of all the builtins
- Only the first URL is inserted as Homepage

cc @roaldarbol


roaldarbol:
Here's a list of commonly used licenses on CRAN with counts for the first 10 most used ones, I think there's a couple more you can add. 😊 https://thinkr-open.github.io/licensing-r/cran-licenses.html

PR #961, opened 2024-07-05
fix: cran recursive recipe generation fixes

wolfv:


src/recipe_generator/pypi.rs
```diff
@@ -34,6 +34,14 @@ pub struct PyPIOpts {
     /// Whether to write the recipe to a folder
     #[arg(short, long)]
     pub write: bool,
+
+    /// Wether to use the conda-forge PyPI name mapping
```
Hofer-Julian, line 38:
```suggestion
    /// Whether to use the conda-forge PyPI name mapping
```

Hofer-Julian (commented):
Small typo

PR #967, opened 2024-07-11
chore: fix clippy warnings

Hofer-Julian:
Also remove `paths` filter for CI.
It doesn't play well with concurrency.
If you add a job that doesn't touch the correct files, it will cancel other jobs, but not start a new one.


nichmor (approved):
looks good!

baszalmstra (approved):

PR #980, opened 2024-07-22
feat: use implicit variants.yaml

### LINKED ISSUE #974: Parse `variants.yaml` per recipe when using `--recipe-dir`
Currently, variants can only be selected globally using the `--variant-config` flag. However, sometimes a per-recipe variant configuration file can be quite useful.

I propose that for each recipe found when using `--recipe-dir` we also read a `variants.yaml` file in the same directory to allow per recipe overrides.

baszalmstra:
This PR adds reading from a `variants.yaml` file next to the `recipe.yaml` file without having to specify a config file on the command line.

This is especially useful when using the `--recipe-dir` option.

Some questions:
* This cannot be turned off atm. The `variants.yaml` file is always included. Is that what we want?
* Should this file be appended or prepended to the complete list of configurations.

Do not merge, I need to add tests.

Fixes #974


wolfv:
I think the configuration file "closest" to the recipe should win, but I am not sure what to do when we add variant files from the CLI as well ... e.g. `-m variant.yaml`. Should we then ignore them, and add another flag (`--use-variant-next-to-recipe`)?

wolfv:
Or with a glob somehow `--variant */variant.yaml`, but only use the globs that also match the recipe prefix ...

baszalmstra:
Personally Im more inclined to make this the default behavior and have the option to _opt out_ instead.

baszalmstra:
> Personally Im more inclined to make this the default behavior and have the option to _opt out_ instead.

Perhaps a `--ignore-recipe-variants` flag? (I hate more flags)

wolfv:
yes, flag works for me!

wolfv (approved):

PR #986, opened 2024-07-25
feat: include extra field in rendered recipe

Hofer-Julian:


baszalmstra (commented):
Should we use indexmap to preserve the order of keys?

Hofer-Julian:
> Should we use indexmap to preserve the order of keys?

Yes, sounds good

baszalmstra (approved):

wolfv:
Tested locally and works as expected :)

PR #988, opened 2024-07-25
prepare release 0.18.2

wolfv:


Hofer-Julian:
Maybe also a good time to update the toolchain

PR #989, opened 2024-07-25
fix: filter cache run_exports with the right ignores

### LINKED ISSUE #977: Cache run exports are propagating to outputs
I'm not sure if this is a bug or just me misreading the documentation. [The documentation on the experimental cache feature](https://prefix-dev.github.io/rattler-build/dev/multiple_output_cache/) states that

> If the cache has an "ignore run exports" section, than we apply those filters at the cache level.

Since the cache does not itself produce any outputs, I assumed that the above statement meant that ignoring run exports in the cache would propagate to any outputs. However, that does not appear to be the case. Given the recipe below:

```
recipe:
  version: 1.0.0

source:
  path: .

cache:
  build:
    script:
      content: "echo foo"
  requirements:
    ignore_run_exports:
      from_package:
        - numpy
    host:
      - numpy

outputs:
  - package:
      name: example
    build:
      script:
        content: "echo bar"
    requirements:
      # Why is this necessary?
      #ignore_run_exports:
      #  from_package:
      #    - numpy
      host:
        - python
```

I observe the following:
```
$ rattler-build build --experimental --no-build-id
$ cd output/linux-64
$ cph x example-1.0.0-hb0f4dca_0.conda
$ cat example-1.0.0-hb0f4dca_0/info/index.json
{
  "arch": "x86_64",
  "build": "hb0f4dca_0",
  "build_number": 0,
  "depends": [
    "numpy >=1.19,<3",
    "python_abi 3.12.* *_cp312"
  ],
  "name": "example",
  "platform": "linux",
  "subdir": "linux-64",
  "timestamp": 1721520871303,
  "version": "1.0.0"
}%
```

If I include the commented out ignore run export in the package output section, then the numpy dependency is removed as I intended. Is that supposed to be necessary? If so, the documentation on the cache should be updated. However, I'm not sure what purpose ignoring run exports in the cache would serve if those run exports are propagated to the outputs.

wolfv:
this is an attempt to fix #977 

I think part of the issue was (at least!) that we didn't use the right "ignore_run_exports" (ie. not the one from the cache. 

My quick test showed that it still doesn't fix the issue, so will have to dig a little bit more tomorrow. But should be fixed swiftly!

### DIFF
--- a/src/recipe/parser/requirements.rs
+++ b/src/recipe/parser/requirements.rs
@@ -100,8 +100,13 @@ impl Requirements {
     }
 
     /// Get run exports that are ignored.
-    pub const fn ignore_run_exports(&self) -> &IgnoreRunExports {
-        &self.ignore_run_exports
+    pub fn ignore_run_exports(&self, merge: Option<&IgnoreRunExports>) -> IgnoreRunExports {
+        let mut ignore = self.ignore_run_exports.clone();
+        if let Some(merge) = merge {
+            ignore.by_name.extend(merge.by_name.iter().cloned());
+            ignore.from_package.extend(merge.from_package.iter().cloned());
+        }
+        ignore
     }
 
     /// Get all requirements at build time (combines build and host requirements)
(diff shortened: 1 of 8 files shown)


wolfv:
It's getting closer, but now filtering the run export from the cache in the output doesn't work anymore.

PR #1001, opened 2024-08-01
refactor: parse version as version

baszalmstra:
Small refactor to parse a version as a version instead of a string.


wolfv:
Great!

wolfv:
I am wondering if we should make it a `VersionWithSource` to not trip up users?

wolfv:
Judging by the snapshot errors it is a VersionWithSource :)

wolfv (approved):

PR #1006, opened 2024-08-04
fix: fix typo of feature recipe-generation

### LINKED ISSUE #1005: v0.19.0 lack generate-recipe subcommand
I think this should be related to https://github.com/prefix-dev/rattler-build/pull/996

today I installed `rattler-build` using `pixi global install rattler-build` and found that
image

Edit: 
sorry, when I built from source with `--features recipe-generation` the issue persists, and I've found that `recipe-generation` is a default-on feature, so there should be something to do in the code, not just the build.

will check later

TTTPOB:
Fixes #1005

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -132,7 +132,7 @@ async fn main() -> miette::Result<()> {
             .await
         }
         Some(SubCommands::Upload(upload_args)) => upload_from_args(upload_args).await,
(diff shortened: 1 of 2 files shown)


baszalmstra (approved):

baszalmstra:
Thanks! Good find!

PR #1020, opened 2024-08-15
refactor: make build string optional in recipe

baszalmstra:
The `build` string in the recipe was an optional value in name only. The code always assumed it has a value, which it always got during parsing. This PR changes this behavior so that the recipe datastructure actually contains an optional build but the code uses `Output::build_string()` to get the actual builds string which is either based on the hash or on the user-defined value.

This also allowed me to remove a bunch of unwraps.

I also reformatted some files with the nightly formatter.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -271,8 +271,8 @@ pub async fn get_build_output(
                 build_string: recipe
                     .build()
                     .string()
-                    .map(ToString::to_string)
-                    .unwrap_or_else(|| format!("{}_{}", hash, recipe.build().number())),
+                    .resolve(&hash, recipe.build().number())
+                    .into_owned(),
             },
         );
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -271,8 +271,8 @@ pub async fn get_build_output(
                 build_string: recipe
                     .build()
                     .string()
-                    .expect("Shouldn't be unset, needs major refactoring, for handling this better")
-                    .to_owned(),
+                    .map(ToString::to_string)
+                    .unwrap_or_else(|| format!("{}_{}", hash, recipe.build().number())),
             },
         );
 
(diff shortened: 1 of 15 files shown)


src/lib.rs
```diff
@@ -271,8 +271,8 @@ pub async fn get_build_output(
                 build_string: recipe
                     .build()
                     .string()
-                    .expect("Shouldn't be unset, needs major refactoring, for handling this better")
-                    .to_owned(),
+                    .map(ToString::to_string)
+                    .unwrap_or_else(|| format!("{}_{}", hash, recipe.build().number())),
```
Hofer-Julian, line 275: Maybe the logic that the default build string is hash joined with build number could be put into a function that is also used by the `build_string` function

src/metadata.rs
```diff
@@ -335,9 +345,18 @@ impl Output {
         self.recipe.package().version()
     }
 
-    /// The build string is usually set automatically as the hash of the variant configuration.
-    pub fn build_string(&self) -> Option<&str> {
-        self.recipe.build().string()
+    /// The build string is either the build string from the recipe or computed
+    /// from the hash and build number.
+    pub fn build_string(&self) -> Cow<'_, str> {
+        if let Some(build) = self.recipe.build().string() {
```
nichmor, line 351: could we use here unwrap_or?

Hofer-Julian (commented):
> I also reformatted some files with the nightly formatter.

Maybe we could use the nightly formatter once and then enforce it on the CI to avoid blowing up our diffs.

One comment, apart from that it looks good!

nichmor (approved):

src/lib.rs
baszalmstra: You are right. I changed the implementation to use a special type to make sure this cant go wrong in the future. 

Actually while doing this refactor I found a couple of different cases where this DID go wrong previously! Nice catch @Hofer-Julian

Hofer-Julian: Typestate pattern ❤️

PR #1023, opened 2024-08-16
add option to continue building on failure when using --recipe-dir

kerimbirgi:


KGB99:
I haven't figured out how to add the option to the CLI, any guidance is appreciated.

src/lib.rs
```diff
@@ -128,6 +128,7 @@ pub fn get_tool_config(
         use_bz2: args.common.use_bz2,
         render_only: args.render_only,
         skip_existing: args.skip_existing,
+        cont_on_fail: args.cont_on_fail,
```
baszalmstra, line 131: I would personally not specify this in the tool configuration but instead just pass a boolean to the `run_build_from_args` function.

src/opt.rs
```diff
@@ -342,6 +342,11 @@ pub struct BuildOpts {
     /// Extra metadata to include in about.json
     #[arg(long, value_parser = parse_key_val)]
     pub extra_meta: Option<Vec<(String, Value)>>,
+
+    /// Whether to continue building upon the failure of one recipe
+    /// Only relevant when building multiple recipes with `--recipe-dir`
+    #[arg(long, default_value = "false")]
+    pub cont_on_fail: bool,
```
baszalmstra, line 349: Can you make this argument a little more descriptive? Perhaps just `continue_on_failure`?

src/opt.rs
```diff
@@ -342,6 +342,11 @@ pub struct BuildOpts {
     /// Extra metadata to include in about.json
     #[arg(long, value_parser = parse_key_val)]
     pub extra_meta: Option<Vec<(String, Value)>>,
+
+    /// Whether to continue building upon the failure of one recipe
+    /// Only relevant when building multiple recipes with `--recipe-dir`
```
baszalmstra, line 347: You can enforce this by adding `requires = "recipe-dir"` to the `#[arg` section below.

baszalmstra (changes requested):
Do you still need help with the CLI? Looks like you already managed to make it work?

baszalmstra:
Also, can you make sure that the program still exists with an exit code if one of the ouputs failed to build?

KGB99:
> Do you still need help with the CLI? Looks like you already managed to make it work?

Yes it seems to work, thank you, i used the wrong command from the CLI myself, so I was confused.

> Also, can you make sure that the program still exists with an exit code if one of the ouputs failed to build?

This would be for a singular recipe right? I believe the latest commit includes that as well.

baszalmstra:
@wolfv Are you still interested in this PR? Should we upstream the changes?

PR #1030, opened 2024-08-23
refactor: seperate solving from installing

baszalmstra:
This PR splits solving the dependencies from the installation of the environments. This can make `--render-only` much faster because no installation is needed.

Determining the `run_exports` is still part of solving. The packages are still downloaded and extracted to get the run_exports (that's an improvement for later). The `PackageCache` is now shared which ensures that validation/download/extract is not duplicated (which was previously the case).

The gateway is also shared between all runs to reduce overhead. 

I also removed storing a number of CLI arguments in `Configuration`. These didnt make sense to me to store in the configuration and instead could better be passed as arguments to the functions that require them.

However, currently the download progress of packages is lost. I wanna fix this before we merge.


baszalmstra:
I also added fancy progress now:

image

baszalmstra:
This is now passing CI and ready for a review! @wolfv !

src/render/solver.rs
```diff
@@ -159,14 +183,16 @@ impl rattler_repodata_gateway::Reporter for GatewayReporter {
...
-        progress_bar.enable_steady_tick(Duration::from_millis(100));
+            .with_finish(indicatif::ProgressFinish::AndLeave)
+            .with_prefix("Downloading");
 
         // use the configured style
         if let Some(template) = &self.progress_template {
             progress_bar.set_style(template.clone());
         }
+
+        // progress_bar.enable_steady_tick(Duration::from_millis(100));
```
wolfv, line 194: keep or remove?

wolfv:
Functionality is great! I am happy to merge. The only thing I found when testing is that the indent doesn't seem to be _quite_ correct (screenshot attached).

Maybe we should start a new tracing span that says "installing environments"?

Screenshot 2024-09-06 at 09 18 49

wolfv:
Also the bar from download & extract doesn't "go away" (but that's also a super minor thing).

src/render/solver.rs
baszalmstra:
```suggestion
```

PR #1032, opened 2024-08-26
docs: add `go` tutorial

wolfv:


docs/tutorials/go.md
```diff
@@ -0,0 +1,72 @@
...
+  summary: Temporal CLI
+  description: |
+    Command-line interface for running Temporal Server and interacting with
+    Workflows, Activities, Namespaces, and other parts of Temporal.
+  license: MIT
+  license_file:
+    - LICENSE
+    # These license files are generated at build time in the `build.sh` script
+    # from all the dependencies of `temporal.io`.
+    - license-files/
```
Hofer-Julian, line 57: Could you please move the code blocks to concrete files so it is easier to test locally? Similar to how it is done [here](https://github.com/prefix-dev/pixi/pull/1879).

Mid-term it would be good to at run `rattler-build --render-only` on CI for a basic sanity check.

PR #1036, opened 2024-09-02
fix: CLI docs

Hofer-Julian:
The CLI docs were outdated. I've updated them by running `pixi run generate-cli-docs`.
No idea why so much changed here.


wolfv:
Probably because @ruben-arts fork of clap-markdown was commented out in the `Cargo.toml`. Unfortunately I don't remember what the fork did...

Hofer-Julian:
@wolfv I added the patch again

Issue #1028, opened 2024-08-20
NumPy variants should not influence `noarch` packages


wolfv:
I am not sure how to fix this, but @hmaarrfk noted an issue where `rattler-build` is building 3 variant packages for a `noarch` Python package.

The variant is different in the numpy pinning. 

As far as I am aware, we only remove the pin on Python for a noarch package. I am not sure what `connda-build` logic is here, exactly. Should all pins be removed for noarch packages?

For reference the PR: https://github.com/conda-forge/staged-recipes/pull/27281
And the logs: https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1003575&view=logs&jobId=6f142865-96c3-535c-b7ea-873d86b887bd&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823

hmaarrfk:
Thanks for writing this out. Here is an archive of the log since they tend to disappear on branches:
[18_log.txt](https://github.com/user-attachments/files/16681528/18_log.txt)

wolfv:
I believe that anything _zipped_ with Python (by using `zip_keys`) will also be ignored in the pinnings. 

I tried a bunch of combinations with conda-build and the following does give only one build:

```yaml
zip_keys:
  -
    - python
    - numpy

numpy:
  # part of a zip_keys: python, python_impl, numpy
  - 1.22
  - 1.22
  - 1.22
  - 1.23
  - 1.26
python:
  # part of a zip_keys: python, python_impl, numpy
  - 3.8.* *_cpython
  - 3.9.* *_cpython
  - 3.10.* *_cpython
  - 3.11.* *_cpython
  - 3.12.* *_cpython
python_impl:
  # part of a zip_keys: python, python_impl, numpy
  - cpython
  - cpython
  - cpython
  - cpython
  - cpython
```

I think we should follow that logic for now.

PR #1037, opened 2024-09-02
fix used vars with run dependencies

wolfv:
rattler-build was (wrongly) including run dependencies in the `used_vars`.w


wolfv:
fixes #1028

PR #1057, opened 2024-09-11
add in env vars from variant config

wolfv:
@h-vetinari this is the initial fix for the problem.

However, it will exhibit some problems:

A python variant would overwrite the PYTHON env var on Windows (because env vars are not case sensitive on Windows).

We should rethink how to do this. Maybe only inject them in addition to other env vars when they are not set?

Any ideas?

### DIFF
--- a/src/env_vars.rs
+++ b/src/env_vars.rs
@@ -329,5 +329,8 @@ pub fn vars(output: &Output, build_state: &str) -> HashMap<String, String> {
         timestamp_epoch_secs.to_string(),
     );
 
+    // Insert all variant variables into the environment
+    vars.extend(output.variant().clone().into_iter());
+
     vars
 }


h-vetinari:
> A python variant would overwrite the PYTHON env var on Windows (because env vars are not case sensitive on Windows).

Not sure I understand the problem. each output should be building _at most_ for one python version. How can we get this wrong? If a given output in a recipe wants to use `PYTHON` but doesn't declare a dependency on `python`, that's a bug in the recipe.

> We should rethink how to do this. Maybe only inject them in addition to other env vars when they are not set?

If the logic for that is not too cumbersome, that sounds like a legit approach as well (but again, I haven't yet understood what you consider to be the problem)

wolfv:
The python env var points to the path of the interpreter.

h-vetinari:
So? My point is: each output that has a legitimate use for the env var (and therefore declares a dependency on python) should get the right one when the config for that output is instantiated.

At least conceptually, there's only one python variant left at that point - though I cannot tell if this is what your current implementation does.

wolfv:
Yeah I am not debating that. The problem is that there is also a `${PYTHON}` env variable that would be overwritten by a `python` env variable on systems with case insensitive env vars.

h-vetinari:
> The problem is that there is also a `${PYTHON}` env variable that would be overwritten by a `python` env variable on systems with case insensitive env vars.

Ah, ok, sorry, I hadn't gotten that aspect so far... Yeah, I guess `python` at least needs to be special-cased for that reason. But probably better to do the thing which you described, i.e. don't update env vars which are already set.

wolfv:
I think this needs a slightly bigger refactor to match `conda-build`'s behavior better.

First of all, conda-build does indeed have a list of environment vars that are not pushed through from the variant config: https://github.com/wolfv/conda-build/blob/8e8ff69b739cc53af1d7fa512f79c7c2a04e6e5a/conda_build/environ.py#L49

However, conda-build also overrides some env vars from the variant config which also marks them as used. We should follow this logic in rattler-build.

wolfv:
This is going to be super-seded by #1088

PR #1062, opened 2024-09-20
fix: latest_tag get oldest tag

fecet:
The current implementation of `git.latest_tag` seems a bit off. Using `-v:refname` together with `last()` is actually selecting the earliest tag, not the latest one. 

One can check this by 

```yaml
context:
  git_repo_url: https://github.com/prefix-dev/rattler-build.git
  latest_tag: ${{ git.latest_tag( git_repo_url ) }}

package:
  name: "rattler"
  version: ${{ latest_tag }}

source:
  git: ${{ git_repo_url }}
  tag: ${{ latest_tag }}


```


wolfv:
great, thanks!

PR #1067, opened 2024-09-24
glob folders as well from simple name

### LINKED ISSUE #1055: ENH: also glob folders under `files:` without trailing backslash
Currently, rattler-build will -- in contrast to conda-build -- fail to pick up `$PREFIX/include/...` when doing
```yaml
build:
  files:
    - include
```
but _will_ pick it up if there's a trailing backslash, i.e. `- include/`. In https://github.com/prefix-dev/rattler-build/issues/1051 we discussed:
> @h-vetinari: Is it necessary to require the trailing `/` for folders...? 🤔

> @wolfv: Define necessary :D I think if you do `ls include` you won't see the contents of the folder. So I think in terms of "glob" it's more appropriate, and I also like things to be stricter. But we can totally debate this in another issue.

> @h-vetinari:
> > I think if you do `ls include` you won't see the contents of the folder.
> 
> I see no difference (on my CLI) between `ls folder` and `ls folder/` (both on linux & on win). 🤷
> 
> > I also like things to be stricter. But we can totally debate this in another issue.
> 
> I think we have a breakage-budget for the transition that we shouldn't overstretch. The more subtle behaviour changes accumulate, the harder it's going to be to pull off. [...] for folder-globs I don't see the benefit in terms of clarity gained compared to all the recipe's we'll break -- perhaps silently, because why duplicate in the tests what's already specified under `files:`, right? -- and then have to remember to look out for and fix.

IMO rattler-build should match conda-build here. It could _additionally_ warn on a missing backslash if it matches a folder, but given that even `ls` makes no distinction there, this added strictness is IMO a net negative.

wolfv:
This would fix #1055 but needs more work because it doesn't compile right now. Idea is to store the "source" glob alongside the evaluated glob for better serialization. Also later we need this sometimes to change how the glob behaves in the package content tests (where we amend it with e.g. `*.so`.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -67,12 +67,10 @@ jobs:
           - target: x86_64-pc-windows-msvc
             os: windows-2022
     env:
-      #
       # These are some environment variables that configure the build so that the binary size is reduced.
       # Inspiration was taken from this blog: https://arusahni.net/blog/2020/03/optimizing-rust-binary-size.html
       # They're only enable it on main and releases.
       #
-
       # Enable Link Time Optimization (LTO) for our release builds. This increases link time but drastically reduces
       # binary size.
       CARGO_PROFILE_RELEASE_LTO: ${{ (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/')) && 'true' || 'false' }}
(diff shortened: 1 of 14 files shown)


wolfv:
@minrk - I think you also had a test for this. Do you want to give this a whirl?

minrk:
I don't think I do? I had [this test](https://github.com/minrk/test-recipe/tree/main) for #1167, but not this one.

PR #1079, opened 2024-10-02
Include `git` output error context on git failures

andrewjcg:
Previously, we'd capture and just throw the error messages away. Instead, format them with the returned error.


wolfv (approved):

wolfv:
LGTM, will merge when tests are passing! Thank you!

PR #1081, opened 2024-10-02
detect virtual packages from environment

wolfv:
@hadim this is the fix for `CONDA_OVERRIDE_CUDA` to work 🤦 

To be honest though, I would like to remove this in favor of explicit `system-requirements` in the recipe at some point.

### DIFF
--- a/src/render/solver.rs
+++ b/src/render/solver.rs
@@ -71,7 +71,7 @@ pub async fn solve_environment(
     let virtual_packages = tool_configuration.fancy_log_handler.wrap_in_progress(
         "determining virtual packages",
         move || {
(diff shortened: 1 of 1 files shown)


hadim:
Ahah I just pushed the same fix at https://github.com/hadim/rattler-build/commit/68ee5b85ca1cdec8fad456a22a0bb6faac525ae1 :-)

Thanks for doing it!

hadim:
> To be honest though, I would like to remove this in favor of explicit `system-requirements` in the recipe at some point

Yes, it makes sense to me.

---

I tested it and it indeed works (using the Dockerfile example from https://github.com/conda-forge/xformers-feedstock/pull/33#issuecomment-2389282544

wolfv:
Awesome!

PR #1089, opened 2024-10-07
Avoid fetching all tags on git clone/fetch

andrewjcg:
By default, git fetch/clone will download all tags from the remote repo, so pass `--no-tags` to avoid some unnecessary downloading -- we'll still explicitly fetch any tag we need to checkout.

### DIFF
--- a/src/source/git_source.rs
+++ b/src/source/git_source.rs
@@ -46,6 +46,8 @@ pub fn fetch_repo(
             // This should be safe, as we do a `git checkout` below to refresh
             // the working copy.
             "--update-head-ok",
+            // Avoid overhead of fetching unused tags.
+            "--no-tags",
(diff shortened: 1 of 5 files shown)


wolfv:
Thanks! Looks like a few tests don't work with this change yet :/

andrewjcg:
Ah interesting, it passes locally for me.  I wonder if this is related to git version -- looking...

andrewjcg:
Ah nevermind, just wasn't using the correct command line to include this test :P

andrewjcg:
THe issue was that the test recipe was specifying a "tag" using `rev:` (fixed to use `tag:`).

wolfv:
Ouch! Thanks for fixing all this! :)

PR #1094, opened 2024-10-09
fix: correct `PackageContentsTest` glob prefix on windows

chawyehsu:


wolfv:
Thank you @chawyehsu!

PR #1119, opened 2024-10-14
feat: keep installed packages in environments

wolfv:


wolfv:
The way this is done right now is a bit ad-hoc.

For the host-env cleaning we are not making sure that any files were touched. This might be something to fix by e.g. looking at all the hashes in the host env and comparing with what's written in the `PrefixRecord`.

For the work dir, we just `rm -rf` the whole thing. That removes any `CMakeCache` file or other intermediary output from the previous run.

We could instead:

- render out the sources into a JSON and store that in the workdir, to check if there are any different patches and only then recreate the whole work dir from scratch?
- add a `--keep-work-dir` flag to not recreate it, and keep the current recreate-from-scratch logic? This might actually be useful if people do little tweaks to the contents directly.

PR #1122, opened 2024-10-16
WIP: new variant resolving / rendering 

wolfv:
This is an attempt to separate things better into "stage 1" and "stage 2" rendering.

### Stage 1:

Get all used variables from Jinja and conditional expressions and then render these out for all combinations that are found based on the variant config. This already creates a number of different combinations.

### Stage 2:

From the rendered recipes (no more Jinja, conditionals) we need to get all the `requirements` at build/link time and create additional variants.

### Stage 3:

Topologically sort and de-duplicate outputs (with multiple outputs, these renders might create duplicate outputs, for example if you have a C library and multiple variants for Python bindings).

### DIFF
--- a/examples/rich/recipe.yaml
+++ b/examples/rich/recipe.yaml
@@ -38,6 +38,11 @@ tests:
   - python:
       imports:
         - rich
+  - script:
+      - python -e "print(\"foo\")"
+    requirements:
+      run:
+        - python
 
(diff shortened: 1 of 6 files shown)


wolfv:
Note: missing in this PR are updated handling of `-` and `.` in used variables :)

PR #1132, opened 2024-10-22
add shebang to bash preamble

pbauwens:
Scripts with no shebang are blocked by certain endpoint security tools, failing the rattler build at hand. Adding a bash shebang to the bash preamble should fix this issue.

### DIFF
--- a/src/script.rs
+++ b/src/script.rs
@@ -28,8 +28,7 @@ use crate::{
     },
 };
 
-const BASH_PREAMBLE: &str = r#"
-#!/bin/bash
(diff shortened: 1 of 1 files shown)


src/script.rs
```diff
@@ -29,6 +29,7 @@ use crate::{
 };
 
 const BASH_PREAMBLE: &str = r#"
+#!/bin/bash
```
wolfv, line 32: looks good but I think we need to put it even one line higher (right after the #")

pbauwens: moved it up :)

wolfv (approved):

Issue #940, opened 2024-06-21
Failed to extract archive on windows


Krande:
When using pixi and rattler-build v0.18 `pixi run dca -c https://repo.prefix.dev/code-aster -c conda-forge` on [this ](https://github.com/Krande/condapackaging) repository, I get

```
 ╭─ Running build for recipe: code-aster-17.0.21-py_nompi_release_h_200
 │
 │ ╭─ Fetching source code
 │ │ Downloaded file from https://gitlab.com/krande/src/-/archive/17.0.21.3/src-17.0.21.3.tar.gz
 │ │ Validated SHA256 values of the downloaded file!
 │ │ Extracting tar file to cache: "C:\\Work\\code\\condapackaging\\output\\src_cache\\src-17_0_21_3_9da7ea83.tar.gz"
 │ │
 │ ╰─────────────────── (took 29 seconds)
 │
 ╰─────────────────── (took 29 seconds)
 × error Error building package: Failed to extract archive: failed to unpack `\\?\C:\Work\code\condapackaging\output\src_cache\src-17_0_21_3_9da7ea83\.tmpWso4Cf\src-17.0.21.3\waf`
Error:   × Failed to extract archive: failed to unpack `\\?\C:\Work\code\condapackaging\output\src_cache\src-17_0_21_3_9da7ea83\.tmpWso4Cf\src-17.0.21.3\waf`
```

Let me know if you need any additional information from me?

Best Regards
Kristoffer

wolfv:
Interesting - does the archive contain symlinks?

Krande:
yeah, I think there are a few symlink files at the root of the archive 

![image](https://github.com/prefix-dev/rattler-build/assets/9642232/24b1ae69-3c99-4a38-b505-728e90b3e254)

wolfv:
Hmm symlinks are tricky on Windows. You can enable `developer mode` to allow symlinks on Windows without elevated permissions. I don't know how `conda-build` (or other archive extraction programs) deal with this. 

Do you know if this used to work in conda-build?

wolfv:
https://learn.microsoft.com/en-us/windows/apps/get-started/enable-your-device-for-development

Krande:
These files arent used on windows. So they are only relevant for linux. So I guess they're treated as simple binary files?

wolfv:
Could you try to extract this archive with `tar` or `conda-build` on Windows? I am just wondering what they do...

[...]

PR #1138, opened 2024-10-26
fix: add tolerant tar extraction on windows to solve failing on symlinks

Krande:
Hey guys, 

this is supposed to close https://github.com/prefix-dev/rattler-build/issues/940. 

This is my first time writing rust beyond just "hello world", so any suggestions for improvements are most welcome!

### DIFF
--- a/src/source/extract.rs
+++ b/src/source/extract.rs
@@ -1,6 +1,6 @@
 //! Helpers to extract archives
 use std::{ffi::OsStr, io::BufRead, path::Path};
-
+use tar::EntryType;
 use crate::console_utils::LoggingOutputHandler;
 
 use fs_err as fs;
@@ -136,9 +136,29 @@ pub(crate) fn extract_tar(
     ));
 
     let tmp_extraction_dir = tempfile::Builder::new().tempdir_in(target_directory)?;
-    archive
-        .unpack(&tmp_extraction_dir)
-        .map_err(|e| SourceError::TarExtractionError(e.to_string()))?;
+    for entry in archive.entries()? {
+        match entry {
+            Ok(mut file) => {
+                let path = tmp_extraction_dir.path().join(file.path()?);
+
+                // Attempt to unpack symlinks, log errors if they fail on Windows
+                if file.header().entry_type() == EntryType::Symlink && cfg!(target_os = "windows") {
+                    if let Err(e) = file.unpack(&path) {
+                        println!("Warning: failed to extract symlink {:?} due to {:?}", path, e);
+                    }
+                    continue;
+                }
+
+                // Attempt to unpack other files individually, handle errors gracefully
+                if let Err(e) = file.unpack(&path) {
+                    println!("Warning: failed to unpack {:?} due to {:?}", path, e);
+                }
+            }
+            Err(e) => {
+                println!("Warning: failed to read an entry due to {:?}", e);
+            }
+        }
+    }
 
     move_extracted_dir(tmp_extraction_dir.path(), target_directory)?;
     progress_bar.finish_with_message("Extracted...");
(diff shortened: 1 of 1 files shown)


wolfv:
Thank you! On first glance, looks good!

Krande:
@wolfv i think I've fixed the linting issue. Let me know if there is anything else I should consider.

src/source/extract.rs
```diff
@@ -136,9 +136,38 @@ pub(crate) fn extract_tar(
...
+                // Ensure all intermediate directories exist before unpacking
+                if let Some(parent) = path.parent() {
+                    fs::create_dir_all(parent)?;
+                }
+
+                // Attempt to unpack symlinks and regular files, logging any issues
+                match file.header().entry_type() {
+                    EntryType::Symlink if cfg!(target_os = "windows") => {
+                        if let Err(e) = file.unpack(&path) {
+                            println!(
```
wolfv, line 153: We should always use `tracing::warn!` or `tracing::error!`.

src/source/extract.rs
```diff
@@ -136,9 +136,38 @@ pub(crate) fn extract_tar(
...
+                    }
+                    _ => {
+                        if let Err(e) = file.unpack(&path) {
+                            println!("Warning: failed to unpack {:?} due to {:?}", path, e);
+                        }
+                    }
+                }
+            }
+            Err(e) => {
+                println!("Warning: failed to read an entry due to {:?}", e);
```
wolfv, line 167: I think in this case, we should also return the error and fail the operation.

src/source/extract.rs
```diff
@@ -136,9 +136,38 @@ pub(crate) fn extract_tar(
...
+                        if let Err(e) = file.unpack(&path) {
+                            println!(
+                                "Warning: failed to extract symlink {:?} due to {:?}",
+                                path, e
+                            );
+                        }
+                    }
+                    _ => {
+                        if let Err(e) = file.unpack(&path) {
+                            println!("Warning: failed to unpack {:?} due to {:?}", path, e);
```
wolfv, line 161: Again we should return the error here :)

wolfv (changes requested):
Hi! Excellent PR and I agree about being more lenient about symlinks on Windows.

I think the error handling is a bit _too_ lenient now, though. There are definitely errors that are worth bailing out.

Krande:
If you have time to implement the changes, feel free to finish them:) if not, I'll likely start to implement them after lunch time tomorrow!

wolfv:
I looked a little bit deeper and I think unfortunately this doesn't cover a bunch of (other) edge cases that the original tar extraction code covers. That's really unfortunate. Ideally we could implement this kind of behavior / filtering in tar-rs directly.

Krande:
I see. I checked tar-rs and found this https://github.com/alexcrichton/tar-rs/issues/280 which I guess is related.

Krande:
Would it be possible to make this tolerant extraction feature optional by hiding it behind an opt-in flag until a better long-term solution can be found?
 
That way, it is at least possible for us forced to use windows in our day jobs to use rattler-build for projects containing symlinks :)

wolfv:
A "shitty" workaround is to keep the source as tarball. This is possible since the last release or two by adding a option:

```
file_name: source.tar.gz
```

Then you can add e.g. `7zip` to your build dependencies, and extract the source code during the build step. 

Is that an OKish workaround? Sorry for not coming up with something better. I would love an option in the tar crate. Maybe you could also work on a PR over there?

Krande:
> Is that an OKish workaround?

Yeah, anything that works is fine for now :)  That at least gives me a chance at using rattler-build for more packages :) 

> Maybe you could also work on a PR over there?

Sure, I can try. It would also help if you can shed some light on the types of edge cases a filtering behaviour in tar-rs should be able to cover? Referring to your comment

> I looked a little bit deeper and I think unfortunately this doesn't cover a bunch of (other) edge cases that the original tar extraction code covers.

wolfv:
@Krande I think if you add this behavior to the original tar, then we can keep using the existing `unextract` functionality that covers all the edge cases. But we can just gracefully warn instead of fail when we cannot create a symlink. 

You could add a SymlinkBehavior enum to tar-rs and have Fail, Warn or Copy options and then implement them accordingly when symlinking fails.

PR #1159, opened 2024-11-04
fix relinking issues on macOS

wolfv:


munechika-koyo:
This is related to the issue #1159

Issue #1168, opened 2024-11-07
Python tests should allow custom requirements


hadim:
I think the python tests should allow custom deps spec. For example, in [this feedstock]() I need to set the python version of a noarch package and it is not allowed:

```yaml 
tests:
  - python:
      imports:
        - pptx
      requirements:
        run:
          - python =${{ python_min }}
 ```

so I had to replace it with:

```yaml 
tests:
  - script:
      - python -c "import pptx"
    requirements:
      run:
        - python =${{ python_min }}
 ```

---

What do you think is the best way to address this?

1) allow `requirements` or `requirements.run`?
2) add `python_version` so only the python version can be customized?

Let me know, what you think is best here and I can submit a PR if you want.

hadim:
xref https://github.com/conda-forge/python-pptx-feedstock/pull/28

PR #1170, opened 2024-11-07
Introduce `python_version` in `tests[0].python`

hadim:
It's a proposal to fix https://github.com/prefix-dev/rattler-build/issues/1168

Please let me know if you would prefer another approach.


hadim:
Here is the kind of recipe that now works:

```yaml
package:
  name: test-with-custom-deps
  version: 0.0.1

build:
  number: 0
  noarch: python
  script: touch $PREFIX/test-with-custom-deps

requirements:
  run:
    - python

tests:
  - python:
      imports:
        - pandas
      pip_check: true
      requirements:
        run:
          - pandas
```

hadim:
Recipe format PR at https://github.com/prefix-dev/recipe-format/pull/43

hadim:
@wolfv what do you think about this approach?

wolfv:
Excellent PR, really awesome! 

Hmm I am a bit conflicted about this. The python check is pretty simplistic anyways and one can also relatively easy make better tests using the script tests. One can even write Python scripts directly in the scripts test, e.g.

```
- script:
    interpreter: python
    content:
      - import foo
      - import bar
      - print("hello world")
    requirements:
      run:
        - python
        - foo
```

Although it's clearly a bit more verbose.

Instead of the total flexibility of requirements we could also add a python version field to the python test as another idea:

```
python:
  imports:
     - foo
  python_version: 3.8  # or [3.8, 3.9, ...]
```

So in general I am a bit on the fence.

hadim:
> The python check is pretty simplistic anyways and one can also relatively easy make better tests using the script tests.

Yes, I understand your point here.

> Instead of the total flexibility of requirements we could also add a python version field to the python test as another idea

This is an alternative solution I also proposed at https://github.com/prefix-dev/rattler-build/issues/1168. 

I think that one is better since it's specific to python and, as you said, for a more complex scenarios, the user already has other options.

---

So would you be okay if I give it a try to `python_version: 3.8  # or [3.8, 3.9, ...]`?

hadim:
@wolfv I gave it a try to `python_version`.

No stress if you still don't want that feature. Just let me know.

docs/tutorials/javascript.md
```diff
@@ -45,4 +45,4 @@ requirements:
 tests:
   - script:
     - bibtex-tidy --version
-```
\ No newline at end of file
+```
```
hadim, line 48: Unrelated from `pixi run lint`

Issue #1175, opened 2024-11-09
Add new flags to facilitate automated forge


hadim:
I am trying to setup a private forge with a mix of inter-dependents noarch and arch-specific packages. The setup is similar to https://github.com/prefix-dev/ai-forge/.

<details>
<summary>The variants file looks like this.</summary>

```yaml
zip_keys:
  - - ${{ "cxx_compiler_version" if unix }}
    - ${{ "c_compiler_version" if unix }}
  - - python
    - numpy
    - python_impl

c_stdlib:
  - ${{ "sysroot" if linux }}
  - ${{ "macosx_deployment_target" if osx }}
  - ${{ "vs" if win }}

c_stdlib_version:
  - ${{ "2.17" if linux }}
  - ${{ "10.13" if osx and x86_64 }}
  - ${{ "11.0" if osx and arm64 }}

c_compiler:
  - ${{ "gcc" if linux }}
  - ${{ "clang" if osx }}
  - ${{ "vs2019" if win and x86_64 }}
  - ${{ "vs2022" if win and arm64 }}

c_compiler_version:
  - ${{ "13" if linux }}
  - ${{ "17" if osx }}

cxx_compiler:
  - ${{ "gxx" if linux }}
  - ${{ "clangxx" if osx }}
  - ${{ "vs2019" if win and x86_64 }}
  - ${{ "vs2022" if win and arm64 }}

cxx_compiler_version:
  - ${{ "13" if linux }}
  - ${{ "17" if osx }}

rust_compiler:
  - rust

rust_compiler_version:
  - 1.80.0

python:
  # part of a zip_keys: python, python_impl, numpy
  - 3.12.* *_cpython
python_impl:
  # part of a zip_keys: python, python_impl, numpy
  - cpython
python_min:
  # minimum supported python version per CFEP-25
  # bump to next minor version when we drop python versions
  - "3.12"

# we build for the oldest version possible of numpy for forward compatibility
# we roughly follow NEP29 in choosing the oldest version
numpy:
  # part of a zip_keys: python, python_impl, numpy
  - 1.26
```

</details>

---

Here are two flags I think could improve the single-repo based way of setting up a forge.

- `rattler-build build --no-test-if-emulate`:

Do not run tests if `target-platform != build-platform`. Sometimes a test-time dep is not available on `osx-arm64` while the package has been built on `osx-64`.

- `rattler-build build --noarch-platform=linux-64`

[...]

PR #1190, opened 2024-11-13
Add `--no-test-if-emulate` flag 

hadim:
See https://github.com/prefix-dev/rattler-build/issues/1175 for context.

It adds a new flag to `build` that allow skipping the tests if cross compiling: `--no-test-if-emulate`

I have a few questions:

- should I prefer "emulate" or "cross compiling" for the variables as well as the CLI doc?
- how should `noarch` should be handled here? or when should we consider the build is a cross compilation or not?

### DIFF
--- a/src/tool_configuration.rs
+++ b/src/tool_configuration.rs
@@ -45,6 +45,9 @@ pub struct Configuration {
     /// Whether to skip the test phase
     pub no_test: bool,
 
+    /// Whether to skip the test phase if cross-compiling
+    pub no_test_if_emulate: bool,
+
     /// Whether to use zstd
     pub use_zstd: bool,
 
@@ -112,6 +115,7 @@ pub struct ConfigurationBuilder {
     client: Option<ClientWithMiddleware>,
     no_clean: bool,
     no_test: bool,
+    no_test_if_emulate: bool,
     use_zstd: bool,
     use_bz2: bool,
     skip_existing: SkipExisting,
@@ -135,6 +139,7 @@ impl ConfigurationBuilder {
             client: None,
             no_clean: false,
             no_test: false,
+            no_test_if_emulate: false,
             use_zstd: true,
             use_bz2: false,
             skip_existing: SkipExisting::None,
@@ -216,6 +221,14 @@ impl ConfigurationBuilder {
         }
     }
 
+    /// Sets whether tests should be executed if cross-compiling.
+    pub fn with_testing_if_emulate(self, testing_enabled_if_emulate: bool) -> Self {
+        Self {
+            no_test_if_emulate: !testing_enabled_if_emulate,
+            ..self
+        }
+    }
+
     /// Whether downloading repodata as `.zst` files is enabled.
     pub fn with_zstd_repodata_enabled(self, zstd_repodata_enabled: bool) -> Self {
         Self {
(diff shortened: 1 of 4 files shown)


wolfv:
maybe `--only-test-native` or some variation like that? Or `--no-emulation` or `--no-test-emulation`?

Or we could also make this opt-in, by saying: `--has-emulation`. Or something like `--run-test`.

wolfv:
If we make it opt-in, we can print a big fat warning and then say: to run tests in non-native environments, use `--run-test` or something liek that.

hadim:
Do you prefer this to be opt-in? I originally wanted to keep the default behavior the same but up to you.

- if we go with opt-in then `--run-tests` or `--run-tests-non-native` makes sense to me.
- if we go opt-out then I like `--only-test-native`

I guess the question (to decide the best default) is whether "on average" running tests on non-native will work more often than it will fail. 

Let me know what you prefer and I will adapt the code.

baszalmstra:
I think opt-in makes sense here. Looking from the perspective of someone who just wants to invoke rattler-build. I would not expect a big fat error.

hadim:
So, would `--run-tests-non-native` work for you?

hadim:
Also should I add a test in `test_simple.py`?

wolfv:
We had a build tools meeting and I brought up this question :)

One suggestion was to copy what conda-forge does: https://conda-forge.org/docs/maintainer/conda_forge_yml/#test

```
--test=native

--test=native-and-emulated
```

Although we don't really know when there is an "emulator". We could however potentially infer that if there are any `host` dependencies configured in a test ...

We could also retire the current `--skip-test` flag and do 

```
--test=skip
# and maybe the inverse
--test=always
```

Anyways, just ideas.

hadim:
I like the idea of something more expressive than a boolean flag.

I also think it would be nice to propose something related to emulation/cross-compilation so we don't have to compute that logic "outside" of `rattler-build`. Even if it's opinionated, we can always finetune and adapt it moving forward.

So, considering all the comments above, we could do:

- `--test=skip `: replace `--skip-test`
- `--test=native`: run when `host_platform != build_platform` + always run `target_platform == 'noarch'`
- `--test=native-and-emulated`: always run tests

We can have `--test=native` as default, so it's opt-in.

WDYT?

wolfv:
Looks good to me! I still would print a warning if tests are not run when cross compiling though.

hadim:
LGTM

baszalmstra:
Nice!

docs/reference/cli.md
```diff
@@ -344,14 +350,6 @@ Rebuild a package from a package file instead of a recipe
 	The package file to rebuild
 
 
-- `--no-test`
```
wolfv, line 347: unfortunately removing this option will break conda-forge. We should keep it but deprecate it.

src/tool_configuration.rs
```diff
@@ -29,6 +29,20 @@ pub enum SkipExisting {
...
 
+/// Container for the CLI test strategy
+#[derive(Debug, Clone, Copy, ValueEnum, Default)]
+pub enum TestStrategy {
+    /// Skip the tests
+    Skip,
+    /// Run the tests only if the build platform is the same as the host platform.
+    /// Otherwise, skip the tests. If the target platform is noarch,
+    /// the tests are always executed.
+    #[default]
```
wolfv, line 40: I am still undecided wether this or the `native-and-emulated` should be the default ...

docs/reference/cli.md
hadim: ok sounds good.

src/tool_configuration.rs
hadim: No real opinion on this. Will put `native-and-emulated`.

wolfv:
Thanks!

PR #1192, opened 2024-11-13
Skip noarch build if `--noarch-platform` == `build_platform`

hadim:
See https://github.com/prefix-dev/rattler-build/issues/1175 for context.

When `--noarch-platform` is set, the noarch builds will be skipped if they are being built on a different platform than the platform set.

---

It's an opinionated flag, and the idea is to make it easy to build single repos rattler-based forge. By setting this flag to `linux-64` for example, all the noarch packages of the forge will be **only** built on the `linux-64` CI runs and not the others.

If that too specific for rattler-build, let me know and I will close.


wolfv:
I did a tiny bit of cleanup to remove the `unwrap` :) 

I am wondering if we shoudl call this `--noarch-build-platform=...`, and similar in the code. I think that would make it very clear? What do you think @hadim?

PR #1193, opened 2024-11-14
feat: use gitignore when copying the recipe files

wolfv:


tdejager:
Just tried it on this: https://github.com/prefix-dev/pixi-build-backends/pull/9

And the result are smaller binaries, in comparison the previous ones were like 300mb
image

tdejager:
Yeah actually, it does still seem to include the `.git` itself. The rest is okay. `.git` by itself is not included in the gitignore, is this something to exclude manually?

image

wolfv:
We can add `.hidden(true)` and then that would also be ignored.

tdejager:
Sounds good!

tdejager:
LGTM!:

image

PR #1195, opened 2024-11-15
feat: add pixi build deps

nichmor:
We need to lock on rattler from feat/pixi-build, otherwise we cannot compile


wolfv:
Is this still needed?

PR #1205, opened 2024-11-21
do not use gitignore when copying test files

### LINKED ISSUE #1180: Bug: Copying of test source files
When building a new conda package ([here](https://github.com/conda-forge/staged-recipes/pull/28099) for reference), rattler-build is not copying my binary h5-file from the source-code, which would be needed for a quick test. I'm currently not able to reproduce the error in a more minimal setting, as it does seem to only occur in that special setting. I'm currently downloading the file via `curl` to circumvent that behaviour.

The problem is: I'm copying the complete test directory of the source, and all files are copied with the h5-file in `microstructures` being the exception. The folder `test` referenced in the source files looks like this:
```
- test
    - input_files
        - file1.json
        - file2.json
    - microstructures
        - sphere32.h5
    - run_test.sh
    - CMakeLists.txt
```

I'm also surprised by the rather deeply nested placement of the source files, but maybe I'm doing sth wrong there...? I was not able to get that information from the documentation.

Any help would be greatly appreciated!

wolfv:
Should fix #1180 

I think it's fine to not use gitignore in these cases. We'll switch to [`wax`](https://docs.rs/wax/latest/wax/) for this in the future anyways.

### DIFF
--- a/src/package_test/serialize_test.rs
+++ b/src/package_test/serialize_test.rs
@@ -32,7 +32,7 @@ impl CommandsTest {
                 folder,
             )
             .with_globvec(globs)
-            .use_gitignore(true)
+            .use_gitignore(false)
             .run()?;
 
             test_files.extend(copy_dir.copied_paths().iter().cloned());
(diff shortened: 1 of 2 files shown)


wolfv:
I changed the `parents` option in gitignore now. Hopefully that helps! Would love it if you could test @rgommers / @minrk

Issue #1149, opened 2024-11-01
Bug: --no-test causing output cache repodata.json to not get updated


tl-hbk:
When building a multi-output recipe with the following command

```
rattler-build build -r <recipe path> --no-test --package-format tar-bz2 --output-dir <dist path>
```

The `repodata.json` file is missing the last package built.  In my recipe I have 4 packages being built and only 3 end up in the file despite all 4 packages showing up in the output directory

I'm using `rattler-build-0.28.2-h49672d7_0.conda` on windows

An example recipe and repodata.json

```
context:
  version: "1.0.0"

recipe:
  name: foo
  version: ${{ version }}

source:
    path: ..

outputs:
- package:
    name: foo-1

  build:
    noarch: generic
    script: echo 'foo' > bar.txt

  requirements:
    host:
      - python
      - pip
      - setuptools
    run:
      - python

- package:
    name: foo-2

  build:
    noarch: generic
    script: echo 'foo' > bar.txt

  requirements:
    host:
      - python
      - pip
      - setuptools
    run:
      - python

- package:
    name: foo-3

  build:
    noarch: generic
    script: echo 'foo' > bar.txt

  requirements:
    host:
      - python
      - pip
      - setuptools
    run:
      - python
```

```
{
  "info": {
    "subdir": "noarch"
  },
  "packages": {
    "foo-1-1.0.0-h4616a5c_0.tar.bz2": {
      "build": "h4616a5c_0",
      "build_number": 0,
      "depends": [
        "python"
      ],
      "md5": "86deddf3dcb00a53519cb7efd899b0f1",
      "name": "foo-1",
      "noarch": "generic",
      "sha256": "3271e95a04546d8b39e68d968fc149ce1b4b360cd2e43b0eb77a7a37a90306f0",
      "size": 4370,
      "subdir": "noarch",
      "timestamp": 1730486483166,
      "version": "1.0.0"
    },
    "foo-2-1.0.0-h4616a5c_0.tar.bz2": {
      "build": "h4616a5c_0",
      "build_number": 0,
      "depends": [
        "python"
      ],
      "md5": "800b2e754eac832c1dc2dc431711ad00",
      "name": "foo-2",
      "noarch": "generic",
      "sha256": "d7067a37b11aa9fc10d47a46f19b1be02d0840d52dad77549db38bacf34b182a",
      "size": 4352,
      "subdir": "noarch",
      "timestamp": 1730486483166,
      "version": "1.0.0"
    }
[...]

PR #1209, opened 2024-11-21
fix: reindex output cache after build (#1149)

tl-hbk:
Fix https://github.com/prefix-dev/rattler-build/issues/1149

The change in `build.rs` fixes the `--no-test` arg causing the repodata.json to not be regenerated after a package was built.


tl-hbk:
I removed the concurrency stuff from this PR as I'm having trouble with deadlocks.  Going to create a separate PR/issue for that

wolfv (approved):

wolfv:
Thanks a lot for the fix!

Issue #343, opened 2023-11-22
Allow to change channel priority


beenje:
It looks like channel priority is set to strict mode when testing with rattler-build.

This is an issue for us as we have 2 internal channels: the default one for stable packages and a "dev" one for testing.
When building a dev package we pass the "dev" channel and usually want the most recent version of some dependencies to be installed. That could come from the dev or stable channel. 

I haven't seen a way to configure the channel priority. It would be great to have a flag or another way to change that.

wolfv:
Hi @beenje indeed, we need to figure out how to do all of this. 
Another question is wether channel-specific matchspecs should work (e.g. `conda-forge::pytorch`).

beenje:
>  Another question is wether channel-specific matchspecs should work (e.g. `conda-forge::pytorch`).

Unfortunately, I don't think that would help with our workflow :-(

chebee7i:
Any update on this for channel-specific matchspecs?

wolfv:
No updates yet, sadly. 

I think our thoughts are to extend the `variant_config.yaml` file to include channels and solver mode. For example:


```yaml
options:
  channels: [...]
  mirrors: ... (or from global configuration)
  channel_priority: strict/lax
  target_platforms: ... ? should probably be a map? do we need this?

compilers:
  cxx: gcc 12.3
  c: co
  linux-64: 
     cxx: bla

variant_configuration:
  boost_cpp:
    - 1.4.5
    - 1.6.5
  bla: ...  
...

```

AaronOpfer:
This is a blocker for my work using `rattler-build` as we have repackaged several third-party packages for compatibility reasons and have inserted them into their own channel, which only works if the solver can pick from our channel or conda-forge without preference. If the compatibility channel is strictly preferred first, then builds on modern versions of Python break, and vice versa.

tl-hbk:
I had seen in the discord that I could work around this by using the channel pin in the `test.requirements.run` and `requirements.host` sections (channel::package_name) and that was working in most of my use-cases except the import tests which doesn't seem to have a way to specify requirements.  

I created a PR adding a `--channel-priority` arg to the cli https://github.com/prefix-dev/rattler-build/pull/1211

wolfv:
Thanks @tl-hbk !

happysalada:
Ive tested the new cli argument and couldnt get it to work, im curious if others had time to test it and made it work.
I tested it by specifying 3 channels as cli arguments.

[...]

PR #1211, opened 2024-11-22
feat: add `--channel-priority={strict|disabled}`

tl-hbk:
I needed a way to change the channel priority due to having a similar situation as the op of this issue https://github.com/prefix-dev/rattler-build/issues/343

I'm new to rust so there's probably a better way of adding it as a cli arg without the wrapper but I couldn't figure it out

### DIFF
--- a/src/opt.rs
+++ b/src/opt.rs
@@ -7,8 +7,8 @@ use clap_complete::{shells, Generator};
 use clap_complete_nushell::Nushell;
 use clap_verbosity_flag::{InfoLevel, Verbosity};
 use rattler_conda_types::{package::ArchiveType, Platform};
-use rattler_solve::ChannelPriority;
 use rattler_package_streaming::write::CompressionLevel;
+use rattler_solve::ChannelPriority;
 use serde_json::{json, Value};
 use url::Url;
 
@@ -200,7 +200,7 @@ pub struct CommonOpts {
 #[derive(Clone, PartialEq, Eq, Debug)]
 pub struct ChannelPriorityWrapper {
     /// The ChannelPriority value to be used when building the Configuration
-    pub value: ChannelPriority
+    pub value: ChannelPriority,
 }
 
 impl FromStr for ChannelPriorityWrapper {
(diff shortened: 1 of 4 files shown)


src/opt.rs
```diff
@@ -189,6 +190,29 @@ pub struct CommonOpts {
...
+}
+
+impl FromStr for ChannelPriorityWrapper {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "strict" => Ok(ChannelPriorityWrapper { value: ChannelPriority::Strict }),
+            "disabled" => Ok(ChannelPriorityWrapper { value: ChannelPriority::Disabled }),
+            _ => Err(format!("Unknown channel priority: {}", s))
```
wolfv, line 213: For a user it would be nice to list the available options for a perfectly helpful error message :)

wolfv:
Looks good to me! Left one small comment.

src/opt.rs
tl-hbk: I've updated the error message

wolfv (approved):

tl-hbk:
@wolfv it looks like some workflows need to be approved before they can run.  Could you approve those?

wolfv:
For some weird reason the button to run the tests does not show up. Can you push an empty commit?

tl-hbk:
> For some weird reason the button to run the tests does not show up. Can you push an empty commit?

done

wolfv:
OK, finally CI is running! :)

happysalada:
Hi was there a specific test with channel priority disabled ?
Im still getting "is excluded because due to strict channel priority not using this"

I am specifying several channels as arguments from the cli, could that be the problem ?

tl-hbk:
I didn't add a test to the repo but after the new version was released I tested it also using several channels as cli args and it worked fine for me

```
rattler-build build -r <recipe> -c <channel 1> -c <channel 2> --channel-priority disabled
```

PR #1214, opened 2024-11-24
fix: resolve file contents properly

### LINKED ISSUE #1212: Script files for testing should resolve at build-time
I think if you do

```yaml
script: foo.bat
```

in a test, the `foo.bat` will be missing at runtime. I think instead we should resolve the test files at test-rendering time and have the contents in the rendered test yaml.

wolfv:
We were not properly embedding the contents of script files in the tests. This fixes that.

Fixes #1212


wolfv:
fixes #1214

PR #1218, opened 2024-11-26
doc: fix experimental env

minrk:
'1' is not a valid value for boolean env, apparently:

```
> RATTLER_BUILD_EXPERIMENTAL=1 rattler-build build
error: invalid value '1' for '--experimental'
  [possible values: true, false]
```

(arguably, it would be more expected for the documented `EXPERIMENTAL=1` to actually work, since it's a far more common pattern than string true/false in env).


wolfv:
Thanks @minrk!

PR #1222, opened 2024-11-27
[doc] mention automatic file extension in output/build/script

minrk:
learned about this [here](https://github.com/conda-forge/suitesparse-feedstock/pull/98#discussion_r1860019072), and it's super useful for multi-platform multi-output recipes!


wolfv:
Thanks!

PR #1226, opened 2024-11-29
feat: cache sources too

wolfv:
This PR enables the caching of the source folder, too. 

This meant we had to change the behavior a bit:

- cache has to be restored to the _exact_ same prefix. That means every output will build in the same directory (should also fix the Windows bug!). For this, we currently use the name of the first output (we don't have the `recipe.name` available, but that could be fixed).
- ~~When the cache is involved, we don't fetch sources for the outputs. We should, however, change this to fetch the "diff" of the sources. Sources that were already fetched for the cache should not be re-fetched, but additional sources should be added on top (user has to make sure that there is no clobbering.~~ Cache has its own source section now.

These changes made the example at https://github.com/wolfv/rattler-build-cache-test work :)

- [x] fix finalized sources in output (not correct right now I think)

### DIFF
--- a/py-rattler-build/rattler_build/__init__.py
+++ b/py-rattler-build/rattler_build/__init__.py
@@ -1,5 +1,4 @@
-from .rattler_build import \
-    get_rattler_build_version_py as _get_rattler_build_version_py
+from .rattler_build import get_rattler_build_version_py as _get_rattler_build_version_py
 
 
 def rattler_build_version() -> str:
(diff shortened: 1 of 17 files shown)


wolfv:
OK, I had a sort-of rethink a little bit and now I think that the cache should have it's own `source` section so that things are neatly separated.

wolfv:
Added a PR for the schema: https://github.com/prefix-dev/recipe-format/pull/45

Issue #790, opened 2024-04-11
Add `perl` test section?


wolfv:
Apparently conda-build has some magic that makes the `test.import` section also work for `perl`.

For example here: 

```
{% set name = "perl-data-dumper" %}
{% set version = "2.183" %}
{% set sha256 = "e42736890b7dae1b37818d9c5efa1f1fdc52dec04f446a33a4819bf1d4ab5ad3" %}

package:
  name: {{ name }}
  version: {{ version }}

source:
  url: https://cpan.metacpan.org/authors/id/N/NW/NWCLARK/Data-Dumper-{{ version }}.tar.gz
  sha256: {{ sha256 }}

build:
  number: 0
  # nothing provides perl >=5.32.1,<6.0a0 *_perl5 needed by perl-extutils-makemaker-7.36-pl5321hd8ed1ab_0
  skip: true  # [win]
  script:
    - perl Makefile.PL INSTALLDIRS=vendor NO_PERLLOCAL=1 NO_PACKLIST=1
    - make
    - make test
    - make install VERBINST=1

requirements:
  build:
    - {{ compiler('c') }}        # [not win]
    - {{ compiler('m2w64_c') }}  # [win]
    - make
  host:
    - perl
    - perl-extutils-makemaker
  run:
    - perl

test:
  imports:
    - Data::Dumper

about:
  home: https://metacpan.org/pod/Data::Dumper
  license: GPL-1.0-or-later OR Artistic-1.0-Perl
  license_family: OTHER
  summary: 'seeds germane, yet not germinated'
  license_file:
    - {{ environ["PREFIX"] }}/man/man1/perlartistic.1
    - {{ environ["PREFIX"] }}/man/man1/perlgpl.1
```

danielnachun:
That magic lives here: https://github.com/conda/conda-build/blob/3cf75b660e73babd4772d2cf65ac0026406ca558/conda_build/create_test.py, and it actually supports R and Lua as well.

I've been mass migrating all my personal recipes and feedstocks I manage in conda-forge over to `rattler-build` (it's worth it when it's 60x faster for some builds) and the lack of this feature is a huge blocker for migrating Perl recipes.

A couple of thoughts on this:
- This should support Perl, R, Lua, Ruby, and should probably also work for NodeJS (though shipping libraries for that ecosystem through Conda is not as useful).
[...]

PR #1229, opened 2024-11-30
feat: refactor script and add `perl` as an interpreter

wolfv:
This adds `perl` as an interpreter, and I verified that the following "works":

```yaml
package:
  name: test-perl-exec
  version: "0.1.0"

build:
  script:
    interpreter: perl
    content: |
      use strict;
      use warnings;
      use Config;

      print "Hello World!\n";
      print "Perl prefix: $Config{prefix}\n";

requirements:
  build:
    - perl
```

We can add support for `R`, `nodejs`, and a number of other languages in the same way.

cc @danielnachun

### DIFF
--- a/src/render/solver.rs
+++ b/src/render/solver.rs
@@ -35,8 +35,7 @@ fn print_as_table(packages: &[RepoDataRecord]) {
             package
                 .channel
                 .as_ref()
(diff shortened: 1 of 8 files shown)


wolfv:
This should also help with writing tests for perl (https://github.com/prefix-dev/rattler-build/issues/790)

PR #1233, opened 2024-12-04
feat: expose python as struct

nichmor:
## Overview
We need it from pixi-build-backend side to setup entry points

### DIFF
--- a/src/render/solver.rs
+++ b/src/render/solver.rs
@@ -35,8 +35,7 @@ fn print_as_table(packages: &[RepoDataRecord]) {
             package
                 .channel
                 .as_ref()
-                .map(|s| s.rsplit('/').find(|s| !s.is_empty()))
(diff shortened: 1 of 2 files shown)


src/render/solver.rs
```diff
@@ -35,8 +35,7 @@ fn print_as_table(packages: &[RepoDataRecord]) {
             package
                 .channel
                 .as_ref()
-                .map(|s| s.rsplit('/').find(|s| !s.is_empty()))
-                .flatten()
+                .and_then(|s| s.rsplit('/').find(|s| !s.is_empty()))
```
nichmor, line 38: clippy complained about it

wolfv: yeah i made the same change a few days ago but maybe it's not merged yet.

wolfv (approved):

PR #1234, opened 2024-12-04
docs: add some perl docs

wolfv:
Pinging @danielnachun for review

### DIFF
--- a/docs/tutorials/perl.md
+++ b/docs/tutorials/perl.md
@@ -1,0 +1,93 @@
+# Packaging a Perl (CPAM) package
+
+Packaging a Perl package is similar to packaging a Python package!
+
+## Building a Perl Package
+
+### A perl `noarch: generic` package
+
+The following recipe is for the Perl package `Call::Context`. We use `perl` in the `host` requirements, and install the package using `make`.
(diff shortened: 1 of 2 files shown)


danielnachun:
This reference has been helpful for me in understanding the usage of `INSTALLDIRS=vendor` vs `INSTALLDIRS=site`: https://stackoverflow.com/a/54470382.   My view is that is as a package manager, `rattler-build` should use `INSTALLDIRS=vendor` as `INSTALLDIRS=site` seems more intended towards CPAN installations.

As for the effect on `noarch`, the choice of that directory does not matter.  The structure inside of `${PREFIX}/lib/perl5` looks like this:
```
5.32
   core_perl
   site_perl
   vendor_perl
core_perl
site_perl
vendor_perl
```
`noarch: generic` packages with no native code are installed in the top-level `vendor_perl` folder, while package with native code are installed in the `vendor_perl` subdirectory for whatever version of Perl was used to build that package (in this case `5.32`).  This is because like with Python, the C ABI does not guarantee compatibility between major versions, and attempting to a load package with native code on a version of Perl other than the one used to compile will fail with an error that there is a mismatch in the Perl version.

Hopefully this explanation helps to clear up some of the details on Perl packages.  I'm very much looking forward to doing the same for other languages soon!

wolfv:
OK, sounds good! The noarch example was reading `site` before and got me confused :)

PR #1247, opened 2024-12-09
feat: make undefined more strict

wolfv:
This is the start of making parsing undefined Jinja variables more strict. 

I think we should insert some pseudo values for `hash` and `build_number` as those should be the only two values that we allow as undefined.


wolfv:
superseded by #1378

PR #1271, opened 2024-12-17
Strict parsing

wolfv:
We currently parse `python 3` as `python ==3` which is a deviation from conda-build (which has some complicated logic to add stars sometimes).

We propose to simplify the logic and ask the user to provide operators, suhc as `>=3`, `3.*`, `==3` etc.

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -3,7 +3,7 @@
 use std::{path::PathBuf, vec};
 
 use miette::{Context, IntoDiagnostic};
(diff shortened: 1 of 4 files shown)


baszalmstra:
Can you also make a Pr in rattler? Because now the behavior in rattler-build deviates from rattler itself.

wolfv:
Already done!

baszalmstra:
Jep just noticed https://github.com/conda/rattler/pull/989 Thanks!

PR #1279, opened 2024-12-18
fix: Fix "the env Jinja functions" example not to ref PKG_HASH

mgorny:
Fix the example in "the env Jinja functions" part of recipe documentation to use `${{ hash }}` rather than `${{ PKG_HASH }}`, as the latter does not work.


mgorny:
Thanks!

wolfv:
Thank you!

PR #1299, opened 2025-01-03
feat: Add S3 upload

delsner:
# Motivation

`rattler-build` part of https://github.com/conda/rattler/issues/960

# Changes

* Implement S3 upload in `rattler-build`
* Add python e2e test for upload to R2 S3 bucket


test/end-to-end/test_simple.py
```diff
@@ -208,6 +210,106 @@ def test_anaconda_upload(
     assert requests.get(URL).status_code == 200
 
 
+@pytest.fixture()
+def minio_server(tmp_path: Path):
+    tmp_dir = tmp_path / "minio"
+    # Start minio server
+    proc = subprocess.Popen(
+        [
+            f"minio{'.bat' if os.name == 'nt' else ''}",
```
delsner, line 219:
```suggestion
            "minio",
```

test/end-to-end/test_simple.py
```diff
@@ -208,6 +210,106 @@ def test_anaconda_upload(
...
+            str(tmp_dir),
+        ],
+    )
+    # Create a bucket (with retries, in case we have to wait until minio is up and running)
+    retries = 0
+    max_retries = 3
+    while retries < max_retries:
+        try:
+            subprocess.run(
+                [f"mc{'.bat' if os.name == 'nt' else ''}", "mb", "local/s3-forge"],
```
delsner, line 232:
```suggestion
                ["mc", "mb", "local/s3-forge"],
```

test/end-to-end/test_simple.py
```diff
@@ -208,6 +210,106 @@ def test_anaconda_upload(
...
+        "--force-path-style",
+        "true",
+        str(get_package(tmp_path, "globtest")),
+    ]
+    rattler_build(*cmd)
+
+    # Make sure the package was uploaded
+    assert b"globtest" in subprocess.check_output(
+        [
+            f"mc{'.bat' if os.name == 'nt' else ''}",
```
delsner, line 297:
```suggestion
            "mc",
```

wolfv:
Sorry, we merged some rather large refactor in #1371 and now this one has a few conflicts. 

I'll try to get Bas to start merging S3 support in rattler tomorrow and then we can hopefully get the rest in swiftly as well!

src/opt.rs
pavelzw, line 1: could you also add something here: https://github.com/prefix-dev/rattler-build/blob/main/docs/authentication_and_upload.md ?

src/upload/mod.rs
```diff
@@ -263,6 +265,75 @@ pub async fn upload_package_to_anaconda(
...
+    force_path_style: Option<bool>,
+    package_files: &Vec<PathBuf>,
+) -> miette::Result<()> {
+    // If either endpoint_url or region is set, ensure that both are set
+    if endpoint_url.is_some() != region.is_some() {
+        return Err(miette::miette!(
+            "Both endpoint_url and region must be specified together"
+        ));
+    }
+    // TODO: Add CLI args for SECRET_ACCESS_KEY, ACCESS_KEY_ID, and SESSION_TOKEN
```
delsner, line 283: @pavelzw do you think it's necessary to implement these arguments? To me it feels unnecessary to introduce a new process how to pass authentication information.

src/variant_config.rs
```diff
@@ -480,7 +480,7 @@ impl VariantConfig {
                     } else {
                         already_used_vars
                             .iter()
-                            .all(|(key, value)| combination.get(key).map_or(false, |v| v == value))
+                            .all(|(key, value)| combination.get(key) == Some(value))
```
delsner, line 483: Fixes clippy error

src/opt.rs
delsner: done

src/upload/mod.rs
pavelzw: given we already support non-persisted authentication using `AWS_SECRET_ACCESS_KEY` and `AWS_ACCESS_KEY_ID`, it's not strictly needed tbf

but it's maybe a bit awkward that this is the only upload subcommand that doesn't support credentials via CLI flag

src/upload/mod.rs
```diff
@@ -263,6 +265,75 @@ pub async fn upload_package_to_anaconda(
...
+    let bucket = channel
+        .host_str()
+        .ok_or_else(|| miette::miette!("Failed to get host from channel URL"))?;
+
+    let config_map = HashMap::from([(bucket.to_string(), s3_config.clone())]);
+    let s3 = S3::new(config_map, storage.clone());
+    let s3_client = s3
+        .create_s3_client(channel.clone())
+        .await
+        .map_err(|e| miette::miette!(e.to_string()))?;
```
pavelzw, line 306: aren't we usually using `.into_diagnostic` for these things?

src/upload/mod.rs
```diff
@@ -263,6 +265,75 @@ pub async fn upload_package_to_anaconda(
...
 
+/// Uploads a package to a channel in an S3 bucket.
+pub async fn upload_package_to_s3(
+    storage: &AuthenticationStorage,
+    channel: Url,
+    endpoint_url: Option<Url>,
+    region: Option<String>,
+    force_path_style: Option<bool>,
+    package_files: &Vec<PathBuf>,
+) -> miette::Result<()> {
```
pavelzw, line 276: we had some internal discussions whether we should reindex directly after uploading or leave that to `rattler-index s3`. for a proper package server i think having a cron job that runs `rattler-index s3` is still necessary to avoid potential race conditions when uploading two files at once.
i don't have any strong opinion on whether we should include it here.

pavelzw (approved):
thanks! lgtm

src/upload/mod.rs
```diff
@@ -263,6 +265,75 @@ pub async fn upload_package_to_anaconda(
...
+        s3_client
+            .put_object()
+            .key(key.clone())
+            .if_none_match("*") // Do not allow overwriting existing files
+            .body(body)
+            .bucket(bucket)
+            .send()
+            .await
+            .into_diagnostic()?;
+        tracing::info!("Uploaded package to s3://{bucket}/{key}");
```
pavelzw, line 331: depending on what we decide with https://github.com/prefix-dev/rattler-build/pull/1299#discussion_r1957168534, we should `info` that you should run `rattler-index s3 ...` afterwards IMO

src/upload/mod.rs
delsner: If we want to include it here, it's still beyond the scope of this PR as we don't have `rattler-index s3` yet. I'd go ahead and implement `rattler-index s3` next and then we can discuss if it makes sense to run it here or not.

src/upload/mod.rs
delsner: We're returning an `anyhow::Error` from `create_s3_client`, which I believe does not implement [`IntoDiagnostic`](https://docs.rs/miette/latest/miette/trait.IntoDiagnostic.html) (it's not a `std::error::Error`?). Therefore, we have to map the error here.

src/upload/mod.rs
delsner: We don't have `rattler-index s3` yet, so do you think we should wait to merge this PR?

pavelzw: No I think these are two parallel discussions.
For me it would be fine to first merge this and then afterwards adjust the uploading here if we decide we want to reindex after uploading.

src/upload/mod.rs
pavelzw: No I think these are two parallel discussions.
For me it would be fine to first merge this and then afterwards adjust the uploading here if we decide we want to reindex after uploading.

wolfv:
Looks good to me. Shoudl we remove the TODO comment annd merge?

delsner:
Let's go! :)

wolfv:
Now we also need OCI upload! :)

PR #1314, opened 2025-01-09
docs: add sandbox docs

wolfv:


docs/sandbox.md
```diff
@@ -0,0 +1,59 @@
...
+!!! warning
+    The sandbox feature is experimental and may not work as expected. The options might
+    change in future releases.
+
+Since the 0.34.0 release, `rattler-build` has a new experimental feature called
+`sandbox`. With the sandbox feature enabled (via `--sandbox`), the build process
+has much more restricted access to system resources on macOS and Linux. 
+As the sandbox feature is experimental it is disabled by default.
+
+In particular, with the default configuration, the build process can read the entire filesystem,
```
Hofer-Julian, line 12: What was the reason again, to let it read the entire filesystem by default?

docs/sandbox.md
```diff
@@ -0,0 +1,59 @@
...
+
+On Linux, by default, the sandbox configuration is as follows:
+
+- Read access to the entire filesystem (`/`)
+- Read and execute access to `/bin`, `/usr/bin`, `/lib`, `/usr/lib`, `/lib64`, `/usr/lib64`
+- Write access to the build directories and `/tmp`, `/var/tmp`, and `$TMPDIR` (if defined)
+
+### Windows
+
+Sandboxing the build process is not yet supported on Windows, and thus is entirely ignored.
```
Hofer-Julian, line 59:
```suggestion
Sandboxing the build process is not yet supported on Windows, and thus all passed sandbox flags are entirely ignored.
```

docs/sandbox.md
wolfv: Mostly laziness to get something off the ground. It would require some more work to figure out what needs to be avaiable in general.

docs/sandbox.md
```diff
@@ -0,0 +1,59 @@
...
+    The sandbox feature is experimental and may not work as expected. The options might
+    change in future releases.
+
+Since the 0.34.0 release, `rattler-build` has a new experimental feature called
+`sandbox`. With the sandbox feature enabled (via `--sandbox`), the build process
+has much more restricted access to system resources on macOS and Linux. 
+As the sandbox feature is experimental it is disabled by default.
+
+In particular, with the default configuration, the build process can read the entire filesystem, 
+but it cannot write outside of the build directories. The build process also cannot access the network.
```
Hofer-Julian, line 13:
```suggestion
but it cannot write outside of the build directories. The build process also cannot access the network.
In the future, we plan to enable the sandbox per default and restrict it further. 
```

Hofer-Julian (approved):

PR #1322, opened 2025-01-09
Use latest rust

wolfv:
fix a small compilation issue


pixi.toml
```diff
@@ -25,7 +25,7 @@ deploy-dev = "mike deploy --push dev devel"
 
 [dependencies]
 openssl = "3.*"
-rust = "~=1.81.0"
+rust = ">=1.83.0,<1.84"
```
h-vetinari, line 28: That doesn't seem to match the rest of this PR, which is on 1.84 already (but I guess 1.84 isn't packaged yet...)

wolfv: made them all 1.83

PR #1324, opened 2025-01-09
template url bla

wolfv:


baszalmstra:
It disnt work with the generic?

wolfv:
It could have worked but I would have to wire the generic up everywhere so I thought I can just make a custom serialization function for the pieces we need and be done with it.

PR #1332, opened 2025-01-14
fix: allow schema version toplevel

wolfv:
fixes https://github.com/conda-forge/conda-smithy/pull/2207#event-15831388488


lucascolley:
thanks for the quick fix!

PR #1352, opened 2025-01-18
refactor: make source generic

baszalmstra:
This refactors parsing code slightly to accept a generic `SourceCode`. This allows for the propagation of source code information from the input to errors and diagnostics. Miette provides source code information in the errors. However, the input for any parsing function is always `&str` which doesn't provide any information about the input file. This refactor changes that type to `SourceCode`.

This is especially useful in pixi-build where we can use that information to provide more contextual errors. It should not change anything for the rattler-build CLI besides that the error will indicate the filename (`recipe.yaml`).


wolfv:
Awesome, thanks 👍 🚀

Issue #1335, opened 2025-01-16
Jinja variables are being removed in `build.script.content`


hadim:
This recipe works:

```yaml
package:
  name: hellopackage
  version: 1.0.0

build:
  number: 0
  noarch: python
  script: ${{ PYTHON }} --help

requirements:
  host:
    - python
```

While that one fails:

```yaml
package:
  name: hellopackage
  version: 1.0.0

build:
  number: 0
  noarch: python
  script:
    env:
      MY_VARIABLE: "Hello, World!"
    content: ${{ PYTHON }} --help

requirements:
  host:
    - python
```

with this error:

```bash
$ rattler-build build -r recipe.yaml

 ╭─ Finding outputs from recipe
 │ Found 1 variants
 │ Build variant: hellopackage-2.15.1-pyh4616a5c_0
 │ 
 │ ╭─────────────────┬─────────╮
 │ │ Variant         ┆ Version │
 │ ╞═════════════════╪═════════╡
 │ │ target_platform ┆ noarch  │
 │ ╰─────────────────┴─────────╯
 │
 ╰─────────────────── (took 0 seconds)

 ╭─ Running build for recipe: hellopackage-2.15.1-pyh4616a5c_0
 │
 │ ╭─ Fetching source code
 │ │ No sources to fetch
 │ │
 │ ╰─────────────────── (took 0 seconds)
 │
 │ ╭─ Resolving environments
 │ │ 
 │ │ Resolving host environment:
 │ │   Platform: osx-arm64 [__unix=0=0, __osx=15.2=0, __archspec=1=m1]
 │ │   Channels: 
 │ │    - file:///Users/hadim/Code/conda-forge/bug-python-rattler-build/output/
 │ │    - conda-forge
 │ │   Specs:
 │ │    - python
 │ │ 
 │ │ ╭─────────────────┬────────────┬────────────────────┬─────────────┬────────────╮
 │ │ │ Package         ┆ Version    ┆ Build              ┆ Channel     ┆ Size       │
 │ │ ╞═════════════════╪════════════╪════════════════════╪═════════════╪════════════╡
 │ │ │ bzip2           ┆ 1.0.8      ┆ h99b78c6_7         ┆ conda-forge ┆ 120.03 KiB │
 │ │ │ ca-certificates ┆ 2024.12.14 ┆ hf0a4a13_0         ┆ conda-forge ┆ 153.41 KiB │
[...]

PR #1353, opened 2025-01-19
Fix removed jinja variable in `build.script.content`

hadim:
An attempt to fix https://github.com/prefix-dev/rattler-build/issues/1335

See below:

```yaml
# Working with this PR ✅
build:
  script:
    content:
      - ${{ PYTHON }} --version

# Working with this PR ✅
build:
  script:
    content: ${{ PYTHON }} --version

# Working before this PR ✅
build:
  script:
    - ${{ PYTHON }} --version

# Working before this PR ✅
build:
  script: ${{ PYTHON }} --version
```

This issue being that the jinja variables are removed from the string. So `${{ PYTHON }} --version` becomes ` --version` during the build.

---

Should I add test(s) for this? If yes then where?

### DIFF
--- a/src/recipe/parser/script.rs
+++ b/src/recipe/parser/script.rs
@@ -298,6 +298,23 @@ impl TryConvertNode<Script> for RenderedMappingNode {
             (Some(file), None) => file.try_convert("file").map(ScriptContent::Path)?,
             (None, Some(content)) => match content {
                 RenderedNode::Scalar(node) => ScriptContent::Command(node.source().to_owned()),
+                RenderedNode::Sequence(seq) => {
+                    let commands: Result<Vec<String>, _> = seq
+                        .iter()
+                        .map(|node| {
+                            if let RenderedNode::Scalar(scalar) = node {
+                                Ok(scalar.source().to_owned())
+                            } else {
+                                Err(vec![_partialerror!(
+                                    *node.span(),
+                                    ErrorKind::ExpectedScalar,
+                                    label = "expected a scalar value in sequence"
+                                )])
+                            }
+                        })
+                        .collect();
+                    ScriptContent::Commands(commands?)
+                }
                 node => node.try_convert("content").map(ScriptContent::Commands)?,
             },
             (None, None) => ScriptContent::Default,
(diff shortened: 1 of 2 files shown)


src/recipe/parser/script.rs
```diff
@@ -297,7 +297,7 @@ impl TryConvertNode<Script> for RenderedMappingNode {
             }
             (Some(file), None) => file.try_convert("file").map(ScriptContent::Path)?,
             (None, Some(content)) => match content {
-                RenderedNode::Scalar(node) => ScriptContent::Command(node.as_str().to_owned()),
+                RenderedNode::Scalar(node) => ScriptContent::Command(node.source().to_owned()),
                 node => node.try_convert("content").map(ScriptContent::Commands)?,
```
hadim, line 301: That one is the culprit for the remaining failing case (when `build.script.content`is a list) but I haven't found a way to fix it yet.

hadim:
Maybe for the record, I found this PR that introduced support for jinja variables in `build.script`: https://github.com/prefix-dev/rattler-build/pull/894

src/recipe/parser/script.rs
hadim: I found a way

hadim:
I was trying to add a few tests for this but when I remove the fix of this PR, the test still passes, so I am confused. It seems like parsing is done correctly in that simple scenario but fails when using `rattler-build`.

```rust
#[cfg(test)]
mod tests {
    use super::*;
    use serde_yaml;

    #[test]
    fn test_script_content_sequence() {
        let script = Script {
            content: ScriptContent::Commands(vec!["${{ PYTHON }} --version".to_string()]),
            interpreter: Some("bash".to_string()),
            ..Default::default()
        };

        let yaml = serde_yaml::to_string(&script).unwrap();

        let expected_yaml = r#"interpreter: bash
content: ${{ PYTHON }} --version
"#;

        assert_eq!(yaml, expected_yaml);

        let parsed_script: Script = serde_yaml::from_str(&expected_yaml).unwrap();
        assert_eq!(parsed_script, script);
    }
}
```

hadim:
I finally added more "high-level" tests in the rust-tests crate.

PR #1368, opened 2025-01-21
ci: add tbump in order to automate version bumping

Hofer-Julian:
Since py-rattler-build tests require versions to be synchronized, releasing became a bit annoying.

Adding this bumping task should make it nicer again.

### DIFF
--- a/tbump.toml
+++ b/tbump.toml
@@ -1,0 +1,43 @@
+github_url = "https://github.com/prefix-dev/rattler-build"
+
+[version]
+current = "0.35.6"
+
+# Example of a semver regexp.
+# Make sure this matches current_version before
+# using tbump
+regex = '''
+  (?P<major>\d+)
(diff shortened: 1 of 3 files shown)


tbump.toml
```diff
@@ -0,0 +1,43 @@
...
+  \.
+  (?P<patch>\d+)
+  (rc
+    (?P<candidate>\d+)
+  )?
+  '''
+
+[git]
+# The current version will get updated when tbump is run
+message_template = "Bump version: 0.35.6 → {new_version}"
```
wolfv, line 22: Should this be `{version} -> {new...}`?

wolfv:
Nice, this is very useful!

tbump.toml
Hofer-Julian: It can indeed be simplified

Issue #1363, opened 2025-01-20
Small PyPI generation improvements


wolfv:
I found two small issues when generating the recipe for `zarr`:

- Upstream uses `MIT License` which is not valid SPDX. We should maybe make a HashMap or other lookup to associate the right SPDX license if possible.
- Upstream uses a dependency with an extra for `numcodecs` which gets wrongly translated to `numcodecs [crc32c]>=0.14`

Also, the description is printed as a really long single line string but IDK if we can change this with serde_yaml.

wolfv:
When generating a recipe for `quantity-array`, the URL / filename should be normalized to `quantity_array`. Also, any non-alphanumeric character sequences should be reduced to `-` before normalizing I believe. 

https://packaging.python.org/en/latest/specifications/binary-distribution-format/#escaping-and-unicode

https://packaging.python.org/en/latest/specifications/name-normalization/#name-normalization

```
        # Additionally, non-alphanumeric values in the distribution are
        # normalized to underscores (_), meaning hyphens can never occur
        # before `-{version}`.
```

lucascolley:
gh-1373 addressed the url formatting somewhat, but it seems like the edge cases mentioned in https://github.com/prefix-dev/rattler-build/issues/1363#issuecomment-2607786942 require a bit more work

wolfv:
It might be nice to figure out if we can / should add `build_requirements` to the host dependencies. 

E.g. I tried to generate a recipe for `truststore` and it was missing `flit-core` as a host dep.

It will probably be tricky to figure out the right compilers needed.

PR #1373, opened 2025-01-22
fix: pypi url formatting

lucascolley:


src/recipe_generator/pypi.rs
```diff
@@ -232,10 +232,10 @@ pub async fn generate_pypi_recipe(opts: &PyPIOpts) -> miette::Result<()> {
     // replace URL with the shorter version that does not contain the hash
     let release_url = if release.url.starts_with("https://files.pythonhosted.org/") {
         let simple_url = format!(
-            "https://pypi.io/packages/source/{}/{}/{}-{}.tar.gz",
+            "https://pypi.org/packages/source/{}/{}/{}-{}.tar.gz",
             &info.name.to_lowercase()[..1],
             info.name.to_lowercase(),
-            info.name.to_lowercase(),
+            info.name.to_lowercase().replace("-", "_"),
```
wolfv, line 238: I think this is great! Although the actual rules are a bit more involved: https://github.com/prefix-dev/rattler-build/issues/1363#issuecomment-2607786942

The way I understand it:

- any sequence of `[-.` is going to be replaced by a _single_ `_`.

PR #1378, opened 2025-01-25
feat: improved Jinja with stricter undefined variable handling, integer and boolean types

wolfv:
This changes our code a bit:

- variables in the context and in variant configuration files can also be integers or strings. Quoted strings are correctly kept as _strings_ (even if their contents are e.g. integerish).
- I am working towards enabling the strict undefined behavior, but it comes with a caveat: `${{ "foo" if unix }}` stops working, because the non-existent `else` branch evaluates to Undefined which then throws an error on print. I am trying to come up with a workaround (worst case, editing these conditions to add an else branch manually).


wolfv:
Additional work items:

- add a TrackingContext struct to track accessed, undefined variables, and bail
- render variants in test scripts, etc, and "render" runtime variables such as `${{ PREFIX }}` as `${{ PREFIX }}` (so that they are actually rendered at runtime

wolfv:
Closing as this was merged as part of #1385 and we're waiting for minijinja to come out with the medium-strict-undefined.

Issue #1289, opened 2024-12-27
Feature request: allow specifying a list of values in `context`


mgorny:
I would like to request the ability to specify a list values in `context` rather than just a scalar. This is a bit related to #971, but that bug seems more focused on the Jinja expression being flattened, while I'm interested in ability to use YAML lists.

For example, [torchvision-feedstock](https://github.com/conda-forge/torchvision-feedstock/tree/main) originally skipped tests like this:

```yaml
    # skip test_url_is_accessible instead of hitting 20+ servers per run, since
    # each server might be occasionally unresponsive and end up failing our CI
    {% set tests_to_skip = "test_url_is_accessible" %}
    # spurious failure because upstream skip (Image.__version__ >= "7") does not trigger for Pillow "10"
    {% set tests_to_skip = tests_to_skip + " or (test_transforms and test_adjust_saturation)" %}
    # osx warns with nnpack if there is no AVX2, see conda-forge/pytorch-cpu-feedstock#56
    {% set tests_to_skip = tests_to_skip + " or test_adjust_sharpness" %}  # [osx]
```

which is quite cumbersome, but permits including comments.

What I'd ideally like to be able it to specify:

```yaml
context:
  tests_to_skip:
    # skip test_url_is_accessible instead of hitting 20+ servers per run, since
    # each server might be occasionally unresponsive and end up failing our CI
    - test_url_is_accessible
    # spurious failure because upstream skip (Image.__version__ >= "7") does not trigger for Pillow "10"
    - (test_transforms and test_adjust_saturation)
    # osx warns with nnpack if there is no AVX2, see conda-forge/pytorch-cpu-feedstock#56
    - if: osx
      then: test_adjust_sharpness
```

that would create a Jinja list that I could afterwards combine via `${{ tests_to_skip | join(" or ") }}` or likewise.

I suppose getting conditions in there may be problematic, but even without that part, I think the feature would be really helpful.

For contrast, right now I'm working around the problem via:

```yaml
context:
  tests_to_skip: >
    ${{ 'skip test_url_is_accessible instead of hitting 20+ servers per run, since' if 0 }}
    ${{ 'each server might be occasionally unresponsive and end up failing our CI' if 0 }}
    test_url_is_accessible
    ${{ 'spurious failure because upstream skip (Image.__version__ >= "7") does not trigger for Pillow "10"' if 0 }}
    or (test_transforms and test_adjust_saturation)
[...]

PR #1402, opened 2025-01-30
feat: support lists in "context" (#1289)

mgorny:
Add support for specifying lists in addition to scalars in `context`. They are converted to Jinja sequences.

This is literally the first Rust code I have written, so it's probably horrible. I consider it a miracle that it works. I appreciate any hints to make it cleaner. I also need to add error handling, and tests.

rattler_build_conda_compat needs updating too.

### DIFF
--- a/docs/reference/recipe_file.md
+++ b/docs/reference/recipe_file.md
@@ -987,6 +987,11 @@ context:
   # later keys can reference previous keys
   # and use jinja functions to compute new values
   major_version: ${{ version.split('.')[0] }}
+  tests_to_skip:
+    # fails for one reason
+    - test_foo
+    # fails for another reason
+    - test_bar
 ```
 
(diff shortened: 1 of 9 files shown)


wolfv:
Hey, not bad! Definitely looks like it's going in the right direction. 

I think we should not wrap single values in single-element vectors. That will lead to confusion. Lists should be lists (even single-element ones). 

You can also add some tests in this file: https://github.com/prefix-dev/rattler-build/blob/main/test-data/recipes/jinja-types/recipe.yaml

And update the snapshot after rendering.

mgorny:
Added code to check for list uniformity.

Looks like the CI failures are coming from main.

wolfv (approved):

wolfv:
Great stuff!

PR #1429, opened 2025-02-14
prepend path for build

wolfv:
cc @beckermr

### DIFF
--- a/src/script/interpreter/mod.rs
+++ b/src/script/interpreter/mod.rs
@@ -14,7 +14,7 @@ pub(crate) use python::PythonInterpreter;
 
 use rattler_conda_types::Platform;
 use rattler_shell::{
-    activation::{prefix_path_entries, ActivationError, ActivationVariables, Activator},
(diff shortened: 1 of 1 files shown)


beckermr:
This is not a full fix since in the build environment CONDA_PREFIX will still not be the same as PREFIX.

beckermr:
We need to also not make the build env at all.

wolfv:
Hey @beckermr sorry for just doing it - I would have loved your contribution but it is a little bit in the weeds of things. But maybe it can help you get kick-started into rattler-build development anyways :)

If you want to try this locally, I use:

```
cargo r -- build --recipe ./examples/...
```

I had a test recipe that looked like this:

```yaml
package:
  name: foobar
  version: 1.0.0

build:
  merge_build_and_host_envs: true
  script:
    - echo "foo"
    - exit 22
requirements:
  build:
    - ${{ compiler('c') }}
  host:
    - zlib
```

beckermr:
Ah great for the tips and no problem on just doing. I appreciate the fix!

PR #1435, opened 2025-02-19
feat: extract only run-archive

nichmor:


wolfv:
I'll merge this after #1453 and rebase and remove the updates as the changes in rattler should be released now :)

wolfv:
i have one last question: when actually building a package will this now download the whole package twice? Once to get the run exports, and once to actually build things?
I think we should decide based on a "render-only" mode wether to only download to the run exports cache or whether to actually download & unpack.

wolfv:
I believe this is integrated now.

PR #1448, opened 2025-02-26
Update docs on overlinking and overdepending

### LINKED ISSUE #1447: Documentation on overlinking and overdepending defaults is wrong
The [documentation on overlinking and overdepending](https://rattler.build/latest/build_options/#dynamic-linking-configuration) indicates that the default behaviors are to error. However, this is not currently the case: [the default is to ignore these errors](https://github.com/prefix-dev/rattler-build/blob/main/src/recipe/parser/build.rs#L391). Either the documentation should be fixed, or the default attribute should be moved.

vyasr:
This PR closes #1447. I've chosen to update the documentation rather than change the behavior since changing the default will be a breaking change that will likely cause currently working package builds to fail.

### DIFF
--- a/docs/build_options.md
+++ b/docs/build_options.md
@@ -236,10 +236,10 @@ build:
     missing_dso_allowlist: list of globs
 
     # what to do when detecting overdepending
-    overdepending_behavior: "ignore" or "error" # (defaults to "error")
+    overdepending_behavior: "ignore" or "error" # (defaults to "ignore")
 
     # what to do when detecting overlinking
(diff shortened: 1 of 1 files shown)


wolfv:
thank you! I think to change the default, we will need to make it work better. There are a number of corner cases where the reporting is wrong or too aggressive unfortunately.

vyasr:
Makes sense. Also as I said above I don't think you can make that change without breaking people skating by on the current, laxer behavior, so it'd have to be broadcast somewhat loudly.

PR #1477, opened 2025-03-10
mac: relink 'libfoo.dylib' as '@rpath/libfoo.dylib'

### LINKED ISSUE #1476: BUG: mac relinking doesn't rewrite default install_name in dependencies like conda-build
In a package containing a library and an executable or another library that links it, rattler-build does not rewrite the dependencies in the same way as conda-build, resulting in a failure to load the library downstream (even within the same package). In particular, default `libfoo.dylib` install_names don't get rewritten to `@rpath/libfoo.dylib` as they did in conda-build.

For example, if a package contains `liba.dylib` with install_name `liba.dylib` and executable `b` that links `liba`, at build time it has this `otool -L` output:

```
$PREFIX/lib/liba.dylib:
	liba.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1345.120.2)
$PREFIX/bin/b:
	liba.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1345.120.2)
```

During post processing, `conda-build` rewrites the `liba` links to `@rpath/liba.dylib`:

```
$PREFIX/lib/liba.dylib:
	@rpath/liba.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1345.120.2)
$PREFIX/bin/b:
	@rpath/liba.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1345.120.2)
```

but rattler-build leaves them as they are:

```
$PREFIX/lib/liba.dylib:
	liba.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1345.120.2)
$PREFIX/bin/b:
	liba.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1345.120.2)
```

As a result, running the executable `b` produced by rattler-build crashes due to a failure to find `liba`, while the conda-build package works.

Here's a simple [test recipe](https://github.com/minrk/test-recipe/actions/runs/13762743423) demonstrating the difference, which crashes with rattler-build and succeeds with conda-build.

Normally, setting `install_name` in the package works fine, but that's a bit of a pain [in packages like mumps](https://github.com/conda-forge/mumps-feedstock/pull/133), where one package contains layers of `libdmumps -> libmumps_common -> libmpiseq`. Rewriting with `install_name_tool -id` after building worked with conda-build, but is not sufficient for rattler-build, which also needs `install_name_tool -change foo @rpath/foo` for all intermediate libraries applied to all dependent libraries/executables. This is because the Makefiles don't let us set the right `install_name` for every library at link time, so everything is already linked when it's time to start fixing them.

minrk:
_similar_ to conda-build, but not the same. conda-build [searches a lot more](https://github.com/conda/conda-build/blob/61e9bb24588d8b353321c11de5452d57aa2f85ca/conda_build/post.py#L386-L412), and sets the link to find any matching file anywhere in the package, whereas this only handles the case where it's in $PREFIX/lib.

This also fixes an erroneous check for absolute path install_name for dylibs in $prefix/lib. The check was `strip_prefix(prefix)` whereas it should have been `strip_prefix(encoded_prefix.join("lib"))` if it were to match the path of absolute install_names.

I'm not sure how to add a test for this in this repo, but [this sample recipe](https://github.com/minrk/test-recipe/tree/relink) is fixed by it.

closes #1476

(feel free to ignore my code and rewrite with your own fix, I don't know how to rust).

### DIFF
--- a/src/macos/link.rs
+++ b/src/macos/link.rs
@@ -233,10 +233,29 @@ impl Relinker for Dylib {
         }
 
         let exchange_dylib = |path: &Path| {
-            if let Ok(relpath) = path.strip_prefix(prefix) {
+            let mut _path = path;
+            let path_str = path.to_string_lossy();
+            let encoded_prefix_lib = encoded_prefix.join("lib");
+            let resolved_path = encoded_prefix_lib.join(path);
+            // treat 'libfoo.dylib' the same as $PREFIX/lib/libfoo.dylib
+            // if that's where it is installed
+            if path.parent() == Some(Path::new(""))
+                && !path_str.starts_with("@")
+                && prefix.join("lib").join(path).is_file()
+            {
+                _path = resolved_path.as_path();
+                tracing::debug!(
+                    "Treating relative {} as {}",
+                    path.display(),
+                    _path.display(),
+                );
+            }
+            if let Ok(relpath) = _path.strip_prefix(encoded_prefix_lib) {
+                // absolute $PREFIX/lib/...
                 let new_path = PathBuf::from(format!("@rpath/{}", relpath.to_string_lossy()));
                 Some(new_path)
             } else {
+                tracing::debug!("No need to exchange dylib {}", path.display());
                 None
             }
         };
--- a/src/macos/link.rs
+++ b/src/macos/link.rs
@@ -182,7 +182,7 @@ impl Relinker for Dylib {
                     final_rpaths.push(rpath.clone());
                 }
                 tracing::info!(
-                    "Rpath not in prefix or allow-listed: {} – removing it",
+                    "Rpath not in prefix or allow-listed: {} - removing it",
                     rpath.display()
                 );
             } else if let Ok(rel) = rpath.strip_prefix(encoded_prefix) {
@@ -206,7 +206,7 @@ impl Relinker for Dylib {
                 final_rpaths.push(rpath.clone());
             } else {
                 tracing::info!(
-                    "Rpath not in prefix or allow-listed: {} – removing it",
+                    "Rpath not in prefix or allow-listed: {} - removing it",
                     rpath.display()
                 );
             }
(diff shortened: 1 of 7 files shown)


minrk:
While this correctly rewrites the paths now, I think I may have found a different bug, because this produces libraries that crash on loading if the builtin relink is used. If I skip the builtin and go straight to relinking with `install_name_tool`, everything works correctly.

minrk:
The crashes appear to be related to updating the codesignature (e.g. not doing it):

```
Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGKILL (Code Signature Invalid))
Exception Codes:       UNKNOWN_0x32 at 0x0000000100bd8000
Exception Codes:       0x0000000000000032, 0x0000000100bd8000

Termination Reason:    Namespace CODESIGNING, Code 2 Invalid Page
```

The `install_name` code path works fine, but the rust reimplementation doesn't appear to produce valid dylibs.

wolfv:
Yes, after any relink we need to run codesign again. The `install_name_tool` on conda-forge has some facilities that automatically run codesign, but this is not true for macOS own `install_name_tool` IIRC.

minrk:
ok, after fixing my codesigning issue (#1478), this PR does appear to be working as intended, with both relink implementations. But there are still no tests to exercise `exchange_dylib` or dependency rewriting. What should that look like? I've made a [test recipe](https://github.com/minrk/test-recipe/tree/relink) to work with, but I'm not sure how/what to integrate as tests here.

wolfv:
We have some test binaries, and we can re-read the modified binary to check that the value meets the expectations. 

Maybe you can take the test binary in `test-data/binary-files/zlink-macos`, modify it (e.g. using `install_name_tool`), and turn it into a test?

I'll make sure that codesigning works (and fails) as expected in #1479

wolfv:
Hey @minrk - i pushed some changes to use the actual `rpath` which is most often `lib/` - but can be set to something else by the user.

Can you test your recipe against this change?

minrk:
Yes, this produces the expected result. Thanks!

wolfv:
@minrk would you be able to add your test recipe to `test-data/recipes` and we just run it in `test_simple.py`?

wolfv:
When I run your recipe I think the test is exiting with `1` because the diff is not clean.

minrk:
hm, it passes for me because difft doesn't set the exit code by default. I'll look into adding it. The diff isn't part of the test, it's just for viewing purposes, so not really needed when bringing it into this repo

wolfv:
I think it wasn't passing for me because I didn't pull in the codesign changes. Now it passes. Sorry.

minrk:
Added a slightly simplified version that checks for the expected rewrites with a simple `otool -L | grep`

wolfv:
Thank you so much!

PR #1479, opened 2025-03-11
fix: properly raise error if macOS `codesign` - ing fails

wolfv:
@zbowling unfortunately, conda-forge ships and uses `sigtool` which advertises itself as `codesign` as well, but does not support these extra options. 

My current idea is to:

- use extra options if `/usr/bin/codesign` is the codesign
- not use extra options if we find codesign somewhere else

Unfortunately neither `sigtool "codesign"` nor `codesign` have a `--version` that we could use easily. 

Their `--help` outputs do look differnet though, so we could also use that as indicator.

Or we could first try with the extra args, and if it fails, fall back ... 

Happy to hear other thoughts. Also pinging @minrk who uncovered all of this.

### DIFF
--- a/src/macos/link.rs
+++ b/src/macos/link.rs
@@ -271,23 +271,34 @@ impl Relinker for Dylib {
 }
 
 fn codesign(path: &Path, system_tools: &SystemTools) -> Result<(), RelinkError> {
-    tracing::info!("codesigning {:?}", path.file_name().unwrap_or_default());
-    let output = system_tools
-        .call(Tool::Codesign)?
-        .arg("-f")
-        .arg("-s")
-        .arg("-")
-        // .arg("--preserve-metadata=entitlements,requirements")
-        .arg(path)
-        .output()
-        .map_err(|e| {
-            tracing::error!("codesign failed: {}", e);
-            e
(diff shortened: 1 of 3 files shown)


zbowling:
Oh wowwa! Is this implementation? https://github.com/thefloweringash/sigtool  ? 

Another way to fix this, and it's where I started before I went down this path, was to rather pass entitlements from the yaml all the way down explicitly. I have a branch where I was working on that.  It's unfortunate because it makes the rattler-build more aware of details done by lower level build systems but it would work for sigtool and codesign since they both seem to take an explicit plist for entitlements.

wolfv:
@zbowling do you happen to know if Apple's `codesign` has supported these extra arguments for a long time?

wolfv:
@minrk - just to double check - you manually moved `sigtool` to `/usr/bin/codesign`? Because with these changes we should detect if codesign comes from `/usr/bin` -> use extended arguments - otherwise we should always use the simple args.

minrk:
Almost. I didn't modify the system, but I disabled the `if is_system_codesign` check so the arg is always added to make it fail.

PR #1482, opened 2025-03-11
refactor: Align CLI for s3 with rattler-index

delsner:
Since `rattler-index` and `rattler-build` are typically used together for S3-based channels, they should offer the same CLI.
Additionally, it makes sense to use the more lightweight opendal S3 crate for interactions with S3 buckets (similar to `rattler-index`).


delsner:
Thanks for approving the pull request @wolfv! Could you please merge it? I don't have permissions to do so.

PR #1484, opened 2025-03-12
Document prefix patching options

vyasr:
#598 added these features but did not add them to the documentation.


wolfv:
Wow, this is great 🙏 thanks!

vyasr:
I ran into a case where the prefix update was mangling my binaries and I had to search the repo to find if the feature existed here, so I figured adding docs would be helpful!

PR #1501, opened 2025-03-19
feat: skip upload existing package to prefix

Glatzel:
Just add a subtle feat to skip package uploading if package is existed.

This could be useful in my daily build ci since `pixi search` is not able to find specific version or latest version at this time. 

I also see similar requirement in #1027. But this pr could just resolve partial of it.

Before:

After:
image


Glatzel:
Hi @Hofer-Julian , absolutely not the last one. I have tested this on my local pc.

let me know what do you think of it,and if I need to add test somewhere.

I will revise document before this pr merged.

wolfv (approved):

Issue #1445, opened 2025-02-24
`ignore_run_exports` doesn't work with `${{ compiler('cxx') }}`


pavelzw:
```yml
context:
  name: mypkglibtool
  version: '2.5.4'

package:
  name: mypkg
  version: '0.1.0'

build:
  number: 0

requirements:
  ignore_run_exports:
    from_package:
    - ${{ compiler('cxx') }}
  build:
    - ${{ stdlib('c') }}
    - ${{ compiler('c') }}
    - ${{ compiler('cxx') }}
```

```
c_compiler:
- clang
c_compiler_version:
- '16'
c_stdlib:
- macosx_deployment_target
c_stdlib_version:
- '11.0'
cxx_compiler:
- clangxx
cxx_compiler_version:
- '16'

```

```
❯ rattler-build build -r recipe/ -m recipe/variant.yml

 ╭─ Finding outputs from recipe
 │ Loading variant config file: "recipe/variant.yml"
 │
 ╰─────────────────── (took 0 seconds)
Error:   × Failed to parse recipe

Error:
  × failed to parse package name: 'clangxx_osx-arm64 =16' is not a valid package name. Package names can only contain 0-9, a-z, A-Z, -, _, or .
    ╭─[recipe/recipe.yaml:15:7]
 14 │     from_package:
 15 │     - ${{ compiler('cxx') }}
    ·       ───────────┬──────────
    ·                  ╰── here
 16 │   build:
    ╰────
```

PR #1529, opened 2025-04-06
fix(parser): enhance package name extraction logic in TryConvertNode

zelosleone:
Fixes https://github.com/prefix-dev/rattler-build/issues/1445

```
package:
  name: mypkg
  version: '0.1.0'

build:
  number: 0

requirements:
  ignore_run_exports:
    from_package:
    - ${{ compiler('cxx') }}
  build:
    - ${{ stdlib('c') }}
    - ${{ compiler('c') }}
    - ${{ compiler('cxx') }}
```

```
c_compiler:
- clang
c_compiler_version:
- '16'
c_stdlib:
- macosx_deployment_target
c_stdlib_version:
- '11.0'
cxx_compiler:
- clangxx
cxx_compiler_version:
- '16'
```

Results in (i am on windows btw, just wanted to exactly reproduce the issue so used the same macos deps)

```
 ╭─ Finding outputs from recipe
 │ Loading variant config file: "recipe/issue_variant.yml"
 │ Found 1 variants
 │
 │ Build variant: mypkg-0.1.0-h3ee991f_0
 │
 │ ╭──────────────────────┬────────────────────────────╮
 │ │ Variant              ┆ Version                    │
 │ ╞══════════════════════╪════════════════════════════╡
 │ │ c_compiler           ┆ "clang"                    │
 │ │ c_compiler_version   ┆ "16"                       │
 │ │ c_stdlib             ┆ "macosx_deployment_target" │
 │ │ c_stdlib_version     ┆ "11.0"                     │
 │ │ cxx_compiler         ┆ "clangxx"                  │
 │ │ cxx_compiler_version ┆ "16"                       │
 │ │ target_platform      ┆ "win-64"                   │
 │ ╰──────────────────────┴────────────────────────────╯
 │

 ╭─ Running build for recipe: mypkg-0.1.0-h3ee991f_0
 │
 │ ╭─ Fetching source code
 │ │ No sources to fetch
 │ │
 │ ╰─────────────────── (took 0 seconds)
 │
 │ ╭─ Resolving environments
 │ │
 │ │ Resolving build environment:
 │ │   Platform: win-64 [__win=10.0.26100=0, __cuda=12.8=0, __archspec=1=skylake]
 │ │   Channels:
 │ │    - file:///C:/Users/deniz/rattler-build/output/
 │ │    - conda-forge
 │ │   Specs:
 │ │    - macosx_deployment_target_win-64 11.0.*
 │ │    - clang_win-64 16.*
 │ │    - clangxx_win-64 16.*
 │ │                                                                                                                      
 │ ╰─────────────────── (took 1 second)
 │
 ╰─────────────────── (took 1 second)
Error:   × Failed to resolve dependencies: Cannot solve the request because of: No candidates were found for
  │ macosx_deployment_target_win-64 11.0.*.
  │
  ╰─▶ Cannot solve the request because of: No candidates were found for macosx_deployment_target_win-64 11.0.*.
```

### DIFF
--- a/src/recipe/parser/package.rs
+++ b/src/recipe/parser/package.rs
@@ -167,7 +167,22 @@ impl TryConvertNode<PackageName> for RenderedNode {
 
 impl TryConvertNode<PackageName> for RenderedScalarNode {
     fn try_convert(&self, _name: &str) -> Result<PackageName, Vec<PartialParsingError>> {
-        PackageName::from_str(self.as_str())
+        let input = self.as_str();
+
+        let package_name_str = input
+            .split_once(' ')
+            .and_then(|(name, version)| {
+                if name.contains('_')
+                    && (version.starts_with('=') || version.chars().any(|c| c.is_ascii_digit()))
+                {
+                    Some(name)
+                } else {
+                    None
+                }
+            })
+            .unwrap_or(input);
+
+        PackageName::from_str(package_name_str)
             .map_err(|err| vec![_partialerror!(*self.span(), ErrorKind::from(err),)])
     }
 }
--- a/src/recipe/parser/package.rs
+++ b/src/recipe/parser/package.rs
@@ -1,4 +1,6 @@
-use rattler_conda_types::{MatchSpec, PackageName, ParseStrictness, VersionWithSource};
+use std::str::FromStr;
+
+use rattler_conda_types::{PackageName, VersionWithSource};
 use serde::{Deserialize, Serialize};
 
 use crate::{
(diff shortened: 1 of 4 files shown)


wolfv:
I think we need to special case this. Not every PackageName should be parsed from a string that is _not_ a valid package name. 

To parse these special strings, we can rely on the `MatchSpec` parser, and extract the package name from the parse result. 

The MatchSpec parser can already deal with `=` or other additions. 

For example, we do not want `name: foobar=1234` to parse as only foobar, but we want it to give us a proper error.

However, for `ignore_run_exports`, it would be good to first parse the list into a `Vec<MatchSpec>`, and then extract the package names into a `Vec<PackageName>`.

zelosleone:
@wolfv I played around with the parser with both a hacky solution (git commit being c9b3ec4b4c4fac9206861f50d1e90b3b43cb74e0) and what I think as a proper solution (my last commit), I tried to break things by reverting my changes to `package.rs` and see what can i do, turns out, If i actually manually convert values in this special case, which you pointed out actually thank you for guidance, we can handle jinja rendering more nicely by not rendering brackets to quotes etc.

wolfv:
Cool - maybe you can add a test? We have some end to end tests that you could probably extend.

zelosleone:
@wolfv Added the test as well, on that note though I made assertions different for each build system instead of adding a build flag to force it for linux only, i can change it if its not good though, i was hovering between these and decided to go with doing mac-linux-windows instead, but i am open for feedback there

wolfv (approved):
Awesome!

PR #1545, opened 2025-04-10
feat: add `source.filter` to `PathSource`

### LINKED ISSUE #1542: Allow to filter source files
To cite @wolfv in https://github.com/prefix-dev/pixi-build-backends/issues/45#issuecomment-2790100461

We can copy `files` from `build` and add a:

```
filter:
  includes: [list of globs]
  excludes: [list of globs]
```

And pass them to the copy function then. The globs should be relative to the path that we are copying. E.g. to exclude `.pixi` you would add `exclude: [".pixi"]` etc.

Hofer-Julian:
Fixes #1542

Done:
- Implementation
- Tests
- Docs
- Adapt schema: https://github.com/prefix-dev/recipe-format/pull/53


wolfv (approved):

wolfv:
Nice work!

baszalmstra:
I thought this was used for all sources! But it looks like this is only implemented for the `path` source. Id love to also use this with the url source!

wolfv:
Not sure if that would be the best course of action. If you download multiple (clobbering) urls, you can also decide to download them to different `target_directory`'s and then move, copy, replace or otherwise mutate the files.

I am not sure if it should really be in "rattler-build"'s realm.

PR #1548, opened 2025-04-10
add `--no-deps` to pypi generaed recipe scripts

trim21:


trim21:
if we have this env var, yes, this arguments is not needed...

PR #1551, opened 2025-04-10
refactor: improve stack handling by spawning a dedicated thread

zelosleone:
Essentially, it's the same thing we do here: https://github.com/prefix-dev/pixi/pull/2922

Wanted to preserve the error messages we get from main thread as well, so didn't use an wrapping/unwrapping approach but instead used `map_err` with join to preserve the error messages

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -17,20 +17,42 @@ use rattler_build::{
 use tempfile::{tempdir, TempDir};
 
 fn main() -> miette::Result<()> {
-    // Initialize sandbox in sync/single-threaded context before tokio runtime
-    #[cfg(any(
-        all(target_os = "linux", target_arch = "x86_64"),
-        all(target_os = "linux", target_arch = "aarch64"),
-        target_os = "macos"
-    ))]
-    rattler_sandbox::init_sandbox();
+    // Stack size varies significantly across platforms:
+    // - Windows: only 1MB by default
+    // - macOS/Linux: ~8MB by default
+    //
+    // This discrepancy causes stack overflows primarily on Windows, especially in debug builds
+    // To address this, we spawn another main thread (main2) with a consistent
+    // larger stack size across all platforms.
+    //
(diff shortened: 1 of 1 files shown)


wolfv:
Looks good although I think the sandbox still needs to be initialized outside any thread context :) if you could restore that, that would be great.

zelosleone:
> Looks good although I think the sandbox still needs to be initialized outside any thread context :) if you could restore that, that would be great.

I did it but please correct me if i understood it wrongly:

I first thought rust in general is pretty memory-safe and sandbox within thread would be enough but sandbox in this context i guess is for dependencies - build scripts that are actually external, right? This is better than cargo if so, nice.

wolfv:
Yes, so this is about process-level sandboxing of any process that is triggered. What the `rattler_sandbox` does is it uses `birdcage` (https://github.com/phylum-dev/birdcage/) and sandboxes new processes with it. However, on Linux, the sandboxing needs to happen in a single-threaded environment. 

What we do is we add a hidden command to `rattler-build`, something like `rattler-build __sandbox__ --read-write=foobar ./mycommand` that will run `mycommand` with the sandbox restrictions applied. But to set this up, we need to do it before starting any threads. 

You can look at the `rattler_sandbox` crate if you want to peek under the hood.

zelosleone:
> Yes, so this is about process-level sandboxing of any process that is triggered. What the `rattler_sandbox` does is it uses `birdcage` ([phylum-dev/birdcage](https://github.com/phylum-dev/birdcage/)) and sandboxes new processes with it. However, on Linux, the sandboxing needs to happen in a single-threaded environment.
> 
> What we do is we add a hidden command to `rattler-build`, something like `rattler-build __sandbox__ --read-write=foobar ./mycommand` that will run `mycommand` with the sandbox restrictions applied. But to set this up, we need to do it before starting any threads.
> 
> You can look at the `rattler_sandbox` crate if you want to peek under the hood.

Thank you! I will definitely check out rattler_sandbox crate as well!

PR #1557, opened 2025-04-14
feat: add error handling for hypen in context variables

### LINKED ISSUE #995: Disallow `-` in context variable names
Since `-` is also the `minus` operator which will probably not work in jinja, we should forbid it in  the left-hand side in the context, e.g.

```yaml
context:
  foo-bar: baz
```

should trigger a parsing error.

zelosleone:
closes #995

the issue was only mentioned the hyphen, i though dots would have a problem with jinja as well but i saw we already use `(jinja, &format!("context.{}", k.as_str()))?;` so i didn't continue with adding dots


wolfv (approved):
Great PR!

PR #1559, opened 2025-04-14
feat: add Windows support for DLL linking and validation

### LINKED ISSUE #629: Implement linking checks for Windows
While we don't do any relinking / binary patching on Windows, it's still very useful to have the overlinking / overdependening checks. 

Goblin is able to parse `PE` files fine, so we should have an easy time to implement the checks on Windows, too.

zelosleone:
closes #629


wolfv:
For the test, we only need `zstd.dll`, correct? Can we remove all the other files?

PR #1563, opened 2025-04-15
feat: Read channels and package format from pixi config

pavelzw:


pavelzw:
> Would be nice to consume pixi-config from crates.io.

https://github.com/prefix-dev/pixi/issues/3691

wolfv (approved):

PR #1564, opened 2025-04-16
feat: validate python_version specs in recipe parser

### LINKED ISSUE #1493: `rattler-build build` panics when `test.python.python_version` match spec is invalid
### What happened?

While trying to build a rattler-build recipe, I received the following error:

```
thread 'main' panicked at src/package_test/run_test.rs:511:26:
called `Result::unwrap()` on an `Err` value: InvalidVersionAndBuild("==.*")
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This was due to an invalid match spec in the testing section of my recipe. Here's what that looked like:

```yaml
    tests:
      - python:
          imports:
            - my_package
          python_version: ["${{ python_min ~ '.*' }}", "3.12.*"]
```

The `python_version` section was generating an invalid match spec.

### How to recreate

First, create this recipe (`recipe.yaml`):

```yaml
package:
  name: test
  version: 0.1.0

build:
  script:
    - echo "test" > test.txt

requirements:
  build:
    - python

tests:
  - python:
      imports:
        - time
      python_version: "=.*"
```

Then try to build it:

```
rattler-build build 
```

It should spit out the panic error.

### What should be done?

I think it would be nice to trigger a better looking error than just having the program panic. Maybe something that could try to hint at which line of the `recipe.yaml` that should be changed?

zelosleone:
closes #1493

I also change the args of the abi3 test to use --recipe, for some reason it would pick up the recipe file on the root folder without --recipe command. After improving the parser to use matchspec and making it more strict, it would just fail (since it was not using the correct recipe)

### DIFF
--- a/src/recipe/parser/test.rs
+++ b/src/recipe/parser/test.rs
@@ -549,11 +549,11 @@ mod test {
           - python:
               imports:
                 - pandas
-              python_version: "3.10"
+              python_version: ">=3.10"
           - python:
               imports:
                 - pandas
-              python_version: ["3.10", "3.12"]
+              python_version: [">=3.10", ">=3.12"]
         "#;
 
         // parse the YAML
@@ -576,7 +576,7 @@ mod test {
                 assert!(python.pip_check);
                 assert_eq!(
                     python.python_version,
-                    PythonVersion::Single("3.10".to_string())
+                    PythonVersion::Single(">=3.10".to_string())
                 );
             }
             _ => panic!("expected python test"),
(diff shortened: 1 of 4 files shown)


src/recipe/parser/test.rs
```diff
@@ -304,18 +304,38 @@ impl TryConvertNode<PythonVersion> for RenderedNode {
...
             )])?,
-            RenderedNode::Scalar(version) => PythonVersion::Single(version.to_string()),
+            RenderedNode::Scalar(version) => {
+                let version_str = version.to_string();
+                if rattler_conda_types::MatchSpec::from_str(&version_str, rattler_conda_types::ParseStrictness::Strict).is_err() {
+                    return Err(vec![_partialerror!(
+                        *self.span(),
+                        ErrorKind::InvalidValue(("python_version".to_string(), "Invalid version specification".into())),
+                        label = format!("invalid python version specification: {}", version_str),
+                        help = "Use a valid version specification like '3.12', '3.12.*', or '>=3.12'"
```
wolfv, line 314: I am quite sure that `3.12` will not parse with the Strict mode, so we should not mention it here :) You could mention `==3.12` but I don't think that's what most people want.

wolfv (approved):

PR #1566, opened 2025-04-16
feat: introduce debug mode, custom outputs

### LINKED ISSUE #1438: cannot properly capture stderr and stdout in commands run in subshells via backticks for tests
I have a test script does

```
echo -n "setting up 'pex_exceptions' ... "
val=`setup -j pex_exceptions 2>&1`
if [[ ! ${val} ]]; then
    echo "worked!"
else
    echo "failed!"
    exit 1
fi
```

Running it results in

```
 | + echo -n 'setting up '\''pex_exceptions'\'' ... '
 │ ++ setup -j pex_exceptions
 │ + val='+++ $PREFIX/eups/bin/eups_setup DYLD_LIBRARY_PATH= -j pex_exceptions
          ^ this bit is rattler build somehow outputing subshell-esque items it should not be...
```

Apparently rattler-build is somehow injecting its own stderr/stdout into the subshell's one.

zelosleone:
closes #1438

this is not much of a breaking change, but purely for capturing the output better we limit the debug output to stderr using exec

now it will properly output it as
```

 │ │ + echo -n 'setting up '\''pex_exceptions'\'' ... '
 │ │ ++ setup -j pex_exceptions
 │ │ + val='$SRC_DIR/conda_build.sh: line 12: setup: command not found'
 │ │ setting up 'pex_exceptions' ... 
```

the original issue didn't had the recipe, so i wanted to make a breaking one using bash with:

```

package:
  name: test-subshell
  version: 0.1.0

requirements:
  build:
    - bash

build:
  number: 0
  script: |
    echo -n "setting up 'pex_exceptions' ... "
    val=`setup -j pex_exceptions 2>&1`
    if [[ ! ${val} ]]; then
        echo "worked!"
    else
        echo "failed!"
        exit 1
    fi
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -127,9 +127,11 @@ pub fn get_tool_config(
     build_data: &BuildData,
     fancy_log_handler: &Option<LoggingOutputHandler>,
 ) -> miette::Result<Configuration> {
-    let client =
-        tool_configuration::reqwest_client_from_auth_storage(build_data.common.auth_file.clone())
-            .into_diagnostic()?;
+    let client = tool_configuration::reqwest_client_from_auth_storage(
+        build_data.common.auth_file.clone(),
+        build_data.common.allow_insecure_host.clone(),
+    )
+    .into_diagnostic()?;
 
     let configuration_builder = Configuration::builder()
         .with_keep_build(build_data.keep_build)
@@ -138,7 +140,8 @@ pub fn get_tool_config(
         .with_test_strategy(build_data.test)
         .with_skip_existing(build_data.skip_existing)
         .with_noarch_build_platform(build_data.noarch_build_platform)
-        .with_channel_priority(build_data.common.channel_priority);
+        .with_channel_priority(build_data.common.channel_priority)
+        .with_allow_insecure_host(build_data.common.allow_insecure_host.clone());
 
     let configuration_builder = if let Some(fancy_log_handler) = fancy_log_handler {
         configuration_builder.with_logging_output_handler(fancy_log_handler.clone())
@@ -607,8 +610,11 @@ pub async fn run_test(
         .with_keep_build(true)
         .with_compression_threads(test_data.compression_threads)
         .with_reqwest_client(
-            tool_configuration::reqwest_client_from_auth_storage(test_data.common.auth_file)
-                .into_diagnostic()?,
+            tool_configuration::reqwest_client_from_auth_storage(
+                test_data.common.auth_file,
+                test_data.common.allow_insecure_host.clone(),
+            )
+            .into_diagnostic()?,
         )
         .with_channel_priority(test_data.common.channel_priority)
         .finish();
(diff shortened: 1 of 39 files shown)


src/lib.rs
```diff
@@ -893,3 +895,91 @@ pub async fn build_recipes(
...
+        sandbox_configuration: None,
+    };
+
+    let tool_config = get_tool_config(&build_data, log_handler)?;
+
+    let outputs = get_build_output(&build_data, &recipe_path, &tool_config).await?;
+
+    tracing::info!("Build and/or host environments created for debugging.");
+
+    for output in outputs {
```
wolfv, line 942: In the debug data we should have a `--output-name` field that allows you to specify which output you want to debug. We will then only set up that particular output. 

If there is only one output, that flag is optional.

src/lib.rs
```diff
@@ -893,3 +895,91 @@ pub async fn build_recipes(
...
+            }
+            if deps.host.is_some() {
+                tracing::info!(
+                    "Host dependencies available in {}",
+                    output.build_configuration.directories.host_prefix.display()
+                );
+            }
+        }
+
+        tracing::info!("\nTo run the actual build, use:");
```
wolfv, line 977: Here we should print the instructions to find the "WORK_DIR" inside the environment, so that the user can `cd` there and run `bash conda_build.sh` etc. to debug.

beckermr (changes requested):
This will fix the problem for non-debug builds but still leaves the issue for debug builds.

wolfv:
At least all I see is that they are also passing `-x` ...

beckermr:
OK yes replicating the behavior of conda-build will be fine. Thank you!

beckermr (approved):

wolfv:
I just tested `rattler-build debug --recipe ./examples/curl` and it works pretty well!
The one thing that is missing is that it did not create the `build_env.sh` and `conda_build.sh` scripts. We should create those scripts, but not execute them before putting out the message to the user.

src/script/mod.rs
```diff
@@ -413,6 +415,86 @@ impl Output {
 
         Ok(())
     }
+
+    /// Create build scripts without running them
+    pub async fn create_build_script(&self) -> Result<(), std::io::Error> {
```
wolfv, line 420: Is it possible to remove code duplication for this feature and create the script with the original code but just skip execution, perhaps?

src/metadata.rs
```diff
@@ -311,6 +311,34 @@ impl<'de> Deserialize<'de> for PlatformWithVirtualPackages {
...
+        Self(debug)
+    }
+
+    /// Returns true if debug output is enabled
+    pub fn is_enabled(&self) -> bool {
+        self.0
+    }
+}
+
+impl From<bool> for Debug {
```
wolfv, line 330: Can we remove these methods? Both `From<bool>` and `From<Debug>`? Then we have even more type safety! :)

PR #1570, opened 2025-04-17
fix: Add test for bad package versions with hatch-vcs

pavelzw:


wolfv:
@pavelzw I think you should look into fixing this in rattler: https://github.com/conda/rattler/pull/1263

wolfv:
pushed what I think shoudl be the fix :)

Issue #1569, opened 2025-04-17
pin_subpackage incompatible with setuptools-scm or hatch-vcs


pavelzw:
setuptools-scm produces version strings like `0.46.1.post1+g703fbb0`.

```yaml
outputs:
  - package:
      name: abc
      version: '0.46.1.post1+g703fbb0'
  - package:
      name: abc-sub
      version: '0.1.0'
    requirements:
      run:
        - ${{ pin_subpackage('abc', lower_bound="x.x.x", upper_bound="x.x.x") }}
    tests:
      - script: echo hello
```

here, i see bad run dependencies

```
❯ rm -rf output && rattler-build build -r recipe/
...
 │ │ Run dependencies:
 │ │ ╭──────────────────┬────────────────────────────────────╮
 │ │ │ Name             ┆ Spec                               │
 │ │ ╞══════════════════╪════════════════════════════════════╡
 │ │ │ Run dependencies ┆                                    │
 │ │ │ abc              ┆ >=0.46.1+g703fbb0,<0.46.2.0a0 (PS) │
 │ │ ╰──────────────────┴────────────────────────────────────╯
...
```

---

hatch-vcs produces version strings like `0.46.1.dev1+g703fbb0`.

```yaml
outputs:
  - package:
      name: abc
      version: '0.46.1.dev1+g703fbb0'
  - package:
      name: abc-sub
      version: '0.1.0'
    requirements:
      run:
        - ${{ pin_subpackage('abc', lower_bound="x.x.x", upper_bound="x.x.x") }}
    tests:
      - script: echo hello
```

```
❯ rm -rf output && rattler-build build -r recipe/
...
  │ └─ abc-sub 0.1.0 would require
  │    └─ abc >=0.46.1+g703fbb0,<0.46.2.0a0, for which no candidates were found.
  │ The following packages are incompatible
  │ └─ abc-sub ==0.1.0 h60d57d3_0 cannot be installed because there are no viable options:
  │    └─ abc-sub 0.1.0 is excluded because due to strict channel priority not using this option from: 'file:///private/var/folders/rz/
  │ q1r4tv0n6ll38q18fbk419mc0000gn/T/tmp.5JOJmbQhcx/output/'
  │
  ╰─▶ Cannot solve the request because of: abc-sub ==0.1.0 h60d57d3_0 cannot be installed because there are no viable options:
      └─ abc-sub 0.1.0 would require
[...]

PR #1572, opened 2025-04-17
fix: more `VersionWithSource` in additional places

wolfv:
cc @pavelzw does that fix your pin_subpackage issue?


pavelzw:
nope, still the same behavior as described in https://github.com/prefix-dev/rattler-build/issues/1569

PR #1575, opened 2025-04-21
feat: add CarriageReturnToNewline wrapper for async reading

### LINKED ISSUE #1525: carriage-return build output never displayed
Some build processes overwrite lines with carriage returns (CR) instead of newlines (LF), to avoid flooding interactive terminals with a lot of steps as they go by. With rattler-build, this output is never displayed.

conda-build and many such tools (I think including most CI systems) treat CR as LF, so no output is lost from the build logs. rattler-build, on the other hand, appears to never show any output until LF is seen, meaning that a long-running configure step (e.g. petsc's) is completely silent with rattler-build, whereas it produces quite a bit of informative output with conda-build.

An example build.sh to see the difference:

```bash
#!/usr/bin/env bash
set +x
set -eu

for i in {1..10}; do
  echo -ne "\rline ${i}"
  sleep 1
done
echo -e "\ndone"
```

Output from rattler-build:

```
[10 seconds of silence]
line 10
done
```

conda-build:

```
line 1
line 2
line 3
line 4
line 5
line 6
line 7
line 8
line 9
line 10
done
```

and [test recipe](https://github.com/minrk/test-recipe/tree/carriage-return).

zelosleone:
closes #1525

I needed some time to plan how to approach this. I had two options:

- Use async
- Use sync, spawn threads that wait for every line for output to reach them and return immediately

After some thinking and research specially reading this discussion https://users.rust-lang.org/t/avoid-async-rust-at-all-costs-comments-from-experts/105860/44 I decided to go with async approach since what we are dealing with about script's run returns are a lot close to network operations approach. So I decided to go with async approach. Checked the overall integration test result time, surprisingly it went from 118 seconds to 109 seconds now.

output: 
image

On the visual changes, the old version would display the script's end line straight in blue line (they would overlap with each other) so that's fixed as well

### DIFF
--- a/src/script/mod.rs
+++ b/src/script/mod.rs
@@ -18,9 +18,11 @@ use std::ffi::OsStr;
 use std::{
     collections::HashMap,
     path::{Path, PathBuf},
+    pin::Pin,
     process::Stdio,
+    task::{Context, Poll},
 };
-use tokio::io::AsyncBufReadExt as _;
+use tokio::io::{AsyncBufReadExt, AsyncRead, ReadBuf};
 
 use crate::{
     env_vars::{self},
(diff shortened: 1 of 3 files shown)


zelosleone:
Actually pretty nice considering these UNIX systems don't have real async

Mac tests on github runners went from 5m 16s to 4m 57s
Ubuntu tests on github runners went from 5m 26s to 5m 21s

Windows runner seemed to increase its time though, but on my local computer with default settings it decreased the time needed, so i think it's more about github small runner struggling with async function due to its low capacity with 2/4 CPU cores, in that case definitely sync would be better but hard to pass the benefit we gain from async here.

wolfv (approved):

minrk:
> what happens when programs print \r\n - will we get two newlines?

Yes, `\r\n` will produce  `\n\n` in this implementation. You'd need a look-ahead to the next character and handle a change in length  of the buffer to avoid this. Python's universal newline implementation [looks like this](https://github.com/python/cpython/blob/e84624450dc0494271119018c699372245d724d9/Objects/fileobject.c#L230-L262). I'm not sure how that would be spelled in rust. In a regex, it's `/\r\n?/\n/` (or even `/\r+\n?/\n/` which would look more like terminal output, though not what Python's universal newlines does).

If the change in buffer length is a problem, you might get a byte-for-byte match if you replace `\r\n` with `\0\n`, so for each `\r`, sub:

- `\0` if next char is `\n`
- otherwise `\n`

PR #1577, opened 2025-04-22
chore: remove outdated URL from SSL test recipe

zelosleone:
@wolfv seems like their host is down so github runners are struggling, we are using badssl anyways so i removed them from recipe


wolfv (approved):

wolfv:
Works for me :)

PR #1579, opened 2025-04-22
feat: optimize CarriageReturnToNewline for better performance with buffer processing

zelosleone:
this pr directly adds the condition to replace \r with \n only if the following character is not \n

also added some checks, like for empty buffers we just skip them so they dont overload the loop, plus i made sure to use smaller buffer threshold (32 bytes) which significantly sped up the output printing. since the github windows runner increased its time, we now don't scan chunks one by one we just directly process them. 

results are incredibly interesting, hopefully its not just my local machine

from 176.24 sec to 123 sec now

@wolfv would be perfect if you also tested it locally to check the e2e tests time

### DIFF
--- a/src/script/mod.rs
+++ b/src/script/mod.rs
@@ -4,6 +4,7 @@ mod sandbox;
 pub use sandbox::{SandboxArguments, SandboxConfiguration};
 
 use crate::script::interpreter::Interpreter;
+use futures::TryStreamExt;
 use indexmap::IndexMap;
 use interpreter::{
     BashInterpreter, CmdExeInterpreter, NuShellInterpreter, PerlInterpreter, PythonInterpreter,
@@ -15,14 +16,17 @@ use rattler_conda_types::Platform;
 use rattler_shell::shell;
 use std::collections::HashSet;
 use std::ffi::OsStr;
+use std::io;
 use std::{
     collections::HashMap,
     path::{Path, PathBuf},
-    pin::Pin,
     process::Stdio,
-    task::{Context, Poll},
 };
-use tokio::io::{AsyncBufReadExt, AsyncRead, ReadBuf};
+use tokio::io::{AsyncBufReadExt, AsyncRead};
+use tokio_util::bytes::{BufMut, BytesMut};
+use tokio_util::codec::Decoder;
+use tokio_util::codec::FramedRead;
+use tokio_util::compat::FuturesAsyncReadCompatExt;
 
 use crate::{
     env_vars::{self},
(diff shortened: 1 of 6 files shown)


src/script/mod.rs
```diff
@@ -547,12 +547,45 @@ impl<R: AsyncRead + Unpin> AsyncRead for CarriageReturnToNewline<R> {
...
+                        continue;
+                    }
+
+                    while i < chunk_end {
+                        if newly_filled[i] == b'\r'
+                            && (i + 1 >= len || newly_filled[i + 1] != b'\n')
+                        {
+                            newly_filled[i] = b'\n';
+                        }
+                        i += 1;
```
wolfv, line 585: I think this still needs to _erase_ `\r` if necessary, right?

src/script/mod.rs
```diff
@@ -547,12 +547,45 @@ impl<R: AsyncRead + Unpin> AsyncRead for CarriageReturnToNewline<R> {
...
+                if len <= 32 {
+                    let has_cr = newly_filled.iter().take(len).any(|&byte| byte == b'\r');
+
+                    if !has_cr {
+                        return Poll::Ready(Ok(()));
+                    }
+                }
+
+                let mut i = 0;
+                while i < len {
```
wolfv, line 566: I think using Rust native `chunks()` iterator would look nicer.

E.g. 

```rust
let chunk_size = 64;
for chunk in newly_filled.chunks(chunk_size) {
    ...
}
```

src/script/mod.rs
```diff
@@ -530,33 +534,85 @@ impl Output {
...
+                    break;
+                }
+                b'\r' => {
+                    result.put_u8(b'\n');
+                    read_idx += 1;
+                }
+                b => {
+                    result.put_u8(b);
+                    read_idx += 1;
                 }
```
baszalmstra, line 590:
```suggestion
        while read_idx < src.len() {        
            match src[read_idx] {
                '\r' => {  
                    // Skip it.
                }
                b => {
                    if self.pending_cr && b != '\n' {
                       result.put_u8('\r');
                    }
                    result.put_u8(b);
                }
             }
             self.pending_cr = src[read_idx] == '\r';
             self.read_idx += 1;
```

wolfv:
Looks good

baszalmstra:
What was broken rhat you had to revert?

src/script/mod.rs
```diff
@@ -530,33 +534,83 @@ impl Output {
...
+        if src.is_empty() {
+            return if std::mem::replace(&mut self.pending_cr, false) {
+                Ok(Some(BytesMut::from(&b"\n"[..])))
+            } else {
+                Ok(None)
+            };
+        }
+
+        self.output_buffer.clear();
+        self.output_buffer
```
baszalmstra, line 564: Why subtract the capacity?

src/script/mod.rs
```diff
@@ -530,33 +534,83 @@ impl Output {
...
+            .reserve(src.len().saturating_sub(self.output_buffer.capacity()));
+        let mut read_idx = 0;
+
+        if std::mem::replace(&mut self.pending_cr, false) {
```
baszalmstra, line 568: Why is this block needed?

src/script/mod.rs
```diff
@@ -530,33 +534,83 @@ impl Output {
...
+        let _processed = src.split_to(read_idx);
+
+        let has_output = !self.output_buffer.is_empty();
+        Ok(has_output.then(|| self.output_buffer.split().freeze().into()))
+    }
+
+    fn decode_eof(&mut self, buf: &mut BytesMut) -> Result<Option<Self::Item>, Self::Error> {
+        let final_chunk = self.decode(buf)?;
+
+        if std::mem::replace(&mut self.pending_cr, false) {
```
baszalmstra, line 603: Is this correct? You are not processing any additional \r in the final block?

baszalmstra (commented):
I think we should add a nimber of unit tests to verify the correct behavior

src/script/mod.rs
```diff
@@ -530,33 +534,83 @@ impl Output {
...
+                self.output_buffer.put_u8(b);
             }
-            Poll::Ready(Err(e)) => Poll::Ready(Err(e)),
-            Poll::Pending => Poll::Pending,
+        }
+
+        let _processed = src.split_to(read_idx);
+
+        let has_output = !self.output_buffer.is_empty();
+        Ok(has_output.then(|| self.output_buffer.split().freeze().into()))
```
baszalmstra, line 597: You can also use std::mem::take instead of cloning the bytes

zelosleone:
> What was broken rhat you had to revert?

Basically the original approach of #1525 was not working correctly, I made a test on #linebreak folder with a recipe that has scripts on windows/unix as well, so lines were not showing correctly after the approach and it was back to old behavior

baszalmstra:
No I mean why did you revert the “simplify” commit? What was broken about it? 

I would really like to see some unit tests..

src/script/mod.rs
zelosleone: yes, in general i think decode method should take care of this in my understanding, since it will stop processing them if true before. i will try to add an unit test to see if my logic is correct

src/script/mod.rs
zelosleone: thanks for this, will do

src/script/mod.rs
zelosleone: it checks the previous buffer for \r lines, so basically we check when \r is at the end of the buffer and on the next buffer \n is at the beginning

src/script/mod.rs
zelosleone: i think it was required to check the memory capacity during buffers to avoid overfilling it. so only when we see a difference we use it, which should be pretty fast and performance wise p good as well

zelosleone:
> I think we should add a nimber of unit tests to verify the correct behavior

Will do this as well, thank you!

zelosleone:
> No I mean why did you revert the “simplify” commit? What was broken about it?
> 
> I would really like to see some unit tests..

Ah, you can run the recipe on line breaks with reverted commit and your commit, script results should have 10 lines shown, there are end to end tests for it

zelosleone:
with simplified commit:
image

with revert:
image

sorry for the light theme :(


`cargo run -p rattler-build -- build --recipe test-data/recipes/line-breaks/recipe.yaml` should show it

zelosleone:
@baszalmstra I added the unit tests as well, also tried with the simplified commit on the unit tests and results with simplified commit:

```
---- script::tests::test_crlf_normalizer_with_multiple_cr_at_end stdout ----
thread 'script::tests::test_crlf_normalizer_with_multiple_cr_at_end' panicked at src\script\mod.rs:805:9:
assertion `left == right` failed
  left: b"line1\r\r"
 right: "line1\n\n"

---- script::tests::test_crlf_normalizer_with_cr_only stdout ----
thread 'script::tests::test_crlf_normalizer_with_cr_only' panicked at src\script\mod.rs:757:9:
assertion `left == right` failed
  left: b"line1\rline2\rline3"
 right: "line1\nline2\nline3"
```

zelosleone:
> I guess I misunderstood the assignment then because those test outputs look correct to me?
> 
> I understood that if a \r without a \n behind it was encountered it would simply retain the \r.

No no, it was my mistake to not to explain the core issue. Specially with the manual testing reports, I simply informed as "edge cases" but could have worded it better to you about how lines before final line isn't showing on the script and I didn't even mention about how we have a script + recipe to test it. Sorry for that really, could definitely have worded better and informed you before taking your time for simplifying the logic

baszalmstra:
I just pushed a change that passes all tests and is much simpler. :)

baszalmstra:
And now it doesnt even need to allocate any additional bytes. :) 

@zelosleone I was mostly surprised that you just reverted the code without checking what the actual problem was. But Im happy you added tests to verify it!

zelosleone:
> I just pushed a change that passes all tests and is much simpler. :)

One day i will reach this level 🥹 I need to study your solution now so that i know why yours worked with simplified and elegant code :D 

I couldn't understand the problem of why the tests would fail without overcomplicating it, mainly because of that and trying to solve it would imply i would mess up your solution to add more complexity because honestly the only thing i could think of was adding more if/else to it. So instead, since you also wanted to check the tests, I added the tests. Thank you Bas, you are definitely a rock star :)

wolfv (approved):

baszalmstra:
Would be curious to any speed improvements this gives you now @zelosleone

zelosleone:
> Would be curious to any speed improvements this gives you now @zelosleone

Github runners seems to improved actually but on local there is a decrease in improvement

167 seconds of full runtime of the e2e tests in this branch now
116 seconds of full runtime of the e2e tests in main branch x.x

zelosleone:
Okay, I did not expect this improvement on github runners as well. From formatting (pre commit) to e2e tests to CI/tests etc all jobs improved in runtime optimization!?

src/script/mod.rs
```diff
@@ -530,34 +535,40 @@ impl Output {
...
+    fn decode(&mut self, src: &mut BytesMut) -> Result<Option<Self::Item>, Self::Error> {
+        if src.is_empty() {
+            return Ok(None);
+        }
+        let mut bytes = src.split_off(0);
+        let mut w = 0;
+        for r in 0..bytes.len() {
+            self.last_was_cr = match (bytes[r], self.last_was_cr) {
+                (b'\n', true) => false,
+                (b, _) => {
```
baszalmstra, line 563: You are again mixing match and if statements here. I find that harder to understand. Just use multiple matches. You are also checking if the byte is a CR twice.

baszalmstra (commented):
I strongly doubt that this code has such a big impact on the performance. You are measuring the e2e tests which has lots of variables that affect its performance. Unless this part of the code is 90% of the overall runtime small micro optimizations around caches and such are never gonna contribute a major speedup.

I also think it makes the code less readable which is ultimately what I care about most. Thats why Im hammering on simplicity.

If you really want to squeeze out every bit of performance here you should write a benchmark that feeds this code lots of text to get a much better representation of speed.

zelosleone:
> I strongly doubt that this code has such a big impact on the performance. You are measuring the e2e tests which has lots of variables that affect its performance. Unless this part of the code is 90% of the overall runtime small micro optimizations around caches and such are never gonna contribute a major speedup.
> 
> I also think it makes the code less readable which is ultimately what I care about most. Thats why Im hammering on simplicity.
> 
> If you really want to squeeze out every bit of performance here you should write a benchmark that feeds this code lots of text to get a much better representation of speed.

Okay, after writing a benchmark https://gist.github.com/zelosleone/cdde5003385c75390c93286ace6cb16e and using it, i can 100% say its just bas's solution was slower on e2e tests and prob network issues, and some others contributed to it. its perfectly valid and in fact, 35% beats my solution. I will be switching to Bas's solution

baszalmstra (approved):
Nice! Feel free to also add the benchmark!

zelosleone:
> Nice! Feel free to also add the benchmark!

Added them as well

PR #1581, opened 2025-04-23
feat: add vswhere recipe and ensure environment variable directories exist

### LINKED ISSUE #1576: Repackaging `.exe` file harder than it should be
There seems to be a little bug when using a `.exe` file as the source for a package.

I have tried the following recipe to repackage a simple `.exe` file as a Conda package but it fails to find the file in the work-dir.

The same recipe used to work with `conda-build`.

```yaml
context:
  version: "3.1.7"

package:
  name: vswhere
  version: ${{ version }}

source:
  url: https://github.com/microsoft/vswhere/releases/download/${{ version }}/vswhere.exe
  sha256: c54f3b7c9164ea9a0db8641e81ecdda80c2664ef5a47c4191406f848cc07c662
  file_name: vswhere.exe  # this didn't change anything

build:
  number: 0
  script: MOVE vswhere.exe %LIBRARY_BIN%\vswhere.exe  # here it fails to find `vswhere.exe`
  skip: not win

tests:
  - script:
      - vswhere

about:
  homepage: https://github.com/microsoft/vswhere
  license: MIT
  license_file: LICENSE.txt
  summary: CLI tool to locate Visual Studio 2017 and newer installations
  description: |
    vswhere is designed to be a redistributable, single-file executable that
    can be used in build or deployment scripts to find where Visual Studio
    - or other products in the Visual Studio family - is located.

extra:
  recipe-maintainers:
    - wolfv
```

zelosleone:
closes #1576

while manually debugging i noticed and mentioned to @wolfv mkdir fixes the recipe, so checked further on why path envs were not being created as a folder even though they are being SET in the build script, turns out none of them were being created. I filter out the basic system information that is being SET in build script while trying to create the default PREFIX path folders here

### DIFF
--- a/src/render/solver.rs
+++ b/src/render/solver.rs
@@ -313,6 +313,16 @@ pub async fn install_packages(
             )
         })?;
 
+    // Ensure required environment variable directories exist
+    let env_vars = crate::env_vars::os_vars(target_prefix, &target_platform);
+
+    // Create directories for path variables
+    for dir in env_vars.values().flatten() {
+        if dir.contains('/') || dir.contains('\\') {
+            let _ = tokio::fs::create_dir_all(dir).await;
+        }
+    }
+
     let installed_packages = PrefixRecord::collect_from_prefix(target_prefix)?;
 
     if !installed_packages.is_empty() && name.starts_with("host") {
--- a/test-data/recipes/package-content-tests/recipe-test-succeed.yaml
+++ b/test-data/recipes/package-content-tests/recipe-test-succeed.yaml
@@ -5,8 +5,13 @@ package:
 build:
   number: 0
   script:
-    - echo "Hello World" > %PREFIX%/test-execution.txt
-    - echo "Hello World" > %PREFIX%/Library/bin/rust.exe
+    - if: unix
+      then:
+        - echo "Hello World" > $PREFIX/test-execution.txt
+        - echo "Hello World" > $PREFIX/bin/rust
+      else:
+        - echo "Hello World" > %PREFIX%\test-execution.txt
+        - echo "Hello World" > %PREFIX%\Library\bin\rust.exe
 
 tests:
   - package_contents:
--- a/test-data/recipes/package-content-tests/recipe-test-succeed.yaml
+++ b/test-data/recipes/package-content-tests/recipe-test-succeed.yaml
@@ -3,14 +3,10 @@ package:
   version: 0.1.0
 
 build:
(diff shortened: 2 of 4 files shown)


zelosleone:
@wolfv i think the changes when package-content-tests were added broke that vswhere recipe, because in test-data/recipes/package-content-tests we also try to create the lib bin folder via mkdir commands. Updated the recipe file of that test as well

wolfv:
Hmmmmmmmm, I am not sure that is correct to be honest. I need to double check if `conda-build` does anything like that.

wolfv:
Just to be sure, everything works if you change the recipe to run `mkdir %LIBRARY_BIN%` before the `MOVE`?

zelosleone:
> Just to be sure, everything works if you change the recipe to run `mkdir %LIBRARY_BIN%` before the `MOVE`?

Yup

wolfv:
Can you check - if you run this recipe with conda-build (https://github.com/conda-forge/vswhere-feedstock) - whether the folder exists at the beginning of the build (e.g. using `dir %LIBRARY_BIN%`) or something like that?

You can install `conda-build` with `pixi global install conda-build`.

wolfv:
OK, idk what `conda-build` does but I don't think we should be creating empty directories in the $PREFIX. And I also think I led you down a wrong path as this is IMO user error - I should have created this folder as part of the recipe by using `mkdir %LIBRARY_BIN%` or whatever is needed on Windows.

zelosleone:
> OK, idk what `conda-build` does but I don't think we should be creating empty directories in the $PREFIX. And I also think I led you down a wrong path as this is IMO user error - I should have created this folder as part of the recipe by using `mkdir %LIBRARY_BIN%` or whatever is needed on Windows.

Oh thats fine then, I was just setting up the conda build env with python libs too, we should close the PR then?

zelosleone:
BTW conda build passes the recipe in this pr, so i think they create empty folders from env path values too

wolfv:
Yeah we can close the PR and the issue. Sorry again. I would still be curious about what conda-build actually does :)

wolfv:
ie. what directories it creates

zelosleone:
> ie. what directories it creates

I will check it out

zelosleone:
> ie. what directories it creates

they create whatever empty or full:

L¦¦¦Library
    +¦¦¦bin
    +¦¦¦include
    L¦¦¦lib

PR #1586, opened 2025-04-24
feat: add R interpreter support and documentations

zelosleone:
As we talked in discord, I didn't add the auto script detection yet. I will open a new issue for dropping the sh/bat files in favor of script files for the languages themselves. This is also my first serious documentation writing, so I am definitely open to feedback here.


docs/tutorials/r.md
```diff
@@ -2,8 +2,79 @@
...
+    content: |
+      # Note, to install the package via source, we need to set the SRC_DIR environment variable
+      install.packages(Sys.getenv("SRC_DIR"), repos=NULL, type="source")
+
+requirements:
+  build:
+    - ${{ compiler('c') }}                # [not win]
+    - ${{ compiler('cxx') }}              # [not win]
+    - mingwpy                             # [win]
+    - ucrt                                # [win]
```
wolfv, line 45: This syntax `# [win]` does not work anymore with v1 recipes.

Instead, you need to use

```
if: win
then:
  - m2-filesystem
  - ...
```

docs/tutorials/r.md
```diff
@@ -2,8 +2,79 @@
...
+      interpreter: r
+      content: |
+        library('SpecsVerification')
+        TRUE
+
+about:
+  homepage: https://CRAN.R-project.org/package=SpecsVerification
+  license: GPL-2.0-or-later
+  summary: A collection of forecast verification routines developed for the SPECS FP7 project. The emphasis is on comparative verification of ensemble forecasts of weather and climate.
+  license_family: GPL2
```
wolfv, line 71: license_family is not available in `v1` recipes anymore

docs/tutorials/r.md
```diff
@@ -2,8 +2,79 @@
...
+source:
+  url:
+    - https://cran.r-project.org/src/contrib/SpecsVerification_0.5-3.tar.gz
+    - https://cran.r-project.org/src/contrib/Archive/SpecsVerification/SpecsVerification_0.5-3.tar.gz
+  sha256: 630fd876b51cb5e22061fa64dbb447c09e88c14e81fb801001ae18e969a4e6ec
+
+build:
+  number: 0
+  script:
+    interpreter: r
```
wolfv, line 35: lets keep using the regular "bash" invocation for this (although this is pretty cool!)

docs/tutorials/r.md
```diff
@@ -0,0 +1,120 @@
...
+
+Test scripts should load the package and return `TRUE` for success:
+
+```yaml
+library(mypackage)
+# Verify functionality if needed
+TRUE
+```
+
+Although there will be packages that you will need to compile from source code, our suggestion is to always find and prefer pre-built conda packages over compiling from source when possible. Regardless, rattler-build will still function faster than it's alternatives even while compiling from source code!
```
wolfv, line 120: this sentence doesn't really make sense since rattler-build is used to build packages from source :)

test-data/recipes/r-test/recipe.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+requirements:
+  host:
+    - r-base
+    - r-ggplot2
+  run:
+    - r-base
+    - r-ggplot2
+
+tests:
```
wolfv, line 40: You are not testing the new R test you added at all?

wolfv: @zelosleone make sure to also test the new R test you added.

src/package_test/run_test.rs
```diff
@@ -916,7 +916,7 @@ impl RTest {
         prefix: &Path,
         config: &TestConfiguration,
     ) -> Result<(), TestError> {
-        let span = tracing::info_span!("Running R test");
+        let span = tracing::info_span!("Running script test for", recipe = pkg.to_string());
```
wolfv, line 919: Why `script`?

wolfv:
Maybe we should call the script interpreter `Rscript` or `rscript` instead of `R` to make it more clear?

src/package_test/run_test.rs
zelosleone: I mean essentially, we are running a Rscript, i think that makes a better sense/more informative about what we are doing

src/recipe/parser/test.rs
```diff
@@ -265,10 +277,14 @@ impl TryConvertNode<TestType> for RenderedMappingNode {
                     let perl = as_mapping(value, key_str)?.try_convert(key_str)?;
                     test = TestType::Perl { perl };
                 }
+                "rscript" => {
```
wolfv, line 280: This test should be `r`

src/recipe/parser/test.rs
```diff
@@ -265,10 +277,14 @@ impl TryConvertNode<TestType> for RenderedMappingNode {
...
+                    let rscript = as_mapping(value, key_str)?.try_convert(key_str)?;
+                    test = TestType::R { r: rscript };
+                }
                 invalid => Err(vec![_partialerror!(
                     *key.span(),
                     ErrorKind::InvalidField(invalid.to_string().into()),
-                    help = format!("expected fields for {name} is one of `python`, `perl`, `script`, `downstream`, `package_contents`")
+                    help = format!("expected fields for {name} is one of `python`, `perl`, `rscript`, `script`, `downstream`, `package_contents`")
```
wolfv, line 287: This should be `r` (not rscript). `rscript` is the interpreter, but for the test name R is fine.

wolfv:
Last comment and then we can merge!

zelosleone:
> Last comment and then we can merge!

Finished the last comment as well!

PR #1589, opened 2025-04-28
feat: Implement retry logic for directory deletion on Windows

### LINKED ISSUE #1431: Windows os errors during build/test cleanup
As always, thanks for `rattler-build`!

I have now seen issues during both the `build` and `test` phases where windows is unable to proceed due to (presumably) file permissions and/or dangling processes:

On https://github.com/conda-forge/granian-feedstock/pull/42:

```
 ╰─────────────────── (took 3 minutes)
  all tests passed!
Error: 
  × failed to remove directory `D:\bld\bld\rattler-build_granian-with-
  │ all_1739238925\work\test`: Access is denied. (os error 5)
```

I tried using some python tricks (`psutil` and its lovely child tracking), but it was no good. I just skipped the test on windows and moved on.

On https://github.com/conda-forge/plantuml-feedstock/pull/62:

``` 
 │ │ Archive written to 'D:\bld\win-64\plantuml-1.2025.1-he596248_0.conda'
 │ │
 │ ╰─────────────────── (took 4 seconds)
 │
 ╰─────────────────── (took 9 minutes)
Error: 
  × failed to remove directory `D:\bld\bld\rattler-
  │ build_plantuml_1739725916\build_env`: The process cannot access the file
  │ because it is being used by another process. (os error 32)
```

This is of course harder to work around, as I can't just _skip_ the build.

While leaking disk/memory is of course A Bad Thing, I'd really rather be able to opt into getting a warning... which I could duly ignore on windows CI until it _really failed_ due to resource exhaustion.

zelosleone:
closes #1431

It was extremely hard to reproduce this bug constantly, I could reproduce it 5 times, sixth time it would just vanish. The solution was obviously retrying for a few times after encountering the either OS error 5 or 32, so i just added unit tests that simulates this with a locked file.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -127,7 +127,7 @@ fn try_remove_with_retry(path: &Path, first_err: Option<std::io::Error>) -> std:
 
     while attempts < max_retries {
         if let Some(e) = &last_err {
-            tracing::debug!("Retrying deletion {}/{}: {}", attempts + 1, max_retries, e);
+            tracing::info!("Retrying deletion {}/{}: {}", attempts + 1, max_retries, e);
             std::thread::sleep(
                 std::time::Duration::from_millis(500 * (1 << attempts.saturating_sub(1)))
                     .min(std::time::Duration::from_secs(3)),
@@ -182,6 +182,7 @@ mod tests {
         use super::*;
         use std::fs::File;
         use std::fs::OpenOptions;
+        use std::io::Write;
         use std::os::windows::fs::OpenOptionsExt;
         use std::sync::{Arc, Mutex};
         use std::time::Duration;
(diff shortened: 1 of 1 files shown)


wolfv:
I think, in general, this makes sense and is also something that `conda-build` implements. Great! 

However, what I am missing is a test that creates a `read-only` file and then attempts to remove that (which is what is going wrong in the Go case). The Go compiler creates a cache directory that is marked as read-only.

src/utils.rs
```diff
@@ -106,7 +106,47 @@ pub fn remove_dir_all_force(path: &Path) -> std::io::Result<()> {
...
+#[cfg(windows)]
+/// Retries clean up when encountered with OS 32 and OS 5 errors on Windows
+fn try_remove_with_retry(path: &Path, first_err: Option<std::io::Error>) -> std::io::Result<()> {
+    let max_retries = 5;
+    let mut attempts: i32 = if first_err.is_some() { 1 } else { 0 };
+    let mut last_err = first_err;
+
+    while attempts < max_retries {
+        if let Some(e) = &last_err {
+            tracing::debug!("Retrying deletion {}/{}: {}", attempts + 1, max_retries, e);
```
wolfv, line 130: This can be `info` I think so that the user knows what's happening and why things are slow.

PR #1606, opened 2025-05-02
feat: add cmake/pkg-config tests

wolfv:
Ideating some additional checks:

- test if cmake can execute `find_package` successfully
- test if pkg-config finds the package successfully

The `cmake` test unfortunately _requires_ compilers to be installed as CMake is going to search for them. 

For `pkg-config` I am not sure what to call the second key.

Tests look like:

```
  - cmake:
      find_package: [xtensor]
  - pkg_config:
      pkg_config: [xtensor]
```

### DIFF
--- a/src/package_test/run_test.rs
+++ b/src/package_test/run_test.rs
@@ -33,8 +33,8 @@ use crate::{
     env_vars,
     metadata::{Debug, PlatformWithVirtualPackages},
     recipe::parser::{
-        CommandsTest, DownstreamTest, PerlTest, PythonTest, PythonVersion, RTest, Script,
-        ScriptContent, TestType,
+        CMakeTest, CommandsTest, DownstreamTest, PerlTest, PythonTest, PythonVersion, RTest,
+        Script, ScriptContent, TestType, PkgConfigTest,
     },
     render::solver::create_environment,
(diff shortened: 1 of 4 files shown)


src/package_test/run_test.rs
```diff
@@ -975,3 +985,158 @@ impl RTest {
...
+            config.channel_priority,
+            config.solve_strategy,
+        )
+        .await
+        .map_err(TestError::TestEnvironmentSetup)?;
+
+        let tmp_dir = tempfile::tempdir()?;
+        let cmake_file = tmp_dir.path().join("CMakeLists.txt");
+
+        let mut cmake_content = String::from("cmake_minimum_required(VERSION 3.15)\nproject(cmake_test)\n\n");
```
baszalmstra, line 1027: You can also maybe use: https://cmake.org/cmake/help/latest/manual/cmake.1.html#run-the-find-package-tool

trim21:
> The cmake test unfortunately requires compilers to be installed as CMake is going to search for them.
 https://stackoverflow.com/a/46177192/8062017 does this work？

trim21:
also there is a existing cli tool for cmake https://anaconda.org/conda-forge/cmake-package-check

PR #1607, opened 2025-05-02
feat: Add support for loading files relative to recipe path

### LINKED ISSUE #1540: should `load_from_file` file path relative to recipe.yaml instead of cwd?
We currently load file with cwd as base path, maybe it's better to be relative to recipe.yaml instead of cwd?

This will make rattler-build free to be executed from any path.

similar problem on pixi: https://github.com/prefix-dev/pixi/issues/2615

zelosleone:
closes #1540

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -195,6 +195,10 @@ criterion = { version = "0.5.1", features = ["html_reports"] }
 name = "line_breaks"
 harness = false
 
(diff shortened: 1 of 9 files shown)


trim21:
I accidentally unsubscribed from the original issue and just find the PR now, thank you for the contribution

PR #1613, opened 2025-05-02
fix: Enhance retry (increase retries) and increase sleep time in test

zelosleone:
i don't know why it only happens with some runners, its impossible to reproduce it in local and in at least half of the runners


wolfv:
Let's close this one for now and see if we still hit this issue in the CI. 

There are two things that improved:

- not using fs_err anymore means we get the correct error code from `raw_os_error` now and thus the retrying works
- the tests are run in different directories now which means that the directories are not as often re-used as before

PR #1617, opened 2025-05-04
fix: add delay for Windows in test execution

### LINKED ISSUE #1628: Windows test errors with os error 1224
Image

This is on vs2022 build tools (i assume 2019 and 2017 are fine due to github runners using them)

zelosleone:
currently after we make conda recognize rattler-build environments, we are dealing with reindexing file lock issues in windows where 13 tests fail due to file permission issues, quite esoteric ones as well with os error 1224. This adds a small delay before reindexing operations, which will solve all of these issues on windows. I also thought and tried to make the reindexing separate for tests but that would cause race condition issues, so i think this is the best solution

p.s: https://github.com/prefix-dev/rattler-build/pull/1613 we also need this for the retry logic on file removal

closes #1628

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -552,7 +552,7 @@ pub async fn run_build_from_args(
 
             // let testable = can_test(&test_queue, &all_output_names, &outputs_to_build);
             for (output, archive) in &to_test {
-            // we need a delay here to stop reindexing process undermine the filesystem lock in windows
+                // we need a delay here to stop reindexing process undermine the filesystem lock in windows
                 if cfg!(windows) {
                     tokio::time::sleep(std::time::Duration::from_millis(200)).await;
                 }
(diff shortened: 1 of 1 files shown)


zelosleone:
@wolfv a lot of tests will fail 99% of the times on windows local if we dont use this :(

wolfv (approved):

wolfv:
I am fine with this PR but I don't even see that many PRs failing?

zelosleone:
> I am fine with this PR but I don't even see that many PRs failing?

sorry, thats my fault for not including all the test results! here are they:

image

all same errors with os error

src/lib.rs
```diff
@@ -552,6 +552,16 @@ pub async fn run_build_from_args(
 
             // let testable = can_test(&test_queue, &all_output_names, &outputs_to_build);
             for (output, archive) in &to_test {
+                if cfg!(windows) {
+                    tokio::time::sleep(std::time::Duration::from_millis(200)).await;
```
wolfv, line 556: We need a comment here why this was introduced.

wolfv:
I think this was fixed for same reasons as mentioned in #1613

PR #1618, opened 2025-05-04
feat: add interpreter detection for various script types

### LINKED ISSUE #1611: Auto-discover interpreter based on script filename
We can set the interpreter manually like:

```
script:
  interpreter: python
  file: foobar.py
```

However, it would be nice if we can auto-discover the interpreter based on the filename, so that we can write:

```
script: foobar.py
```

And it will automatically choose the `python` interpreter. Same for Perl, R, Bash, cmd.exe, ...

zelosleone:
closes #1611

i was debating whetever i should add a UTF-8 validation but i thought it should be added just in case for windows computers with asian languages, let me know what you think about it @wolfv 

Also, as shown in recipe examples we still hold the requirements/run to not to cause a magicial workflow, because i thought people want to control exactly what they want to run with


src/recipe/parser/script.rs
```diff
@@ -202,6 +203,42 @@ impl From<ScriptContent> for Script {
...
+    Ok(())
+}
+
+/// Helper function to determine interpreter based on file extension
+fn determine_interpreter_from_path(path: &Path) -> Option<String> {
+    path.extension()
+        .and_then(|s| s.to_str())
+        .map(|ext| ext.to_lowercase())
+        .and_then(|ext_lower| match ext_lower.as_str() {
+            "py" | "pyw" | "pyc" => Some("python".to_string()),
```
wolfv, line 231: We don't need `pyw` and `pyc`, they should not ever occur for a Python script (`pyc` is compiled python).

zelosleone: Oh okay, can you tell me whats wrong with bash though? Having a little bit trouble there

zelosleone: I might have forgotten to add requirements but it shouldn't matter, on WSL it works correctly?

src/recipe/parser/script.rs
```diff
@@ -202,6 +203,42 @@ impl From<ScriptContent> for Script {
...
+            "pl" => Some("perl".to_string()),
+            "r" => Some("rscript".to_string()),
+            "sh" | "bash" | "zsh" => Some("bash".to_string()),
```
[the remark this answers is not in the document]
zelosleone: Okay got my answer

src/script/mod.rs
```diff
@@ -268,9 +269,10 @@ impl Script {
         let interpreter_is_nushell = interpreter == "nushell" || interpreter == "nu";
 
         // Determine the valid script extensions based on the available interpreters.
-        let mut valid_script_extensions = Vec::new();
+        let mut valid_script_extensions = vec!["py", "pl", "r"];
```
wolfv, line 272: Since we are planning to remove the "build.sh" detection in the future, I think we should not add any additional file extensions here.

pixi.toml
```diff
@@ -40,6 +40,7 @@ requests = ">=2.32.2,<2.33"
 syrupy = "4.6.*"
 rust-src = ">=1.86.0,<1.87"
 boto3 = "*"
+nushell = "==0.104.0"
```
wolfv, line 43: I think this can be reverted since you are also requesting `nushell` in the build requirements of the recipe.

wolfv (approved):

pixi.toml
zelosleone: Nope, github runners for some reason doesnt see it without adding specially to lock file. This could be a bug btw

test-data/recipes/interpreter-detection/nu/recipe.yaml
```diff
@@ -0,0 +1,16 @@
...
+  script: build.nu
+
+requirements:
+  build:
+    - nushell
+
+tests:
+  - script: build.nu
+    requirements:
+      run:
```
wolfv, line 15: Wrt to removing `nushell` from pixi, would it work if you use `build` here, instead of run?

zelosleone:
Don't merge pls, i will fix the git test section, i messed up the conflict solving with jetbrains merge tool

zelosleone:
fixed! @wolfv

wolfv (approved):

PR #1621, opened 2025-05-05
feat: add Unicode normalization for path handling and detect case-insensitive filename collisions

### LINKED ISSUE #1262: Add a warning for packages with mixed case
If a package has mix-cased filenames (e.g. `foo/BAR` and `foo/bar`) then the package does not work fine on case-insensitive filesystems (like the standard APFS on Apple).

We can detect this relatively easily at build time and issue a warning.

zelosleone:
closes #1262 hopefully, i am only confident with unit tests but hopefully it passes mac os!

### DIFF
--- a/src/packaging.rs
+++ b/src/packaging.rs
@@ -1,9 +1,9 @@
 //! This module contains the functions to package a conda package from a given
 //! output.
 use std::{
-    collections::HashSet,
+    collections::{HashMap, HashSet},
     io::Write,
-    path::{Component, Path, PathBuf},
+    path::{Component, MAIN_SEPARATOR, Path, PathBuf},
 };
 
 use fs_err as fs;
@@ -15,6 +15,7 @@ use rattler_conda_types::{
 use rattler_package_streaming::write::{
     CompressionLevel, write_conda_package, write_tar_bz2_package,
 };
+use unicode_normalization::UnicodeNormalization;
 
 mod file_finder;
 mod file_mapper;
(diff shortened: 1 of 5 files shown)


zelosleone:
@wolfv forgot to copy the case sensitive files to PREFIX folder :DD thankfully it passes now

zelosleone:
Alright, ubuntu passes, macos doesn't, i will need to check that out too

zelosleone:
@wolfv We cannot create the test files at all on mac os due to case insensivity, same as windows :/ even if i try to make a package it would fail sadly. but i think unit tests + e2e test working on linux could be enough?

wolfv:
Sounds good indeed!

src/packaging.rs
```diff
@@ -421,3 +570,45 @@ impl Output {
...
+        let files = vec![Path::new("foo/bar"), Path::new("foo/baz")];
+        let groups = find_case_insensitive_collisions(&files);
+        assert!(groups.is_empty());
+    }
+
+    #[test]
+    fn test_find_case_insensitive_collisions_unicode() {
+        let files = vec![
+            Path::new("foo/straße"),
+            Path::new("foo/STRASSE"),
```
wolfv, line 608: I don't think these two (straße and STRASSE) would be the same on case-insensitive file systems.. or would they?)

wolfv: Wow, they are! Just checked locally on macOS and this is true :)

Issue #1608, opened 2025-05-02
Certain upstream hosts fail on download with OSX (osx-64 conda-forge issue)


bollwyvl:
Noted on [conda-forge- zulip]([#general > Error only on OSX: Error downloading ftp.gnu.org/...](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/Error.20only.20on.20OSX.3A.20Error.20downloading.20ftp.2Egnu.2Eorg.2F.2E.2E.2E))
Some hosts are returning interesting HTTP errors during CI builds but only on OSX:

> https://github.com/conda-forge/less-feedstock/blob/26405d18edd59d71e08955e3dc64aff7cae58fe1/recipe/recipe.yaml

```
 ╭─ Running build for recipe: less-668-h8eb8b2c_1
 │
 │ ╭─ Fetching source code
 │ │ Fetching source from url: https://ftp.gnu.org/gnu/less/less-668.tar.gz
 │ │
 │ ╰─────────────────── (took 0 seconds)
 │
 ╰─────────────────── (took 0 seconds)
Error:   × Error downloading https://ftp.gnu.org/gnu/less/less-668.tar.gz: invalid
  │ HTTP version parsed


##[error]Bash exited with code '1'.
```

> https://github.com/conda-forge/fossil-feedstock/blob/7367da3c6e41de3ba5dff2a11a72808da9aebd0b/recipe/recipe.yaml#L15

Logs gone, got slightly further with a `400` response, and no further information.

I would say it's IP-based rate limiting in this brave new world of garbage scrapers, but it's not occuring on _other_ azure jobs. Perhaps it's some agent string?

bollwyvl:
From the [zulip thread](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/Error.20only.20on.20OSX.3A.20Error.20downloading.20ftp.2Egnu.2Eorg.2F.2E.2E.2E/near/516002829), with access to both sides of the pipe:

> https://github.com/conda-forge/slepc-feedstock/pull/98#issuecomment-2849036055
>
> By looking at the server log files, the requests that fail appear as
>
> `"GET /download/distrib/slepc-3.23.1.tar.gz" 400 3322 "-" "-"`
> 
> while the successful requests appear as
>
> `"GET /download/distrib/slepc-3.23.1.tar.gz HTTP/1.1" 404 3691 "-" "Wget/1.25.0"`
>
[...]

PR #1623, opened 2025-05-06
[WIP] try fixing missing http version

minrk:
for #1608

I can reproduce the problem with the conda-forge builds, but can't when I build it myself, so I can't test this.

This currently tries setting http1_only on the client instead of http version in the request.

### DIFF
--- a/src/source/url_source.rs
+++ b/src/source/url_source.rs
@@ -72,7 +72,6 @@ async fn fetch_remote(
     let (mut response, download_size) = {
         let resp = client
             .get(url.as_str())
-            .version(reqwest::Version::HTTP_11)
             .send()
             .await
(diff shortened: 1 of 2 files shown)


wolfv:
We can try this change in https://github.com/conda-forge/rattler-build-feedstock/pull/153

pavelzw:
> We can try this change in https://github.com/conda-forge/rattler-build-feedstock/pull/153

doesn't work :( https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1240450&view=logs&j=e0208569-136d-54da-4ec5-14d0e8771cbc&t=25992e30-3e2a-5788-2b3e-e373b0534bf7

PR #1626, opened 2025-05-06
add `--continue-on-failure`

wolfv:


zelosleone (approved):

zelosleone:
I only have read on permissions so cannot push but i think solution for last piece is:

`src/lib.rs:`

`run_build_from_args(outputs, tool_config).await?;`

>>
```
 // If continue_on_failure is enabled, catch and report errors but don't fail the entire build
    match run_build_from_args(outputs, tool_config).await {
        Ok(_) => Ok(()),
        Err(e) => {
            if build_data.continue_on_failure == ContinueOnFailure::Yes {
                tracing::warn!(
                    "Some packages failed to build, but continuing due to --continue-on-failure flag"
                );
                Ok(())
            } else {
                Err(e)
            }
        }
    }
```
Basically, we just need to stop it from failing at the end of the build and log a warning.

src/lib.rs
```diff
@@ -953,7 +953,20 @@ pub async fn build_recipes(
     outputs = skip_noarch(outputs, &tool_config).await?;
 
     sort_build_outputs_topologically(&mut outputs, build_data.up_to.as_deref())?;
-    run_build_from_args(outputs, tool_config).await?;
+        // If continue_on_failure is enabled, catch and report errors but don't fail the entire build
+    match run_build_from_args(outputs, tool_config).await {
+        Ok(_) => Ok(()),
+        Err(e) => {
+            if build_data.continue_on_failure == ContinueOnFailure::Yes {
+                tracing::warn!(
```
wolfv: Is this really right? There is no loop here ...

zelosleone: Yeah after seeing the compile errors on test i changed the logic, can you check it again?

wolfv:
Thanks @kerimbirgi! Sorry for letting your PR linger for so long!

PR #1627, opened 2025-05-06
add a test to analyse the CRLF patch issues

wolfv:


test-data/patch_application/workdir/text.md
```diff
@@ -0,0 +1,5 @@
+I am a long text.
+And I want to be patched.
+Oh, wow, I was patched! Thank you soooo much!
```
zelosleone, line 3: If we delete this line (which i think we should, because patch will try to see if the file is already patched) it will correctly apply the patch

src/source/patch.rs
```diff
@@ -109,4 +113,39 @@ mod tests {
...
+        let clrf_patch = LineEnding::CRLF.apply(&text);
+
+        fs_err::write(tempdir.path().join("patches/test_clrf.patch"), clrf_patch).unwrap();
+
+        apply_patches(
+            &SystemTools::new(),
+            &[PathBuf::from("test_clrf.patch")],
+            tempdir.path().join("workdir").as_path(),
+            tempdir.path().join("patches").as_path(),
+        )
```
zelosleone, line 148: We are applying the patch a second time here, but not in a clean `workdir` folder, thats along with line in markdown failure, the reason it fails

wolfv: sorry! great point1

PR #1647, opened 2025-05-11
fix: make `relink` on linux recognize `${ORIGIN}`

### LINKED ISSUE #1646: RPATH started with `${ORIGIN}` is not recognized by `relink`
As shown in https://github.com/prefix-dev/rattler-build/blob/2621fc17c5c9a93afb24865cdcf54f43c8c15842/src/linux/link.rs#L172, the `relink` function will detect the RPATHs which start with `$ORIGIN`. But as illustrated in https://man7.org/linux/man-pages/man8/ld.so.8.html, the `${ORIGIN}` is equivalently to `$ORIGIN`, and `${ORIGIN}` will not be detected by rattler-build so far.

This will make some inconvenience when packing some projects which prefer `${ORIGIN}`. For example, https://github.com/prophesee-ai/openeb, it use `${ORIGIN}` in [CMakeLists.txt](https://github.com/prophesee-ai/openeb/blob/0499c95d921755ab2085125a29750ef365e6e8ae/hal_psee_plugins/lib/CMakeLists.txt#L230-L231) . So in order to build projects like openeb, I need to do some dirty work to ensure the correctness of RPATH [build.sh](https://github.com/wzf03/conda-recipes/blob/825ea5fdfc0a8e1f66884672ac2030e6bae26bff/openeb/build.sh#L24).

I wonder if it would be OK to make `${ORIGIN}` to be recognized by the `relink`, which may make rattler-build's behavior different from original conda-build's behavior. Thanks for your excellent work!

wzf03:
Fix #1646

### DIFF
--- a/src/linux/link.rs
+++ b/src/linux/link.rs
@@ -118,7 +118,10 @@ impl Relinker for SharedObject {
                 .strip_prefix(prefix)
                 .expect("library not in prefix"),
         );
-        if let Ok(rpath_without_loader) = rpath.strip_prefix("$ORIGIN") {
+        if let Ok(rpath_without_loader) = rpath
+            .strip_prefix("$ORIGIN")
+            .or_else(|_| rpath.strip_prefix("${ORIGIN}"))
+        {
             if let Some(library_parent) = self_path.parent() {
                 return to_lexical_absolute(rpath_without_loader, library_parent);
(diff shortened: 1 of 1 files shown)


wolfv:
Thanks!

PR #1648, opened 2025-05-11
feat: auto detect license files from prefix directory as well

wolfv:
This adds auto-detection from the `$PREFIX` as well.

@mfansler do you think this is acceptable? It would fix the Rust case. I would have to think a bit longer to find a good solution for `${{ PREFIX }}/foobar` case because the way we render the recipe, these variables are not available.

### DIFF
--- a/src/packaging.rs
+++ b/src/packaging.rs
@@ -98,65 +98,90 @@ fn copy_license_files(
     output: &Output,
     tmp_dir_path: &Path,
 ) -> Result<Option<HashSet<PathBuf>>, PackagingError> {
+    // Early return if no license files are specified
     if output.recipe.about().license_file.is_empty() {
-        Ok(None)
-    } else {
-        let licenses_folder = tmp_dir_path.join("info/licenses/");
-        fs::create_dir_all(&licenses_folder)?;
+        return Ok(None);
+    }
 
-        let copy_dir = copy_dir::CopyDir::new(
+    // Create licenses directory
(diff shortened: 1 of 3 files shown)


mfansler:
@wolfv thanks for the heads up. Yes, this would suffice. 👍 

So, our translation will be, for example:

**v0**
```yaml
about:
  # other stuff
  license_file:
    - {{ environ["PREFIX"] }}/lib/R/share/licenses/MIT
    - LICENSE
```

**v1**
```yaml
about:
  # other stuff
  license_file:
    - lib/R/share/licenses/MIT
    - LICENSE
```

I.e., simply remove `{{ environ["PREFIX"] }}/` from any `about.license_file` entries.

wolfv:
Yes, with this PR that would be the case, indeed.

wolfv:
I've had some extra thoughts on this. I think it might be better to compute the `directories` first (without creating them) so that we can already resolve `${{ PREFIX }}`, ${{ SRC_DIR }}` and `${{ RECIPE_DIR }}` when rendering, so that users can be really explicit about where the license is supposed to come from.

For this I would also modify the folder that we create (currently it looks like `package-name_1234513425134`) where the number is the timestamp. The number is not easy to understand anyways (hard to know which is the latest build), so if we have e.g. a running index it might be better regardless.

wolfv:
We implemented "late bound paths" so that you can use `${{ PREFIX }}` in the license files section.

Issue #1635, opened 2025-05-09
lib existence tests w/ mingw on windows


theAeon:
Hi all-currently trying to get a mingw variant of a package working and noticed that on windows, rattler build searches for `Library/lib/$package.lib` but does *not* search for `Library/lib/lib$package.dll.a`, which is the standard import lib format for mingw. Should that be checked for along with the corresponding `lib$package.dll`?

theAeon:
Hold on-its possible I'm stupid and this is because i forgot to pass the install prefix. Will reopen if true.

theAeon:
Okay, now that I've verified its actually *installing* the library, I can safely say that this does not work.

wolfv:
Is the .a a static library?

theAeon:
dll.a is how mingw indicates an import library.

wolfv:
Would you be able to make a PR to add this extension on Windows?

theAeon:
Probably. I'm not exactly experienced in rust (and by that I mean I've never actually used it) but it shouldn't be a particularly difficult tweak.

PR #1649, opened 2025-05-12
add mingw globs for libtest (#1635)

theAeon:
Definitely gonna need some help with the tests here but this *should* work if the code is lazy matching like i think it is.

Obviously the ideal hee would be to do paired matching (either lib$X.dll.a + lib$X.dll or $x.lib + $x.dll) but i don't think the current logic is conducive to that?


theAeon:
I wonder if its easy to select which set is added based on the compiler in use?

Issue #1507, opened 2025-03-21
Support erroring (possibly by default) when detecting the host prefix in the binary


vyasr:
Currently rattler-build supports building conda packages that will have the host prefix replaced in built binaries at install time by conda using similar logic to conda-build as of #598. However, there are flaws with this approach. As discussed extensively in https://github.com/rapidsai/cudf/issues/18251#issue-2914747519 and https://github.com/rapidsai/cudf/issues/18251#issuecomment-2725967075 (those two comments give a complete picture of exactly how this failed), a sufficiently smart compiler can always produce a binary for which binary prefix replacement is unsafe, regardless of suitable null character padding or other safety checks. In https://github.com/conda/conda-build/issues/1674#issuecomment-2744643485, I proposed that conda-build offer a new option to error on package build if such a prefix is detected, since in practice the best solution to these kinds of problems is to fix the package source or build process to avoid embedding the host prefix in the first place. With rattler-build, as a newer packaging tool we have more flexibility and I would suggest that we go further.

I think the default behavior of rattler-build should be to error if the prefix is in the binary. This can be implemented fairly trivially using an extra flag in the config and by checking the value of `contains_prefix_binary` [here](https://github.com/prefix-dev/rattler-build/blob/main/src/packaging/metadata.rs#L181). My reasoning for this is that it is easy enough for a project to opt in to prefix patching if there is absolutely no way they can avoid the need, but at least if it is opt in they must know what they are choosing and accept the potential pitfalls and maybe know where to look if something goes wrong. Conversely, if patching is automatic the error modes are potentially arbitrarily bad (silent failures, data corruption, program crashes) and extremely difficult to debug. Realistically most developers would struggle to fully debug a problem like https://github.com/rapidsai/cudf/issues/18251 and trace it back to conda's binary prefix replacement as the source. If at all possible, we should encourage packages to fix their source code and build process to avoid needing binary prefix patching rather than having them implicitly rely on this behavior without realizing it.

**Text prefix patching**
As an aside, I also have a weak preference for the same behavior in text prefix patching, but in that case I think the problems are far less difficult to debug and also the process is far less likely to go wrong in unpredictable ways, so I am less concerned with the behavior there.

wolfv:
Thanks for the write up. I agree that it might be a good change. Would be interesting to scan the existing packages to see how many do rely on binary prefix replacement (maybe even looking in a few larger environments could be useful, to get an idea). 

[...]

PR #1658, opened 2025-05-14
feat: implement binary prefix detection behavior in packaging

zelosleone:
closes https://github.com/prefix-dev/rattler-build/issues/1507 https://github.com/prefix-dev/rattler-build/issues/1681

I’ve tried to change the usage of bool types to Option Type as well, just because it looked cleaner. I think it is a little bit better for our usage case as well since we will now default to error. I also used proper types instead of bool types for our usage case, just in case we will introduce more types in the future. On windows, binaries aren’t a problem, so I limited the test to UNIX.

### DIFF
--- a/src/packaging/file_mapper.rs
+++ b/src/packaging/file_mapper.rs
@@ -234,7 +234,9 @@ impl Output {
                             }
                         }
                     } else if target.is_absolute() {
-                        tracing::warn!("Symlink {:?} points outside of the prefix", path);
+                        tracing::warn!(
+                            "Symlink {path:?} points to absolute path {target:?} outside of the $PREFIX",
+                        );
                     }
                     // Create the symlink at dest_path
                     #[cfg(unix)]


src/packaging/metadata.rs
```diff
@@ -162,14 +166,6 @@ pub fn create_prefix_placeholder(
     };
 
     if file_mode == FileMode::Binary {
-        if prefix_detection.ignore_binary_files {
```
wolfv, line 165: I think it's still better to short cut this because we can save the expensive search (and rattler-build will be faster!).

zelosleone:
> The idea is good but we cannot easily change the schema at this point, unfortunately. It would destroy backward compatibility, sadly. For now, IMO the only thing we can do is add a CLI flag, and later create a v2 schema where we allow the new value(s).
> 
> Or we could add a third value (true/false/error) but even that is a "breaking" schema change.

Gotcha! I will revert the changes on option type then

test-data/recipes/prefix_detection/recipe.yaml
```diff
@@ -19,6 +19,7 @@ build:
         - force_binary/*
       text:
         - force_text/*
+    ignore_binary_files: true
```
wolfv, line 22: Why is this existing test changed?

wolfv:
I think there is some misunderstanding going on.

What we want to introduce is a sanity check to error the package build if any files contain a binary prefix. This should be implemented as an additional CLI flag.

Actually, I think we can just do these checks in a post-processing stage and then let the package fail if the flag is activated.
We can use the same method to check (on Windows) if the final package contains any symlinks (#1681).

test-data/recipes/prefix_detection/recipe.yaml
zelosleone: I wanted to show the existin tests also passes with our changes as well

wolfv: But they are _not_ testing (you had to change the test outcomes in `test_simple.py`.

zelosleone:
> I think there is some misunderstanding going on.
> 
> What we want to introduce is a sanity check to error the package build if any files contain a binary prefix. This should be implemented as an additional CLI flag.
> 
> Actually, I think we can just do these checks in a post-processing stage and then let the package fail if the flag is activated.
> We can use the same method to check (on Windows) if the final package contains any symlinks (#1681).

I am switching to windows ATM and can only reply on mobile but yeah I might have misunderstood the issue, I will try to include the fix for the symlink issue as well today or tomorrow (I will let you know about it too) and will let you know about the progress

test/end-to-end/test_simple.py
```diff
@@ -605,7 +700,7 @@ def check_path(p, t):
         elif path == "force_text/file_without_prefix":
             check_path(p, None)
         elif path == "force_binary/file_with_prefix":
-            check_path(p, "binary")
```
wolfv, line 608: why was this changed?

test/end-to-end/test_simple.py
```diff
@@ -579,18 +579,113 @@ def test_prefix_detection(rattler_build: RattlerBuild, recipes: Path, tmp_path:
...
+    pkg = get_extracted_package(tmp_path, "prefix_detection_binary_error")
+
+    assert (pkg / "info/index.json").exists()
+    assert (pkg / "info/paths.json").exists()
+
+    index_json = json.loads((pkg / "info/index.json").read_text())
+    subdir = index_json["subdir"]
+    is_win = subdir.startswith("win")
+
+    def check_path(p, t):
```
wolfv, line 667: We should not copy this function. Just put it in the global scope so that we can reuse it.

zelosleone:
> Looks good and clean :)

But i gotta fix the windows issue now, I will work on this today xD

src/packaging/file_mapper.rs
```diff
@@ -220,63 +220,55 @@ impl Output {
...
+                                dest_path,
+                                target,
+                                e
+                            );
+                        }
+                    }
+                    #[cfg(windows)]
+                    {
+                        let res = if metadata.file_type().is_dir() {
+                            std::os::windows::fs::symlink_dir(&target, &dest_path)
```
wolfv, line 252: Can we use `fs_err` here?

wolfv (approved):

PR #1659, opened 2025-05-14
docs: now we do not need tar on host system

trim21:


wolfv:
Yes, that is correct!

wolfv (approved):

trim21:
ci failed on cargo-fmt which I think is not related to this PR

PR #1664, opened 2025-05-15
feat: Add `purls` field to about section

delsner:


JeanChristopheMorinPerso:
Should it also depend on the CEP at https://github.com/conda/ceps/pull/114?

Issue #915, opened 2024-06-07
Package content tests - add `exists / not_exists`


wolfv:
Sometimes you want to check that a file is there and sometimes the inverse.

We had ideas to use `~ foo.txt` to say 'not exists' but people were against it. 

Instead, let's use a dictionary like so:

```yaml
package_content:
  include:
    exists: 
      - foo.hpp
    not_exists:
      - baz.hpp
```

wolfv:
Note: this reflects `includes` and `excludes` from the regular "glob" places.

minrk:
Just ran into this, and it would indeed be nice. The most common exclusion check I think I have in recipes is making sure static libs aren't being installed, or e.g. _any_ include files or _any_ libs when e.g. headers and things are split into multiple outputs.

beckermr:
We need to not make arbitray changes to the yaml. Otherwise other tools won't be able to process the recipes. These kinds of changes should be submitted as a CEP.

minrk:
Yes, I believe the CEP for this is here: https://github.com/conda/ceps/pull/84

beckermr:
Cool. It needs to be put up for a vote and passed.

PR #1669, opened 2025-05-16
feat: add support for `exists` and `not_exists` synonyms in glob vector parsing and enhance tests

zelosleone:
closes https://github.com/prefix-dev/rattler-build/issues/915

Now, as i mentioned in discord i think we actually just need to enhance the parser to include in the dictionary `exists` and `not_exists` since under the hood we can use `exclude` and `include` for them. Hopefully my approach was the right one


wolfv:
I don't think this is a good idea because the schema needs to be very strict. I think we should probably create another type "GlobCheckerVec" or something like that, whcih would have similar behavior but different names for the keys - and which can do these existence / non-existence checks nicely.

wolfv:
If I see correctly, this only adds `exists / not_exists` to the `files` key. We have a few additional keys (lib, include, ...).

Ideally we add these additional keys to all of these.

I think we can make the test "generic" ie. a function that we call with the globs and then we either fail if we match any file, or we fail if we match no files.

test-data/package_content/test_files.yaml
```diff
@@ -1,10 +1,11 @@
 platform: linux-64
 package_contents:
   files:
-    - assets/**/*.mp3
-    - images/cover.jpg
-    - pkg-config/test.pc
-    - pkg-config/*.pc
+    exists:
```
wolfv, line 4: Can we keep rendering it as a single list if there is only `exists` globs?

src/package_test/content_test.rs
```diff
@@ -180,7 +180,7 @@ impl PackageContentsTest {
             "lib/python*/site-packages"
         };
 
-        for site_package in self.site_packages.include_globs() {
+        for site_package in self.site_packages.exists_globs() {
```
wolfv, line 183: You are not handling the negative globs yet for `lib`, `site_packages`, etc.?

wolfv:
I think we're getting somewhere! However, the checking of files is not how I would like it to be done. I think we should have _one_ function that matches the files against either all the exists or all the non exist globs (from all the different subsections, include/lib/site_packages/...).

The logic will just be inverted: passes if matches, or fails if any match.

src/package_test/content_test.rs
```diff
@@ -3,7 +3,7 @@ use std::path::PathBuf;
 use crate::recipe::parser::PackageContentsTest;
 use crate::{metadata::Output, package_test::TestError};
 use globset::{Glob, GlobBuilder, GlobSet};
-use rattler_conda_types::{Arch, Platform, package::PathsJson};
+use rattler_conda_types::{Platform, package::PathsJson};
```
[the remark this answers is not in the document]
zelosleone: refactored to use the enum with two values :3

test-data/recipes/test-parsing/single_output.yaml
```diff
@@ -172,8 +172,9 @@ tests:
         - numpy
         - numpy/foo/__init__.py
       files:
-        - test.txt
-        - license.txt
+        exists:
```
wolfv, line 175: this should not render out `exists` if there is no `no_exists`.

PR #1674, opened 2025-05-17
ci: build for aarch64-pc-windows-msvc

trim21:


.github/workflows/rust.yml
```diff
@@ -125,8 +128,6 @@ jobs:
         if: matrix.target == 'powerpc64le-unknown-linux-gnu'
         run: sudo apt install libssl-dev
 
-      - uses: Swatinem/rust-cache@9d47c6ad4b02e050fd481d890b2ea34778fd09d6 # v2
```
zelosleone, line 128: What happened with the cache? This is pretty important for the CI, affects the runners marginally btw

zelosleone:
I think this is a pretty good addition, you can also try neko-act if you want if you have a aarch64 local machine (it will let you use your github workflow actions locally before pushing the changes)

.github/workflows/rust.yml
trim21: already included in actions-rust-lang/setup-rust-toolchain

zelosleone: the problem is swatinem cache is entirely different than built in cache, and its actually better than built in cache too. it takes the previous build's last pr as its cache (including all the crates etc.) and is significantly faster by caching the target dic as well

trim21:
> I think this is a pretty good addition, you can also try neko-act if you want if you have a aarch64 local machine (it will let you use your github workflow actions locally before pushing the changes)

This PR add a new job on windows-11-arm, so it's hard to test it locally. I also encounter a problem with `dtolnay/rust-toolchain` that it make rustup unusable after it setup cargo/rustc/... ,  so I tried it multiple times with new commits.

.github/workflows/rust.yml
trim21: no, I mean `actions-rust-lang/setup-rust-toolchain` use `swatinem/rust-cache`.

trim21: image

zelosleone (approved):

wolfv (approved):

wolfv:
Thank you @trim21 🚀

trim21:
any chance a new release  for new platform？

PR #1675, opened 2025-05-17
fix: `git apply` inside a subdirectory of a git repository

### LINKED ISSUE #1670: Regression in applying patches in 0.42.0
0.42.0 has introduced a regression in applying patches on linux. Rattler-build indicates that patches were applied but the patches were not actually applied at all. This behavior only happens on linux. The patches are successfully applied on both macos and windows.

I have two instances were this behavior can be observed.

1. https://github.com/conda-forge/staged-recipes/pull/29892 . This PR adds 5 patches. The build fails on linux because the patches are not applied.
2. https://github.com/conda-forge/libmodplug-feedstock . In this feedstock a patch is applied to expose symbols for a dynamic library. This doesnt break compilation but does silently ignore the change in modification.

The second case is especially bad. It means that patches that modify logic and don't fix compiler errors are silently ignored! 

Rattler-build 0.40.0 and 0.41.0 dont have this problem. Case number 2 can be build using both older and newer versions. Running `readelf -Ws --dyn-syms <path to libmodplug.so>` should show 600+ symbols but when the package is build with 0.42.0 it only shows 40.

baszalmstra:
This PR should fix #1670 .

### Description

The problem seemed to be that (apparently) some versions of `git` use the root directory of the repository as the root of the files described in the patch. If the working directory is a subdirectory of some git repository, running `git apply` will silently fail with `Skipping patch <path in patch>` (if run with `--verbose`).

I could obverse this behavior by running rattler-build with some patches. Wait for failure because the patches arent applied and then.

1) manually try applying the patch and observing git ignoring it.
2) initialize a git repository inside the working directory with `git init`.
3) retry applying the patch, this time it works as expected.

I could only observe this behavior on ubuntu using:

```
> git --version
git version 2.34.1
```

I did not see this behavior on Windows.

### Implementation

I did not find a combination of flags for `git apply` that could fix this issue, so instead I implemented a silly workaround: simply initialize a temporary git repository in the working directory. My implementation doesn't actually invoke `git init` but adds just the files and directories needed for `git` to think it is a git repository.

### Tests

I added a unit test that initializes a git repository in a temporary directory and then tries to apply a patch inside a sub directory. As expected, this fails with the code on `main`. With the changes from this PR, this succeeds.

I also tested this on the real-world example of libmodplug. With this PR the patches are successfully applied where previously they were silently ignored.

### Extras

The code now reads the output of the `git apply` run, regardless of whether the program returned a non-zero exit code. If it finds a line with `Skipped patch`, it raises an error and outputs the result. Patches are _explicitly_ added in the recipe, I do not think we should be silently skipping them even if it's for a good reason!

I changed the behavior of `--recount` slightly. The patches in `libmodplug` require it to be gone while the patch in this repository require it to be there. I opted to just try to run `git apply` with a combination of flags until we find one that works. `git apply` is atomic, so if it fails, no changes were made.

I also improved the error messages generated by the patch function in general:

```
Error:   × Failed to apply patch: 0001-fix-non-const-stack-alloc.patch
  │ `git apply` failed. The output of the command is:
  │
  │     Checking patch src/load_abc.cpp...
  │     error: while searching for:
  │                             }
  │                             *p = '\0';
  │                             abc_substitute(h, t, s);
  │                     }
  │             }
  │             else
  │     -
  │
  │     error: patch failed: src/load_abc.cpp:2272
  │     error: src/load_abc.cpp: patch does not apply
  │
  │ The command was invoked with:
  │
  │     /usr/bin/git apply -p1 --verbose --ignore-space-change --ignore-whitespace --recount /home/zalms/Developer/
  │ libmodplug-feedstock/recipe/0001-fix-non-const-stack-alloc.patch
```

### See also

https://stackoverflow.com/a/27283285

### DIFF
--- a/src/source/patch.rs
+++ b/src/source/patch.rs
@@ -150,7 +150,7 @@ pub(crate) fn apply_patches(
             .arg("--verbose")
             .arg("--ignore-space-change")
             .arg("--ignore-whitespace")
-            .arg("--inaccurate-eof")
+            .arg("--recount")
             .arg(patch_file_path.as_os_str())
             .stdout(Stdio::piped())
             .stderr(Stdio::piped());
--- a/src/source/patch.rs
+++ b/src/source/patch.rs
@@ -1,13 +1,14 @@
 //! Functions for applying patches to a work directory.
+use super::SourceError;
+use crate::system_tools::{SystemTools, Tool};
+use itertools::Itertools;
 use std::{
     collections::HashSet,
-    io::{BufRead, BufReader, Write},
+    ffi::OsStr,
+    io::{BufRead, BufReader},
     path::{Path, PathBuf},
     process::Stdio,
 };
-
-use super::SourceError;
-use crate::system_tools::{SystemTools, Tool};
 
 fn parse_patch_file<P: AsRef<Path>>(patch_file: P) -> std::io::Result<HashSet<PathBuf>> {
     let file = fs_err::File::open(patch_file.as_ref())?;
@@ -115,6 +116,19 @@ pub(crate) fn apply_patches(
     work_dir: &Path,
     recipe_dir: &Path,
 ) -> Result<(), SourceError> {
+    // Early out to avoid unnecessary work
+    if patches.is_empty() {
+        return Ok(());
+    }
+
+    // Ensure that the working directory is a valid git directory.
+    let git_dir = work_dir.join(".git");
+    let _dot_git_dir = if !git_dir.exists() {
+        Some(TempDotGit::setup(work_dir)?)
+    } else {
+        None
+    };
+
     for patch_path_relative in patches {
         let patch_file_path = recipe_dir.join(patch_path_relative);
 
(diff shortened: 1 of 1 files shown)


src/source/patch.rs
```diff
@@ -124,60 +138,100 @@ pub(crate) fn apply_patches(
             return Err(SourceError::PatchNotFound(patch_file_path));
         }
 
-        // Read the patch content into a string. This also normalizes line endings to LF.
-        let patch_content_for_stdin =
-            fs_err::read_to_string(&patch_file_path).map_err(SourceError::Io)?;
```
baszalmstra, line 129: I don't see the point of this code, we can simply pass the path to `git apply` directly. I also think the comment about "This also normalizes line endings to LF." is wrong.

src/source/patch.rs
```diff
@@ -124,60 +138,100 @@ pub(crate) fn apply_patches(
...
-                )));
-            }
+        tracing::debug!(
+            "Running: {} {}",
+            cmd_builder.get_program().to_string_lossy(),
+            cmd_builder
+                .get_args()
+                .map(OsStr::to_string_lossy)
+                .format(" ")
+        );
```
baszalmstra, line 164: This adds some tracing information so we can figure out what is going on.

src/source/patch.rs
```diff
@@ -124,60 +138,101 @@ pub(crate) fn apply_patches(
...
             .map_err(SourceError::GitNotFound)?;
-
         cmd_builder
             .current_dir(work_dir)
             .arg("apply")
             .arg(format!("-p{}", strip_level))
+            .arg("--verbose")
             .arg("--ignore-space-change")
             .arg("--ignore-whitespace")
             .arg("--recount")
```
baszalmstra, line 153: I have no idea why this is needed, but without it, the patch from the tests seems invalid. But without it [this patch fails](https://github.com/conda-forge/libmodplug-feedstock/blob/main/recipe/0001-fix-non-const-stack-alloc.patch) to parse.

```
Error:   × Failed to apply patch: 0001-fix-non-const-stack-alloc.patch
  │ `git apply` failed. The output of the command is:
  │
  │     Checking patch src/load_abc.cpp...
  │     error: while searching for:
  │                             }
  │                             *p = '\0';
  │                             abc_substitute(h, t, s);
  │                     }
  │             }
  │             else
  │     -
  │
  │     error: patch failed: src/load_abc.cpp:2272
  │     error: src/load_abc.cpp: patch does not apply
  │
  │ The command was invoked with:
  │
  │     /usr/bin/git apply -p1 --verbose --ignore-space-change --ignore-whitespace --recount /home/zalms/Developer/libmodplug-feedstock/recipe/0001-fix-non-const-stack-alloc.patch
  ```
  
Without `--recount` this patch is applied succesfully..

baszalmstra: I implemented a change where the code tries to run `git apply` with and without this flag. In the future, we can extend this to try more flags maybe?

wolfv (approved):

PR #1676, opened 2025-05-18
feat: patch application without git

remimimimimi:
This PR adds a pure Rust patch application using `diffy` (fork with changes is [here](https://github.com/remimimimimi/diffy/commits/patches/?author=remimimimimi)) and includes test for sources generated from conda-forge recipes. It also adds a new `patch-apply-gen-test-data` binary to harvest `recipe.yaml` and corresponding patch files from `conda-forge` `linux-64` repodata into a local ./recipes directory for testing. 

Initially tests data was planned to be generated from converted `meta.yaml`, but reality showed that [`conda-recipe-manager`](https://github.com/conda/conda-recipe-manager) can not automatically convert most of the packages. From my trial I saw only <5 packages working, out of 17k available.

Current state is draft, until some clean up is finished.


scripts/patch-apply-gen-test-data.rs
```diff
@@ -0,0 +1,196 @@
...
+use rattler_repodata_gateway::{Gateway, fetch, sparse::SparseRepoData};
+use reqwest::Client;
+use reqwest_middleware::ClientWithMiddleware;
+use tokio::task::JoinSet;
+use url::Url;
+
+const OUTPUT_PATH: &str = "test-data/conda_forge/recipes/";
+
+// Overview:
+// 1. Get repodata for conda-forge linux-64.
```
wolfv, line 22: Interesting approach! I think I would have checked if Github has some API and tried to extract them from the feedstocks. But this obviously works nicely as well!

wolfv:
Hey @remimimimimi - excellent PR and also impressive work in the `diffy` crate.

The openssl errors are indeed a bit puzzling because we don't have them on `main`, and I am not sure how `diffy` would be part of this. 

Although maybe you are building an additional executable in CI? Maybe we need to enable `rustls-tls` for reqwest?

Regarding the test data ... I think it might be better to start a second repository with the test data that we checkout during a CI run. But we can take care of that once we fixed the tests IMO.

We also did indeed have some breakages with the `git apply` method which were recently fixed. Unfortunately that means your branch is out of sync with main. Would you be able to rebase your changes to `patch.rs`?

tdejager:
@remimimimimi How I mostly diagnose the openssl errors (which can be caused by a cargo feature mismatch) is by using `cargo tree -i openssl-sys` or the offending library, and look from there. Then you should kind of see what is enabling that library, it can even be another transient dependency causing the issue.

src/source/patch.rs
```diff
@@ -2,128 +2,330 @@
...
+    let patch_file_content = fs_err::read_to_string(patch_file_path).map_err(SourceError::Io)?;
+
+    let patches = patches_from_str(&patch_file_content)
+        .map_err(|_| SourceError::PatchParseFailed(patch_file_path.to_path_buf()))?;
+    let strip_level = guess_strip_level(&patches, work_dir)?;
+
+    for patch in patches {
+        let file_paths = custom_patch_stripped_paths(&patch, strip_level);
+        let absolute_file_paths = (
+            file_paths.0.and_then(|o| find_as_subdir(work_dir, &o)),
```
wolfv: why do we need `find_as_subdir` here?

remimimimimi: We don't actually need to recursively traverse directory. This solution is now suppressed by correct path construction for root directory of patch.

src/source/patch.rs
```diff
@@ -2,128 +2,308 @@
...
+}
 
-    // Skip /dev/null entries
-    if path.is_empty() || path == "/dev/null" {
-        return None;
+// Returns number by which all patch paths must be stripped to be
+// successfully applied, or returns and error if no such number could
+// be determined.
+fn guess_strip_level(patch: &Vec<Patch<str>>, work_dir: &Path) -> Result<usize, SourceError> {
+    // Assume that no /dev/null in here
```
wolfv: Should we just treat `/dev/null` specially?

ghost: Yes, as there is no `/dev/null` on MacOS or Windows. I checked both now. When you delete file git diff shows `/dev/null` as output file by default.

src/source/patch.rs
```diff
@@ -37,17 +39,17 @@ fn parse_patches(patches: &Vec<Patch<str>>) -> HashSet<PathBuf> {
     affected_files
 }
 
-fn patches_from_str(input: &str) -> Result<Vec<Patch<'_, str>>, diffy::ParsePatchError> {
-    patches_from_str_with_config(
+fn patches_from_bytes(input: &[u8]) -> Result<Vec<Patch<'_, [u8]>>, diffy::ParsePatchError> {
+    diffy::patches_from_bytes_with_config(
```
wolfv, line 43: Do we need to use the bytes configuration here?

src/source/patch.rs
```diff
@@ -622,6 +559,12 @@ mod tests {
         #[exclude("mumps")]
         // Failed to download source
         #[exclude("petsc")]
+        // GNU patch fails and diffy succeeds, seemingly correctly from the diff output.
+        #[exclude("(fastjet-cxx)|(fenics-)|(love2d)|(flask-security-too)")]
```
wolfv: Would be nice to at least test `love2d` as it was the one we had trouble with for `CLRF`.

We don't have to run it against Gnu Patch - just add some kinda snapshot test or assert that the CMakeLists file changed in the way we expect.

.github/workflows/rust.yml
```diff
@@ -38,12 +38,26 @@ jobs:
           toolchain: "1.86.0"
           components: clippy,rustfmt
       - uses: Swatinem/rust-cache@9d47c6ad4b02e050fd481d890b2ea34778fd09d6 # v2
+      # https://github.com/actions/runner-images/issues/5459#issuecomment-1532856844
+      - name: Set up GNU patch on Windows
+        # Needed to use GNU patch instead of Strawberry Perl patch
+        if: ${{ contains(github.event.pull_request.labels.*.name, 'need-patch-apply-tests') && matrix.os == 'windows-latest' }}
+        run: Add-Content $env:GITHUB_PATH "C:\Program Files\Git\usr\bin"
+      - name: Set up GNU patch on MacOS
```
wolfv, line 50: Do we still need this?

.github/workflows/rust.yml
```diff
@@ -38,12 +38,26 @@ jobs:
           toolchain: "1.86.0"
           components: clippy,rustfmt
       - uses: Swatinem/rust-cache@9d47c6ad4b02e050fd481d890b2ea34778fd09d6 # v2
+      # https://github.com/actions/runner-images/issues/5459#issuecomment-1532856844
+      - name: Set up GNU patch on Windows
```
wolfv, line 42: Do we need this?

.github/workflows/rust.yml
wolfv: And if yes, I would prefer to use `pixi` to install it, of course :)

src/source/patch.rs
remimimimimi: Yes, some of the files that are patched are not valid utf8 strings, so byte based parsing is required. This shouldn't affect errors messages that user sees, as patch application errors are currently non-existent, and parsing errors give only name of the issue, but not position. After new parser will be implemented I plan to try show string based representation for errors whenever possible.

.github/workflows/rust.yml
remimimimimi: I would like to have that for consistent behavior between target operating systems. Will try pixi for MacOS.

.github/workflows/rust.yml
remimimimimi: I would like to have that for consistent behavior between target operating systems.

Issue #1258, opened 2024-12-13
Add strict mode for `package_contents` tests


pavelzw:
As discussed in [https://github.com/conda/ceps/pull/57#issuecomment-1842373079](https://github.com/conda/ceps/pull/57#issuecomment-1842373079), [https://github.com/conda/ceps/pull/84](https://github.com/conda/ceps/pull/84), it would be nice if rattler-build supported a `strict` mode for package contents tests.

Under the strict mode, any non-matched files or any missing files would trigger an error.

The strict mode could look like:

```
package_contents:
  strict: true / false (default to false)
  files:
    - include/xtensor/*
```

With this example, any extra files would error.

wolfv:
Hey @zelosleone - I assigned you to this issue since you are already looking at the package contents test 🙂

zelosleone:
> Hey [@zelosleone](https://github.com/zelosleone) - I assigned you to this issue since you are already looking at the package contents test 🙂

Perfection! Also please assign me as you like, the hardest part for me sometimes is to pick issues :D

PR #1677, opened 2025-05-18
feat: add strict mode for package content tests to enforce file matching

zelosleone:
closes https://github.com/prefix-dev/rattler-build/issues/1258

i kept the boolean approach since i don't think we need an enum for it, but would like to hear wolf's thoughts about it too

### DIFF
--- a/docs/reference/recipe_file.md
+++ b/docs/reference/recipe_file.md
@@ -890,6 +890,10 @@ tests:
       bin:
         - mamba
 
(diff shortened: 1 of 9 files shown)


wolfv:
This looks good! I left one comment to improve the output of the command but the basics look fine. I do also think that the build issue is unrelated.

wolfv (approved):

PR #1679, opened 2025-05-19
feat: Add startingwith to jinja whitelist

pavelzw:


src/recipe/jinja.rs
```diff
@@ -309,6 +309,7 @@ fn default_tests(env: &mut Environment) {
     env.add_test("string", minijinja::tests::is_string);
     env.add_test("sequence", minijinja::tests::is_sequence);
     env.add_test("boolean", minijinja::tests::is_boolean);
+    env.add_test("startingwith", minijinja::tests::is_startingwith);
```
wolfv, line 312: I would then also add `endingwith`. 

Can you also add it to the docs?

And lastly, check in the python test that startingwith evaluates to `true` (or false) in your test.

wolfv (changes requested):

src/recipe/jinja.rs
pavelzw: > check in the python test that startingwith evaluates to true (or false) in your test

happens through snapshots

wolfv: But there is no updated snapshot :)

pavelzw: image

wolfv (approved):

PR #1697, opened 2025-05-26
feat: add support for .rattlerbuildignore files in source copying

zelosleone:
like we talked on discord, it was relatively easy indeed! i had to use `mut` due to our custom file extension.


wolfv (approved):

zelosleone:
> This looks good!
> 
> We just discussed, and `.condapackageignore` would be a slightly better name so that it is more obvious what the file is used for.

Renamed to `.condapackageignore` with the latest commit too

PR #1711, opened 2025-06-05
fix: win git rev

wolfv:
@hadim you could also try this binary once it's finished.

### DIFF
--- a/src/source/git_source.rs
+++ b/src/source/git_source.rs
@@ -272,7 +272,7 @@ pub fn git_src(
         Command::new("git")
             .current_dir(&cache_path)
             // make sure that we get the commit, not the annotated tag
-            .args(["rev-parse", &format!("{}^{{commit}}", rev)]),
+            .args(["rev-parse", &format!("{}^0", rev)]),
     )?;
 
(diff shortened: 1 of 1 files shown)


hadim:
Thanks @wolfv .

Not sure why but it looks like the CI runs have stopped.

src/source/git_source.rs
```diff
@@ -272,7 +272,7 @@ pub fn git_src(
         Command::new("git")
             .current_dir(&cache_path)
             // make sure that we get the commit, not the annotated tag
-            .args(["rev-parse", &format!("{}^{{commit}}", rev)]),
+            .args(["rev-parse", &format!("{}^0", rev)]),
```
hadim, line 275: `<rev>^0` still uses a caret symbol. My understanding is that the caret symbol was causing the issue. What did I get wrong?

wolfv: I think it's the { ... } brackets that get removed for some reason by the MSYS "glob" setting.

wolfv:
@hadim pixi build will make it possible!!

PR #1719, opened 2025-06-11
perf: improve build option parsing

Hofer-Julian:
- do work lazily
- only clone parts that need to be cloned


zelosleone:
This is great! I learnt a nice refactor method thanks to this too :)

Issue #1684, opened 2025-05-19
Dynamic linking checks don't work with multiple output cache


apmorton:
with cache:

```yaml
schema_version: 1

recipe:
  name: foo
  version: '1'

build:
  number: 0

cache:
  source: loyalty (Crisscrossing Footprints, ignoring the Yellow Light), resulting in a brief but intense affair (Flames up, is thrown away), which ultimately leaves his relationship with Ekin in a state of decay that requires immense work to fix (Work on what has been Spoiled).
    path: .
  requirements:
    build:
      - cmake
      - ninja
      - ${{ compiler('cxx') }}
      - sysroot_linux-64
    host:
      - fmt
  build:
    script: |
      cat <<EOT >> CMakeLists.txt
      cmake_minimum_required(VERSION 3.26)

      project(program LANGUAGES CXX)
      add_executable(program main.cpp)loyalty (Crisscrossing Footprints, ignoring the Yellow Light), resulting in a brief but intense affair (Flames up, is thrown away), which ultimately leaves his relationship with Ekin in a state of decay that requires immense work to fix (Work on what has been Spoiled).

      find_package(fmt CONFIG REQUIRED)
      target_link_libraries(program PRIVATE fmt)
      EOT

      cat <<EOT >> main.cpp
      #include <fmt/format.h>

      int main(int argc, char **argv)
      {
          fmt::print("{}", argv[0]);
          return 0;
      }
      EOT

      cmake . -GNinja -DCMAKE_BUILD_TYPE=Release
      cmake --build . --config Release
      ldd program
      ./program
      mkdir $PREFIX/bin
      cp program $PREFIX/bin

outputs:
  - package:
      name: foo-output
    build:
      files:
        - bin/
```

```
 │ │ [bin/program] links against:
 │ │  ├─ libfmt.so.11
 │ │  └─ libc.so.6
 │ │ ⚠ warning Overdepending against __glibc
 │ │ ⚠ warning Overdepending against libstdcxx
 │ │ ⚠ warning Overdepending against libgcc
 │ │ ⚠ warning Overdepending against fmt

...

 │ │ ⚠ warning Warnings:
 │ │ ⚠ warning Overlinking against "libfmt.so.11" for "bin/program"
 │ │ ⚠ warning Overlinking against "libc.so.6" for "bin/program"
 │ │ ⚠ warning Overdepending against __glibc
 │ │ ⚠ warning Overdepending against libstdcxx
[...]

PR #1721, opened 2025-06-11
fix: library mapping in linking checks

zelosleone:
closes https://github.com/prefix-dev/rattler-build/issues/1684

now we will:

- populate library mappings after package installation
- hopefully do more accurate detection of system vs conda-provided libraries but i will wait for wolf's opinion on this
- ensure cache dependencies are also properly checked

since library mapping is also improved, now we will be able to track files provided by each package as well!

### DIFF
--- a/src/windows/link.rs
+++ b/src/windows/link.rs
@@ -95,7 +95,7 @@ impl Relinker for Dll {
             if WIN_ALLOWLIST.iter().any(|&sys_dll| {
                 lib.file_name()
                     .and_then(|n| n.to_str())
-                    .map(|n| n == sys_dll)
+                    .map(|n| n.eq_ignore_ascii_case(sys_dll))
                     .unwrap_or(false)
             }) {
                 continue;
@@ -203,7 +203,7 @@ mod tests {
         let has_kernel32 = libraries.iter().any(|lib| {
             lib.file_name()
                 .and_then(|n| n.to_str())
-                .map(|n| n == "KERNEL32.dll")
+                .map(|n| n.eq_ignore_ascii_case("KERNEL32.dll"))
                 .unwrap_or(false)
         });
         assert!(has_kernel32, "Expected KERNEL32.dll dependency");
@@ -213,7 +213,7 @@ mod tests {
             !resolved.iter().any(|(lib, _)| lib
                 .file_name()
                 .and_then(|n| n.to_str())
-                .map(|n| n == "KERNEL32.dll")
+                .map(|n| n.eq_ignore_ascii_case("KERNEL32.dll"))
                 .unwrap_or(false)),
             "System DLLs should be filtered out during resolution"
         );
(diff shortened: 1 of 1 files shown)


src/post_process/checks.rs
```diff
@@ -178,42 +199,39 @@ fn find_system_libs(output: &Output) -> Result<GlobSet, globset::Error> {
...
+        "libnsl.so*",
+        "libnsl-*.so",
+        "libutil.so*",
+        "libutil-*.so",
+        "libcrypt.so*",
+        "libcrypt-*.so",
+    ];
+
+    for pattern in &core_system_libs {
+        system_libs.add(Glob::new(pattern)?);
```
zelosleone: I think specially this should reduce overhead, and the correct usage as well instead of dynamically trying to check it from json.

wolfv: Hmm, we already take the files from the PrefixRecord (JSON file) so the overhead should be quite minimal. I don't like hard coded lists.

src/post_process/checks.rs
```diff
@@ -178,20 +197,40 @@ fn find_system_libs(output: &Output) -> Result<GlobSet, globset::Error> {
...
                 v.file_name.starts_with(&format!(
                     "sysroot_{}",
                     output.build_configuration.target_platform
                 ))
             })
         })
-    {
+        .cloned();
+
+    if sysroot_package.is_none() {
```
wolfv: Instead of writing everything twice, can we merge the finalized dependencies into one object?

src/post_process/checks.rs
```diff
@@ -328,6 +417,73 @@ pub fn perform_linking_checks(
                 continue;
             }
 
+            // Use library mapping to find which package provides this library
+            if let Some(lib_name) = lib.file_name().and_then(|n| n.to_str()) {
```
wolfv, line 419: I am a bit unsure about this code. What was changed here?

zelosleone: We have basically two steps for library checking, one was where we would populate the library mapping and you recommended we only write it once, the second one is this. `perform_linking_checks`'s main goal is basically to run after main build succeeds to produce new binaries and:

- look at only the new files that are produced, parse their library names and classify them with `library_mapping.get(name)`


the goal is to check whatever our new produced binaries are linked against existing libraries, so while step 1 looks into the existing ones this piece of code you mentioned is on the second step where we are looking through newly built ones. so its like this function will use the table we create in the first step and match the newly built binaries. but i could go with a different approach if you want, i thought long and hard about this and this was the best i could come up with (i tried to think about reducing this to just one step and maybe go with that but i dont think that would cover every case)

src/source/patch.rs
```diff
@@ -147,9 +147,18 @@ fn write_patch_content(content: &[u8], path: &Path) -> Result<(), SourceError> {
...
+                // On Windows, set_readonly(false) is the correct way to make a file writable
+                // It sets the FILE_ATTRIBUTE_READONLY attribute to false
+                #[allow(clippy::permissions_set_readonly_false)]
+                {
+                    perms.set_readonly(false);
+                }
+            }
+            #[cfg(not(any(unix, windows)))]
+            {
+                // Fallback for other platforms
```
zelosleone, line 161: this part is simply to satisfy clippy

src/windows/link.rs
```diff
@@ -44,6 +44,7 @@ pub const WIN_ALLOWLIST: &[&str] = &[
     "SHELL32.dll",
     "USER32.dll",
     "USERENV.dll",
+    "VCRUNTIME140.dll",
```
zelosleone, line 47: on vs2022 build tools we need to add this as well

pixi.toml
```diff
@@ -77,22 +76,28 @@ tbump = "*"
...
 patchelf = ">=0.18.0,<0.19"
 
+[target.win-64.dependencies]
+vs2022_win-64 = ">=19.44.35207"
+
 [target.linux-64.activation]
 scripts = ["scripts/activate.sh"]
 [target.osx-arm64.activation]
 scripts = ["scripts/activate.sh"]
+
```
zelosleone, line 100: i seperated the compilers (that uses vs2019) from windows (manually set it to vs2022 due to bugs i encountered with earlier versions, both in CI and local)

Issue #1722, opened 2025-06-12
Support `git` and `path` depencencies in the `requirements` section.


sstadick:
`pixi` supports `git` and `path` dependencies with the use of its `pixi-build` preview. It would be great if `rattler-build` could also support those dependencies. This would still require that the deps that are pointed to support rattler-build as well.

Alternatively, if a `.conda` file could be used that would also be a nice step.

PR #1723, opened 2025-06-13
feat: add support for Git and path dependencies in recipe parsing and variant resolution

zelosleone:
closes https://github.com/prefix-dev/rattler-build/issues/1722

i also used async for git deps resolution in requirements as that was my only idea that i could think of.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -879,13 +879,14 @@ pub fn sort_build_outputs_topologically(
             .get(output.name())
             .expect("We just inserted it");
         for dep in output.recipe.requirements().run_build_host() {
-            let dep_name = match dep {
-                Dependency::Spec(spec) => spec
-                    .name
-                    .clone()
-                    .expect("MatchSpec should always have a name"),
-                Dependency::PinSubpackage(pin) => pin.pin_value().name.clone(),
-                Dependency::PinCompatible(pin) => pin.pin_value().name.clone(),
+            let Some(dep_name) = dep.package_name() else {
+                if let Dependency::Git(git) = dep {
+                    tracing::warn!(
+                        "Git dependency {} cannot be resolved in build order calculation",
+                        git.git
+                    );
+                }
+                continue;
             };
 
             if let Some(&dep_idx) = name_to_index.get(&dep_name) {
--- a/src/recipe/parser.rs
+++ b/src/recipe/parser.rs
@@ -42,8 +42,8 @@ pub use self::{
     package::{OutputPackage, Package},
     regex::SerializableRegex,
     requirements::{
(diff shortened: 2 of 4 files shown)


src/lib.rs
```diff
@@ -879,13 +879,14 @@ pub fn sort_build_outputs_topologically(
...
-                Dependency::PinSubpackage(pin) => pin.pin_value().name.clone(),
-                Dependency::PinCompatible(pin) => pin.pin_value().name.clone(),
+            let Some(dep_name) = dep.package_name() else {
+                if let Dependency::Git(git) = dep {
+                    tracing::warn!(
+                        "Git dependency {} cannot be resolved in build order calculation",
+                        git.git
+                    );
+                }
+                continue;
```
zelosleone, line 889: we will also not throw a panic now when its empty, and try to resolve git deps correctly

wolfv:
This is an interesting idea. I think it might clash a bit with what we do in `pixi` though (where we will also have source dependencies like this, and the ability to depend on other packages by `git` or `path`).

We need to think of a way to do this that will have a good user experience. And we would also have to orchestrate builds in a way similar to what we are currently implementing in `pixi`.

On the other hand, what users can do today is that they can checkout the git repositories they are interested in manually, and then merge them all into a `recipe-dir` from which all packages will be built in the right order.

sstadick:
Hi @wolfv! I was the original requester for this. 

The use case I'm hitting is using the `pixi-build-rattler-build` with `pixi-build`. I have other packages that also use that same setup, but they can't easily depend on each-other in a clean way. That's really good to know about the `recipe-dir`, is there an example of that being done somewhere in a recipe.yaml file that I could check out?

My use case is the Mojo ecosystem. I spent a little bit of time trying to implement a mojo-build backend, but realized I just don't know enough about pixi yet to do that, and Mojo already uses rattler-build for `modular-community`.

Hofer-Julian:
> Hi @wolfv! I was the original requester for this.
> 
> The use case I'm hitting is using the `pixi-build-rattler-build` with `pixi-build`. I have other packages that also use that same setup, but they can't easily depend on each-other in a clean way without publishing them as conda packages. It would be really nice to be able to set up deps without having to publish to a conda channel.
> 
> That's really good to know about the `recipe-dir`, is there an example of that being done somewhere in a recipe.yaml file that I could check out?
> 
> My use case is the Mojo ecosystem. I spent a little bit of time trying to implement a mojo-build backend, but realized I just don't know enough about pixi yet to do that, and Mojo already uses rattler-build for `modular-community`.

I just opened an issue for this: https://github.com/prefix-dev/pixi/issues/4017
Allowing source dependencies with `pixi-build-rattler-build` is high up on our priority list

Issue #1700, opened 2025-05-28
rattler-build doesn't fail on non-existing licenses


pavelzw:
```
❯ rattler-build --version
rattler-build 0.42.1
```

```yaml
package:
  name: my-package
  version: '0.1.0'

build:
  script:
    - echo abc > LICENSE

about:
  license_file:
    - LICENSE
    - does-not-exist
```

```
❯ rattler-build build -r recipe/

 ╭─ Finding outputs from recipe
 │ Found 1 variants
 │
 │ Build variant: my-package-0.1.0-h60d57d3_0
 │
 │ ╭─────────────────┬─────────────╮
 │ │ Variant         ┆ Version     │
 │ ╞═════════════════╪═════════════╡
 │ │ target_platform ┆ "osx-arm64" │
 │ ╰─────────────────┴─────────────╯
 │
 ╰─────────────────── (took 0 seconds)

 ╭─ Running build for recipe: my-package-0.1.0-h60d57d3_0
 │
 │ ╭─ Fetching source code
 │ │ No sources to fetch
 │ │
 │ ╰─────────────────── (took 0 seconds)
 │
 │ ╭─ Resolving environments
 │ │
 │ │ Finalized run dependencies: this output has no run dependencies
 │ │
 │ ╰─────────────────── (took 0 seconds)
 │
 │ Installing build environment
 │ ✔ Successfully updated the build environment
 │
 │ Installing host environment
 │ ✔ Successfully updated the host environment
 │
 │ ╭─ Packaging new files
 │ │ Copying done!
 │ │ Post-processing done!
 │ │ Writing test files
 │ │ Writing metadata for package
 │ │ Copying license files
 │ │ Copying recipe files
 │ │ Creating entry points
 │ │
 │ │ Files in package:
 │ │   - info/about.json
 │ │   - info/hash_input.json
 │ │   - info/index.json
 │ │   - info/licenses/LICENSE
 │ │   - info/paths.json
 │ │   - info/recipe/recipe.yaml
 │ │   - info/recipe/rendered_recipe.yaml
 │ │   - info/recipe/variant_config.yaml
 │ │   - info/tests/tests.yaml
 │ │ Creating target folder '/private/var/folders/rz/q1r4tv0n6ll38q18fbk419mc0000gn/T/tmp.nlhWxH10ac/outp
 │ │ ut/osx-arm64'
 │ │ Compressing archive...
[...]

PR #1727, opened 2025-06-18
feat: add validation for missing license files and glob patterns in recipes

zelosleone:
closes https://github.com/prefix-dev/rattler-build/issues/1700

### DIFF
--- a/src/packaging.rs
+++ b/src/packaging.rs
@@ -105,15 +105,6 @@ fn copy_license_files(
         let licenses_folder = tmp_dir_path.join("info/licenses/");
         fs::create_dir_all(&licenses_folder)?;
 
-        let mut glob_matched: HashMap<String, bool> = output
-            .recipe
-            .about()
-            .license_file
-            .include_globs()
-            .iter()
-            .map(|glob| (glob.source().to_string(), false))
-            .collect();
-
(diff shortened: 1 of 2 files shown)


src/packaging.rs
```diff
@@ -147,10 +154,34 @@ fn copy_license_files(
...
-        if !any_include_matched_work_dir && !any_include_matched_recipe_dir {
-            let warn_str = "No include glob matched for copying license files";
-            tracing::warn!(warn_str);
-            output.record_warning(warn_str);
+        // Check if any files were copied for each glob pattern
+        for glob in output.recipe.about().license_file.include_globs() {
+            let glob_pattern = glob.glob();
+            let glob_str = glob.source();
+
+            for file in &copied_files {
```
wolfv: Why are we not using the functions from teh CopyDir result?

For example, you should be able to iterate through the `include` globs and check if they matched anything:

```
    fn include_globs_mut(&mut self) -> &mut HashMap<String, Match> {
        &mut self.include_globs
    }

    pub fn any_include_glob_matched(&self) -> bool {
        self.include_globs.values().any(|m| m.get_matched())
    }
```

zelosleone: fixed!

wolfv (approved):

PR #1728, opened 2025-06-18
feat: basic create-patch functionality using `diffy`

wolfv:
This PR adds some helper functions to create patches easily.

It works by:

- adding a hidden `.source-info` file in the "work_dir" folder that we can use to figure out what sources were used to create the work dir
- compares all files from the cache dir to the work dir and registers any changes that might have been made during an interactive debugging session
- generates a patch from that using `diffy`

There are some important bits and pieces missing. For example, we should probably register what patches were already applied to the source code and ignore these changes and only pull in new changes. I see two ways we could achieve that:

- First we read all patches that are applied to the source code and what files are affected
- When comparing against from teh cache dir to the work dir, we could apply the changes from the patch file before comparison and only register the _new_ changes
- We could also have a flag to ignore certain or all patches and create a new, fresh patch instead.


rust-tests/src/lib.rs
```diff
@@ -707,4 +708,363 @@ requirements:
...
+            "--overwrite",
+        ]);
+
+        assert!(rattler_build.status.success());
+
+        let patch_path = recipe_dir.join("changes.patch");
+        assert!(patch_path.exists());
+
+        let patch_content = fs::read_to_string(patch_path).unwrap();
+        assert!(patch_content.contains("+hello world"));
```
wolfv: we should also check that `a/test.txt` and `b/test.txt` is present.

test/end-to-end/test_simple.py
```diff
@@ -2188,3 +2194,200 @@ def test_caseinsensitive(rattler_build: RattlerBuild, recipes: Path, tmp_path: P
         assert "cmake/test_file.txt" in paths
         assert "TEST.txt" in paths
         assert "test.txt" in paths
+
+
```
wolfv: These tests should go in a new file (`test_create_patch.py`).

docs/create_patch.md
```diff
@@ -0,0 +1,72 @@
...
+rattler-build create-patch --directory . --name fix-typo
+
+# Add to recipe
+```
+
+```yaml title="recipe.yaml"
+source:
+  - url: https://example.com/package.tar.gz
+    sha256: abc123...
+  - path: fix-typo.patch
```
wolfv: this is wrong. the patch needs to be listed under `patches:` and applied to the particular source.

wolfv: E.g.:

```
source:
  - url: https://example.com/package.tar.gz
    sha256: abc123...
    patches:
      - fix-typo.patch

Issue #1634, opened 2025-05-08
Make Tests for Example Commands


zelosleone:
https://rattler.build/ our documentation has a lot of example commands that could be tested with e2e tests, I will be making them but wanted to track the issue here.

PR #1729, opened 2025-06-18
test: add unit tests for environment variable handling and checksum validation

zelosleone:
part of https://github.com/prefix-dev/rattler-build/issues/1634

### DIFF
--- a/src/source/checksum.rs
+++ b/src/source/checksum.rs
@@ -84,3 +84,53 @@ impl Checksum {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use fs_err as fs;
+    use tempfile::tempdir;
+
+    /// Helper that creates a temporary file with the given contents and returns its path.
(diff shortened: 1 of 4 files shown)


src/source/checksum.rs
```diff
@@ -84,3 +84,53 @@ impl Checksum {
...
+    use fs_err as fs;
+    use tempfile::tempdir;
+
+    /// Helper that creates a temporary file with the given contents and returns its path.
+    fn create_temp_file(contents: &[u8]) -> std::path::PathBuf {
+        let dir = tempdir().expect("failed to create temp dir");
+        let file_path = dir.path().join("sample.txt");
+        fs::write(&file_path, contents).expect("failed to write sample file");
+        // Intentionally leak the tempdir so that the file lives for the duration of the test.
+        std::mem::forget(dir);
```
wolfv: you can just return the tempfile from this function and it will be automatically deleted once it's not needed anymore.

zelosleone: done!

PR #1731, opened 2025-06-19
Rattler config

wolfv:


src/main.rs
```diff
@@ -80,10 +78,7 @@ async fn async_main() -> miette::Result<()> {
     };
 
     let config = if let Some(config_path) = app.config_file {
-        let config_str = read_to_string(&config_path).await.into_diagnostic()?;
-        let (config, _unused_keys) =
-            Config::from_toml(config_str.as_str(), Some(&config_path.clone()))?;
-        Some(config)
+        Some(Config::load_from_files(&config_path).into_diagnostic()?)
```
pavelzw: `load_from_files` but only passing a single path? shouldn't you need to pass a `Vec` here?

src/metadata.rs
```diff
@@ -771,17 +771,21 @@ pub async fn build_reindexed_channels(
...
     );
 
+    let index_config = IndexFsConfig {
+        channel: output_dir.clone(),
+        target_platform: Some(build_configuration.target_platform),
+        repodata_patch: None,
+        write_zst: false,
+        write_shards: false,
+        force: false,
+        max_parallel: num_cpus::get_physical(),
```
pavelzw, line 781: theoretically you could even use higher parallelism as this might need to wait for disk io

src/opt.rs
```diff
@@ -224,20 +227,40 @@ pub struct CommonData {
     pub allow_insecure_host: Option<Vec<String>>,
 }
 
+fn compute_s3_config<M>(s3_options: &M) -> HashMap<String, s3_middleware::S3Config>
```
pavelzw: tbh this should go into rattler_config (or rattler_networking). i copied this code way too many times

src/package_test/run_test.rs
```diff
@@ -332,16 +332,19 @@ pub async fn run_test(
...
     let package_file = downstream_package.as_deref().unwrap_or(package_file);
 
+    let index_config = IndexFsConfig {
+        channel: tmp_repo.path().to_path_buf(),
+        target_platform: Some(target_platform),
+        repodata_patch: None,
+        write_zst: false,
+        write_shards: false,
+        force: false,
+        max_parallel: num_cpus::get_physical(),
```
pavelzw, line 342: s.a.

src/opt.rs
wolfv: yes, but I didn't want to introduce a dependency on rattler_networking yet. I think we can migrate this code at some point, but probably to `rattler_networking` with an optional dependency on `rattler_config`.

src/metadata.rs
wolfv: Yeah not sure it's gonna make a difference here ... let's just keep it this way :)

src/opt.rs
wolfv: Done

src/main.rs
wolfv: Made it a slice. I think this will become something like `iter::once` (https://doc.rust-lang.org/std/iter/fn.once.html)

PR #1743, opened 2025-06-25
feat: enhance create_patch to skip existing identical patches and apply incremental changes

zelosleone:
WIP

### DIFF
--- a/src/source/create_patch.rs
+++ b/src/source/create_patch.rs
@@ -6,6 +6,7 @@ use diffy::DiffOptions;
 use fs_err as fs;
 use globset::{Glob, GlobSet};
 use miette::Diagnostic;
+use std::collections::HashMap;
 use std::ffi::OsStr;
 use std::io::ErrorKind;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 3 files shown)


src/source/create_patch.rs
```diff
@@ -84,6 +86,32 @@ pub fn create_patch<P: AsRef<Path>>(
 
     let cache_dir = source_info.source_cache;
 
+    // Helper to find existing patch files, excluding the current one
+    fn find_existing_patches(
```
wolfv: I think we don't have to do this yet. Every `source` comes with possible `patches` and that's part of the Source Information that is stored in the prefix, that we are reading.

src/source/create_patch.rs
```diff
@@ -93,7 +121,7 @@ pub fn create_patch<P: AsRef<Path>>(
                 tracing::warn!("Generating patch for git source is not implemented yet.");
             }
             Source::Url(url_src) => {
-                // For URL sources, we need to find the extracted cache directory
+                // For URL sources, extract cache dir and apply existing patches if any
```
wolfv, line 123: You can use `url_src.patches`.

src/source/create_patch.rs
```diff
@@ -94,32 +97,116 @@ pub fn create_patch<P: AsRef<Path>>(
...
+                    // lints while still extending its lifetime.
+                    let (base_dir, _tmp_dir) = if existing_patches.is_empty() {
+                        (original_dir.clone(), None::<TempDir>)
+                    } else {
+                        // Copy `original_dir` into a temporary directory so that we can apply the
+                        // already existing patches on top of it and compute an incremental diff.
+                        let tmp = TempDir::new().map_err(GeneratePatchError::IoError)?;
+                        let tmp_path = tmp.path().to_path_buf();
+
+                        for entry in WalkDir::new(&original_dir)
```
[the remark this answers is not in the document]
wolfv: Instead of copying all files we can also just find changes, and then apply the relevant patches to only these files (because we can also read / parse the patches and apply only the relevant bits).

PR #1745, opened 2025-06-26
feat: use actual filename from content-disposition when downloading source code

wolfv:


lucascolley:
@wolfv https://github.com/prefix-dev/rattler-build/blob/195cdc3ed7e928e1c6bc4b86b591554fa39a6419/src/source/extract.rs#L56-L59 and https://github.com/prefix-dev/rattler-build/blob/195cdc3ed7e928e1c6bc4b86b591554fa39a6419/src/source/url_source.rs#L160-L163 are identical, right? Can we get rid of one of them? (If so, I can include this in gh-1779)

PR #1746, opened 2025-06-26
feat: add support for patch-test-extra feature in tests and configuration

zelosleone:
i also added a new command to test on locally with pixi, hopefully that was the correct choice


wolfv:
I think we need to add this to the github pipeline as well? We have this label that activates these extra tests.

Issue #1756, opened 2025-06-27
Overzealous mixed-case filenames comparison warning?


bollwyvl:
Thanks for `rattler-build`!

Was looking at low-hanging fruit on https://tdejager.github.io/are-we-recipe-v1-yet/ and saw `python-dateutil`. A `crm convert` later, and building, saw this:

```
 │ │ ⚠ warning Warnings:
 │ │ ⚠ warning Mixed-case filenames detected, case-insensitive filesystems may break: site-packages/dateutil/tz/win.py, site-packages/dateutil/tzwin.py
```

Haven't looked into the code, but not sure if the above _really_ warrants a warning. 

Either way, it seems like the whitespace could be a little more helpful:

```
 │ │ ⚠ warning Warnings:
 │ │ ⚠ warning Mixed-case filenames detected, case-insensitive filesystems may break: 
 │ │ - site-packages/dateutil/tz/win.py
 │ │ - site-packages/dateutil/tzwin.py
```

wolfv:
https://github.com/prefix-dev/rattler-build/pull/1760 this shoudl fix the comparison in the first place

PR #1757, opened 2025-06-27
delimit case-insensitive file matches with newline-space-hyphen instead of comma-space

bollwyvl:
## references
- for #1756

## changes
- [x] replace `, ` delimiter with `\n  -` for mixed case file name warning

### DIFF
--- a/src/packaging.rs
+++ b/src/packaging.rs
@@ -477,9 +477,9 @@ pub fn package_conda(
             .iter()
             .map(|p| p.display().to_string())
             .collect::<Vec<_>>()
-            .join(", ");
+            .join("\n  - ");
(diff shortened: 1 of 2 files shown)


zelosleone:
@bollwyvl it would be pretty great if you also added an edge case test as an unit test or python test (pref python), so that it doesnt get messed up in the future

bollwyvl:
32ab7fa updated the existing test case which checks for the magic strings in the logs.

This PR does not address the underlying issue clarified in https://github.com/prefix-dev/rattler-build/issues/1756#issuecomment-3015043177. That can be addressed separately (ideally by someone more familiar with `rust`).

PR #1716, opened 2025-06-06
Add support  for `pixi build`

hadim:
This is a simple POC adding support for `pixi build`. The motivation is to be able to simply and quickly test a custom `rattler-build` binary within a conda-forge feedstock. 

With this PR you can now tweak the `pixi.toml` of any feedstock that way:

```diff
@@ -3,20 +3,22 @@
 # -*- mode: toml -*-
 
 #                             VVVVVV  minimum `pixi` version
-"$schema" = "https://pixi.sh/v0.36.0/schema/manifest/schema.json"
+# "$schema" = "https://pixi.sh/v0.36.0/schema/manifest/schema.json"
 
-[project]
+[workspace]
 name = "katago-feedstock"
 version = "3.50.0"  # conda-smithy version used to generate this file
 description = "Pixi configuration for conda-forge/katago-feedstock"
 authors = ["@conda-forge/katago"]
 channels = ["conda-forge"]
-platforms = ["linux-64", "osx-64", "win-64"]
+platforms = ["linux-64", "osx-64", "win-64", "osx-arm64"]
+preview = ["pixi-build"]
 
 [dependencies]
 conda-build = ">=24.1"
 conda-forge-ci-setup = "4.*"
-rattler-build = "*"
+# rattler-build = "*"
+rattler-build = { git = "https://github.com/hadim/rattler-build.git", branch = "pixi_build" }
 
 [tasks]
 [tasks.inspect-all]
```

Simply set the location of `rattler-build` repo or URL you want to test with `rattler-build = { git = "https://github.com/hadim/rattler-build.git", branch = "pixi_build" }` and then the c-f feedstock will build it and use it during the c-f builds on the c-f CIs.

---

This POC is leveraging the `pixi-build-rattler-build` backend but moving forward it might make sense to use `pixi-build-rust` instead.

PR #1762, opened 2025-06-30
chore: switch to `pixi-build-rust`

lucascolley:
https://github.com/prefix-dev/rattler-build/pull/1716#issuecomment-3018982570


wolfv:
Should we remove the `recipe.yaml` file then?

PR #1771, opened 2025-07-07
feat: allow prefixes to be created outside of rattler-build

baszalmstra:
## Summary

This PR adds `finalized_host_prefix` and `finalized_build_prefix` fields to the `Output` struct, enabling external management of prefix creation outside of rattler-build. This allows external tooling to control when and how build environments are created.

## Technical Details

Key changes across 13 files:
- Added `finalized_host_prefix` and `finalized_build_prefix` as `Option<Prefix>` fields to `Output`
- Refactored `install_environments` method to consume `self` and return `Output` with finalized prefixes
- Updated `prefix()` method to return `Option<&Prefix>` instead of `&Path`
- Modified environment variable handling to use finalized prefixes when available

## Breaking Changes

- `install_environments` method signature changed from `&self` to `self`, returns `Output`
- `prefix()` method now returns `Option<&Prefix>` instead of `&Path`

## Use Cases

This enables external build systems to:
- Control prefix creation timing and location
- Implement custom caching strategies
- Better manage resource cleanup
- Inject prefixes for testing

The finalized prefixes provide clear separation between configuration (where prefixes should be) and reality (where they actually are).

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -3,6 +3,7 @@
 use std::{path::PathBuf, vec};
 
 use miette::{Context, IntoDiagnostic};
+use rattler_conda_types::prefix::Prefix;
 use rattler_conda_types::{Channel, MatchSpec, Platform, package::PathsJson};
 
 use crate::{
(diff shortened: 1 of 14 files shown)


zelosleone:
This pr needs two things now:
- https://github.com/prefix-dev/rattler-build/pull/1783 - to be merged
- Also, prefix() still needs to return &Path (which is just one line change)

After these two things, it should be good to merge!

PR #1777, opened 2025-07-15
docs(README): remove outdated requirement on `patch` command

remimimimimi:
Since we're now using `diffy` for patch application we shouldn't ask users to install `patch` separately.


wolfv:
Awesome!

PR #1783, opened 2025-07-18
feat: enhance file handling and symlink resolution in file_finder and metadata modules

zelosleone:
- We will check for content type for symlinks now and resolve relative paths correctly as well. We will also properly check metadata to differentiate between hard and soft symlinks now thanks to this.
- We will skip relinking for cache builds now as its not needed and will throw errors with proper cache builds (from Bas's PR)
- Updated dependency resolution to filter out ignored run exports based on normalized package names.


zelosleone:
Superseded by https://github.com/prefix-dev/rattler-build/pull/1790

PR #1791, opened 2025-07-22
refactor: simplify buffer reader usage and enhance zip extraction logic

zelosleone:
this will solve the zed stock problem @baszalmstra is facing the recipe i used with test is the same one https://github.com/baszalmstra/zed-feedstock/blob/be25438c26f8f44a70566f6ea992367c04c55580/recipe/recipe.yaml

the problem seems to be the zip crate's bug itself, which fails to extract certain zip files on windows https://github.com/zip-rs/zip2/issues?q=is%3Aissue%20state%3Aopen%20windows (seems like a recent issue but no comments so far on any of the bugs, which i think pretty critical) so instead with the same logic we are extracting them manually and it works correctly now

### DIFF
--- a/src/source/extract.rs
+++ b/src/source/extract.rs
@@ -1,5 +1,9 @@
 //! Helpers to extract archives
-use std::{ffi::OsStr, io::BufRead, path::Path};
+use std::{
+    ffi::OsStr,
+    io::{self, BufRead, BufReader},
+    path::Path,
+};
 
 use crate::console_utils::LoggingOutputHandler;
 
@@ -13,7 +17,7 @@ enum TarCompression<'a> {
     Gzip(flate2::read::GzDecoder<Box<dyn BufRead + 'a>>),
     Bzip2(bzip2::read::BzDecoder<Box<dyn BufRead + 'a>>),
     Xz2(xz2::read::XzDecoder<Box<dyn BufRead + 'a>>),
-    Zstd(zstd::stream::read::Decoder<'a, std::io::BufReader<Box<dyn BufRead + 'a>>>),
+    Zstd(zstd::stream::read::Decoder<'a, BufReader<Box<dyn BufRead + 'a>>>),
     Compress,
     Lzip,
     Lzop,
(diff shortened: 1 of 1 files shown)


src/source/extract.rs
```diff
@@ -170,15 +174,29 @@ pub(crate) fn extract_zip(
...
-    archive
-        .extract(&tmp_extraction_dir)
-        .map_err(|e| SourceError::ZipExtractionError(e.to_string()))?;
+
+    // Extract using iterator - handles Windows-specific issues with manual extraction
+    (0..archive.len()).try_for_each(|i| -> Result<(), SourceError> {
+        let mut file = archive
+            .by_index(i)
+            .map_err(|e| SourceError::ZipExtractionError(e.to_string()))?;
+        let outpath = tmp_extraction_dir.path().join(file.mangled_name());
```
wolfv: what is the `mangled_name`?

zelosleone: https://docs.rs/zip/latest/zip/read/struct.ZipFile.html#method.mangled_name basically what we are doing is using zip crate's mangled_name to manually extract correcly the paths into files + folders with sanitized file names

wolfv: I don't think we should use that function as the docs recommend: `enclosed_name`.

> This will read well-formed ZIP files correctly, and is resistant to path-based exploits. It is recommended over [ZipFile::mangled_name](https://docs.rs/zip/latest/zip/read/struct.ZipFile.html#method.mangled_name).

Why does mangled name fix anything?

baszalmstra:
It is still unclear to me what this solves? can you maybe explain the exact error this solves? You only link to an issue listing and the recipe you reference fails in a cargo invocation, not in rattler-build.

src/source/extract.rs
zelosleone: enclosed_name causes windows file lock issues, which seems to give me problematic results on windows. I would love your input on that, retrying is an approach but i am not sure about it for zip extraction

PR #1801, opened 2025-07-25
feat: enhance prefix detection to support mixed backslash and forward slash formats on Windows

### LINKED ISSUE #1788: Rewriting POSIX prefixes in JSON files on Windows?
As always, thanks for `rattler-build`!

I'm trying to update [ipykernel-feedstock](https://github.com/conda-forge/ipykernel-feedstock/pull/187) to use `rattler-build`.

On Windows, I'm seeing a bad path coming from a build-time JSON file, deployed into `$PREFIX/share`, which also references `$PREFIX`, but fully resolved. To be encoded in JSON, the windows-style slashes get escaped.

Our old technique of replacing the peril-fraught `\` (or, worse, `\\`) with `/` to avoid JSON issues is not working.

Without digging into the code, this looks like perhaps either or both of:
- the POSIX-style `$PREFIX` (with `/`) is not sought
- the Redmond-style `%PREFIX%` (with `\\`) is not sought

Relevant [log](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1292898&view=logs&j=fe7faf59-69bb-5487-b8fd-a35a7bda3a51&t=55731638-dd49-5144-2129-5101b9dd1960&l=1309)

```
 │ │ Rewriting kernelspec at:
 │ │ 	%PREFIX%\share\jupyter\kernels\python3\kernel.json
 │ │ {
 │ │  "argv": [
 │ │   "D:\\bld\\bld\\rattler-build_ipykernel_1753108933\\h_env\\python.exe",
 │ │   "-m",
 │ │   "ipykernel_launcher",
 │ │   "-f",
 │ │   "{connection_file}"
 │ │  ],
 │ │  "display_name": "Python 3 (ipykernel)",
 │ │  "language": "python",
 │ │  "metadata": {
 │ │   "debugger": true
 │ │  }
 │ │ }
 │ │ Kernel python was:
 │ │ 	%PREFIX%\python.exe
 │ │ ... rewriting kernel python with:
 │ │ 	%PREFIX%/python.exe
 │ │ {
 │ │   "argv": [
 │ │     "%PREFIX%/python.exe",
 │ │     "-m",
 │ │     "ipykernel_launcher",
 │ │     "-f",
 │ │     "{connection_file}"
 │ │   ],
 │ │   "display_name": "Python 3 (ipykernel)",
 │ │   "language": "python",
 │ │   "metadata": {
 │ │     "debugger": true
 │ │   }
 │ │ }

# later...

 │ Checking Kernelspec at:     %PREFIX%\share\jupyter\kernels\python3\kernel.json...
 │ {
 │   "argv": [
 │     "D:/bld/bld/rattler-build_ipykernel_1753108933/h_env/python.exe",
 │     "-m",
 │     "ipykernel_launcher",
 │     "-f",
 │     "{connection_file}"
 │   ],
 │   "display_name": "Python 3 (ipykernel)",
 │   "language": "python",
 │   "metadata": {
 │     "debugger": true
 │   }
 │ }
```

zelosleone:
this will close #1788

using vector string type we can get all of the variants and can convert them to system-native approach easily, i think this a clean approach


zelosleone:
test failures are fixed with: https://github.com/prefix-dev/rattler-build/pull/1802 due to pixi update

src/packaging/metadata.rs
```diff
@@ -58,42 +58,46 @@ pub fn contains_prefix_binary(file_path: &Path, prefix: &Path) -> Result<bool, P
...
+    let mut found_variations = Vec::new();
 
-    if target_platform.is_windows() {
-        use crate::utils::to_forward_slash_lossy;
-        // absolute and unc paths will break but it,
-        // will break either way as C:/ can't be converted
-        // to something meaningful in unix either way
-        let forward_slash: Cow<'_, str> = to_forward_slash_lossy(prefix);
+    {
+        // Open the file in a new scope to ensure the memory map and file handle are dropped
```
Hofer-Julian: I don't get that, right after the end of the scope, the scope of the function ends as well.

src/packaging/metadata.rs
```diff
@@ -58,42 +58,46 @@ pub fn contains_prefix_binary(file_path: &Path, prefix: &Path) -> Result<bool, P
...
             .next()
             .is_some()
         {
-            return Ok(Some(forward_slash.to_string()));
+            found_variations.push(prefix_string);
+        }
+
+        if target_platform.is_windows() {
+            use crate::utils::to_forward_slash_lossy;
+            // absolute and unc paths will break but it,
```
Hofer-Julian:
```suggestion
            // absolute and unc paths will break it, but
```

src/packaging/metadata.rs
```diff
@@ -148,11 +152,36 @@ pub fn create_prefix_placeholder(
...
+
+                // If multiple variations found, normalize them to the primary format
+                if variations.len() > 1 {
+                    let content = fs::read_to_string(file_path)?;
+                    let normalized = variations
+                        .iter()
+                        .filter(|&v| v != primary_format)
+                        .fold(content, |acc, variant| acc.replace(variant, primary_format));
+                    fs::write(file_path, normalized)?;
+                }
```
Hofer-Julian: Did you try if `https://docs.rs/dunce/latest/dunce/fn.canonicalize.html` would work as well?

Hofer-Julian (commented):
Looks good. I had a few comments, please check them out :)

src/packaging/metadata.rs
zelosleone: This is because of a test we have in rattler-build where we are trying to delete a file with retries while its being used. Without properly ensuring we open the file in new scope and file handle, its being locked and test cannot pass on windows.

Hofer-Julian: Can you please try to run that test again without that additional scope?
Maybe it was something else.

src/packaging/metadata.rs
zelosleone: Just tried it, it doesn't work with our current test (not in this PR)

where:             

```
elif path == "force_binary/file_without_prefix":
                check_path(p, None)
            elif path == "ignore/file_with_prefix":
                check_path(p, None)
            elif path == "ignore/text_with_prefix":
                check_path(p, None)
            elif path == "is_text/file_with_forwardslash_prefix":
>               assert "\\" not in p["prefix_placeholder"]
E               AssertionError: assert '\\' not in 'C:\\Users\\...02019\\h_env'
E                 '\\' is contained here:
E                   C:\Users\deniz\AppData\Local\Temp\pytest-of-deniz\pytest-37\test_prefix_detection0\bld\rattler-build_prefix_detection_1753802019\h_env
E                 ?   +
```

So we are getting an error where dunce forces everything to platform-specific slashes, without being context aware (where the recipe/script requires forward slashes but dunce canonicalize will make them backslashes)

wolfv:
I think this should be redesigned.

We should first compute the variations (forward / backward / maybe more) of strings, and store them in a `Vec<Vec<String>>`.

Then our other function should search for all of them (mmaping the file only once, ideally), and return which ones were matched.

If we match more than one, we should issue a warning "MixedPrefixPathVariations in file ...".

This current code has not become more readable IMO :)

wolfv:
I kind of re-did the fix here #1836 with a slightly simpler approach. I was first running the failing recipe and checking whether it has the issue. We did properly search for the forward-slash variant but only when the `target_platform` was `windows` (which is not the case when we are building a `noarch` package).

PR #1802, opened 2025-07-25
fix: ensure package name is defined in the pixi.toml

zelosleone:
I have no idea when did this got broken, assuming its because of the new pixi update


baszalmstra:
Yep although the name is now defined twice, you can also remove the name from the workspace. It will take the name from the package.

zelosleone:
> Yep although the name is now defined twice, you can also remove the name from the workspace. It will take the name from the package.

Alright removed it from the workspace

PR #1813, opened 2025-08-05
fix: allow preserving the working directory

baszalmstra:
I noticed that by default rattler-build always deletes the working directory. However, in pixi-build we want to retain it. This PR allows overwriting the default behavior. I think this aligns more with the [future work comment from the PR that introduced this feature](https://github.com/prefix-dev/rattler-build/pull/1119#issuecomment-2410474855).

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -14,6 +14,15 @@ use crate::{
     tool_configuration,
 };
 
+/// Behavior for handling the working directory during the build process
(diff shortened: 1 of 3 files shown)


wolfv:
@baszalmstra I refactored to use an enum instead of a "naked boolean". what do you think?

wolfv (approved):

Issue #1812, opened 2025-08-05
Replece `-` with `_` in Python import tests


apcamargo:
Recipes generated by `rattler-build` keep `-` characters in the import tests. These should be replaced with `_`.

`rattler-build` output for the `file-read-backwards` library:

```yaml
tests:
  - python:
      imports:
        - file-read-backwards
```

Recipe generated with `greyskull`:

```yaml
test:
  imports:
    - file_read_backwards
```

PR #1814, opened 2025-08-06
fix: sanitize hyphenated package names in python test imports

zelosleone:
closes https://github.com/prefix-dev/rattler-build/issues/1812

also tested as an example with `cargo run -- generate-recipe pypi file-read-backwards`

as the result:
```
 Generating recipe for file-read-backwards

  tests:
  - python:
     imports:
      - file_read_backwards
```

i assume replacing all `-` with `_` will be fine too.

### DIFF
--- a/src/recipe_generator/pypi.rs
+++ b/src/recipe_generator/pypi.rs
@@ -430,7 +430,7 @@ pub async fn create_recipe(
 
     recipe.tests.push(Test::Python(PythonTest {
         python: PythonTestInner {
-            imports: vec![metadata.info.name.clone()],
+            imports: vec![metadata.info.name.replace('-', "_")],
             pip_check: true,
         },
     }));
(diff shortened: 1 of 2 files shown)


src/source/patch.rs
```diff
@@ -150,6 +150,7 @@ fn write_patch_content(content: &[u8], path: &Path) -> Result<(), SourceError> {
             #[cfg(not(unix))]
             {
                 // Assume this means windows
+                #[allow(clippy::permissions_set_readonly_false)]
```
wolfv: Can this go to a different PR? This seems quite unrelated?

src/recipe_generator/pypi.rs
```diff
@@ -520,4 +520,32 @@ mod tests {
 
         assert_yaml_snapshot!(recipe);
     }
+
+    #[tokio::test]
+    async fn test_hyphenated_imports_are_sanitized() {
+        let opts = PyPIOpts {
+            package: "file-read-backwards".into(),
+            version: None,
```
wolfv: Can you specify a version so that the test won't break in the future?

wolfv:
Looks good! 2 small comments :)

wolfv (approved):

Issue #1810, opened 2025-08-01
Malformed entity: cannot parse PE header signature


ocroquette:
While repackaging some internal stuff for integration in pixi, I got the following fatal error from rattler-build:

```
│ ╭─ Packaging new files
│ │ Copying done!
│ │
│ ╰─────────────────── (took 5 minutes)
│
╰─────────────────── (took 10 minutes)
Error:   x failed to parse dynamic file: Malformed entity: cannot parse PE header signature (offset 0x10000)
  `-> Malformed entity: cannot parse PE header signature (offset 0x10000)
```

Using '-vvv' does not provide more details.

I can provide in private the data to reproduce the issue, if required.

wolfv:
I am not sure what exactly is hapenning. But since we parse the PE header _only_ to give the user some extra information & warnings it should be safe to ignore a failure in parsing and maybe emit a warning ("Could not parse file ..."), but not let `rattler-build` crash at this point.

PR #1818, opened 2025-08-07
fix: skip malformed PE files during relinking

zelosleone:
closes https://github.com/prefix-dev/rattler-build/issues/1810

as i mentioned, the issue is goblin crate cannot really deal with malformed DLLs and even so, it still tries to parse them. The only logical fix i found is to parse and validate them before relinking stage (during build stage) then we can skip them in the relinking stage and give a warning that package contains malformed PE files so users can focus on fixing the package at least.

I could also use harcoded unicodes for PE files to determine the malformed files, instead of memap (that's what we did before, i think this is also why we missed it) but i think this will cover a large area we need to cover instead

### DIFF
--- a/src/post_process/relink.rs
+++ b/src/post_process/relink.rs
@@ -104,30 +104,23 @@ pub trait Relinker {
     ) -> Result<(), RelinkError>;
 }
 
-/// Returns true if the file is valid (i.e. ELF or Mach-o or PE)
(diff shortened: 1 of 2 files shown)


wolfv:
This looks OK. I think I am not the biggest fan of parsing the whole thing, and then throwing the result away. 

I would prefer that we just make the other function fallible and then ignore the error? Is that not possible?

PR #1826, opened 2025-08-13
feat: parallelize post-processing with rayon

wolfv:
cc @lucascolley maybe you can give this a go and see if it changes anything :)

### DIFF
--- a/src/packaging/metadata.rs
+++ b/src/packaging/metadata.rs
@@ -21,6 +21,7 @@ use rattler_conda_types::{
     },
 };
 use rattler_digest::{compute_bytes_digest, compute_file_digest};
+use rayon::prelude::*;
 use url::Url;
 
 use super::{PackagingError, TempFiles};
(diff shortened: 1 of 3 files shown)


src/packaging.rs
```diff
@@ -337,29 +343,38 @@ where
...
-        let case_folded = normalize_path_for_comparison(path, true).unwrap_or_else(|err| {
-            tracing::warn!(
-                "Failed to normalize path for comparison: {}: {}",
-                path.display(),
-                err
-            );
-            path.display().to_string().to_lowercase()
-        });
+    // Parallel normalization
+    let normalized_pairs: Vec<(String, PathBuf)> = paths_vec
```
baszalmstra: I highly doubt this will have a significant impact on performance. The function is tiny!

Please take a look if it makes sense to parallelize these things!

wolfv: I am sure we should remove a few of those. I mainly wanted to parallelize the relink functions so that they go faster!

Issue #1827, opened 2025-08-13
allow disable-sharded in rattler-build


yjeand:
## Context

I am attempting to use rattler-build behind a corporate proxy, using an artifactory mirror of prefix.dev conda-forge repository. This mirror has been used successfuly with pixi, but it only works if we disable sharded in our pixi `config.toml`.

It seems that rattler-build does not take `disable-sharded` into account. This results in a request being sent to `https://shards.prefix.dev/...` which is unaccessible for me.

## What I did

- generate a minimal `recipe.yaml` for an empty Python module
- prepare a `rattler-config.toml` containing only the following:
```toml
[repodata-config]
disable-sharded = true
```
- attempt to build the project with a custom artifactory channel: `rattler-build build --config-file rattler-config.toml -c https://my-internal-artifactory-channel`

This results in the following error:
```
 ╭─ Finding outputs from recipe
 │ Found 1 variants
 │ 
 │ Build variant: test-1.1.0-hb0f4dca_1
 │ 
 │ ╭─────────────────┬────────────╮
 │ │ Variant         ┆ Version    │
 │ ╞═════════════════╪════════════╡
 │ │ target_platform ┆ "linux-64" │
 │ ╰─────────────────┴────────────╯
 │
 ╰─────────────────── (took 0 seconds)

 ╭─ Running build for recipe: test-1.1.0-hb0f4dca_1
 │
 │ ╭─ Fetching source code
 │ │ Fetching source from path: /some/local/path/src
 │ │ Copied 1 files into isolated environment
 │ │
 │ ╰─────────────────── (took 0 seconds)
 │
 │ ╭─ Resolving environments
 │ │ 
 │ │ Resolving host environment:
 │ │   Platform: linux-64 [__unix=0=0, __linux=5.15.0=0, __glibc=2.35=0, __archspec=1=skylake]
 │ │   Channels: 
 │ │    - file:////some/local/path/output/
 │ │    - https://my-internal-artifactory-channel/
 │ │   Specs:
 │ │    - python
 │ │    - pip
[...]

PR #1830, opened 2025-08-15
feat: add support for disabling sharded repodata

zelosleone:
closes https://github.com/prefix-dev/rattler-build/issues/1827


wolfv:
Looks like a good start but I believe in `pixi` we can set the configuration per-channel as well. We should match that exactly.

wolfv:
Also note that there are multiple options: https://pixi.sh/latest/reference/pixi_configuration/#repodata-config

PR #1831, opened 2025-08-15
feat: add `use-sharded` and `use-jlap` flags

wolfv:


src/opt.rs
```diff
@@ -203,6 +203,14 @@ pub struct CommonOpts {
     #[clap(long, env = "RATTLER_BZ2", default_value = "true", hide = true)]
     pub use_bz2: bool,
 
+    /// Enable support for sharded repodata
+    #[clap(long, env = "RATTLER_SHARDED", default_value = "true", hide = true)]
+    pub use_sharded: bool,
```
zelosleone, line 208: Are we sure we want to make it true for default on sharded repodata? I think false would be better

zelosleone: never mind, pixi also by default uses true just checked

zelosleone: nevermind seems like pixi according to documentation also uses true by default

zelosleone: https://pixi.sh/latest/reference/pixi_configuration/#repodata-config
yeah it seems like disable sharded is by default true, so i don't think we should use sharded by default?

Issue #1837, opened 2025-08-19
Files are copied from .git


noamgot:
Starting from v0.42.1, I noticed rattler-build tries to copy files from the `.git` folder, even though according to the docs they are not supposed to be copied (https://rattler.build/latest/internals/#general-file-filtering)

Here's a minimal reproducible example (using a small pixi project):

**pixi.toml**
```toml
[workspace]
channels = ["conda-forge"]
name = "rattler-build-bugs"
platforms = ["linux-64"]
version = "1.2.3"

[environments]
build = { features = ["build"], no-default-feature = true }

[dependencies]
numpy = ">=2.3.2,<3"
pandas = ">=2.3.1,<3"
requests = ">=2.32.4,<3"


## Build Feature
[feature.build.dependencies]
rattler-build = "0.42.*"

[feature.build.tasks]
_build-mkdir = "mkdir -p ./__localbuild__/conda/"

[feature.build.tasks.build]
cmd = "rattler-build build --package-format tar-bz2 -r recipe.yml --output-dir ./__localbuild__/conda/ --no-test"
depends-on = ["_build-mkdir"]
description = "build the conda package"
```

The recipe.yaml that I used:
```yml
context: {}

package:
  name: rattler-build-bugs
  version: 1.2.3

source:
  path: .

build:
  script: python -m pip install --no-deps --ignore-installed -vv .
  noarch: python

requirements:
  host:
  - python =3.11
  - pip
  - hatch

  run:
  - numpy >=2.3.2,<3
  - pandas >=2.3.1,<3
  - requests >=2.32.4,<3

about:
  homepage: https://www.example.com
  summary: amazing summary
```

To reproduce you should:
1. create a git repo with some commits
2. run `pixi run build -vv`  - you should see many messages of this form:
```
2025-08-19T15:32:28.703505Z DEBUG rattler_build::source::copy_dir: Failed to copy metadata for ".../rattler-build-bugs/__localbuild__/conda/bld/rattler-build_rattler-build-bugs_1755617547/work/.git/objects/60/1c93b4c6c02c051b8d4736934fa93f4dcac8b8" Os { code: 13, kind: PermissionDenied, message: "Permission denied" }
```
3. downgrade rattler-build to `0.41.*` and see that these logs are gone

noamgot:
@wolfv correct, in my case there's no need to copy the .git files whatsoever.
[...]

PR #1838, opened 2025-08-19
fix: fix copying .git folders by using overrides

zelosleone:
this will close https://github.com/prefix-dev/rattler-build/issues/1837

basically, https://github.com/prefix-dev/rattler-build/commit/19669322d00bacee9e067b4ef09236c710e023bc this commit of mine disabled .ignore files, however it introduced something else that we missed even before that. we didnt really specially ignored .git folders at all during source copying. we relied on .ignore file for that it seems like

now override https://docs.rs/ignore/latest/ignore/overrides/ is very similar to `--include` and  `--exclude` and its a cleaner approach, as it will also allow us to make additions easily in the future

also made sure git folders are included in fetched git sources, while excluded from path sources for source copying


remimimimimi (approved):
Looks good, thanks!

PR #1845, opened 2025-08-22
fix: patch files that use `.orig` or `.bak`

wolfv:
Currently, patch files where the original file has a file ending with `.orig` or `.bak` in them don't work correctly. 

There are two reasons:

- when trying to find the `strip-level` we were looking for _all_ paths matching. This doesn't work properly because the `.orig` file does not actually exist on disk.
- We didn't strip the `.orig` or `.bak` before applying the patch.

Should be fixed now!


src/source/patch.rs
```diff
@@ -77,11 +77,14 @@ fn guess_strip_level(patch: &Patch<[u8]>, work_dir: &Path) -> Result<usize, Sour
...
-        let all_paths_exist = patched_files.iter().all(|p| {
+        // We check for _any_ path existing here. Sometimes patches reference
+        // files that are deleted (e.g. `foobar.orig`) and thus it's more robust to look for
+        // the first match in the affected files.
+        let any_paths_exist = patched_files.iter().any(|p| {
             let path: PathBuf = p.components().skip(strip_level).collect();
             work_dir.join(path).exists()
         });
-        if all_paths_exist {
+        if any_paths_exist {
```
remimimimimi, line 125: I think it would be more robust to check that all paths exist, but also handle file endings here.

But I'm fine with this solution as well, as it is much simpler.

remimimimimi (approved):
Just one small comment, otherwise good!

PR #1853, opened 2025-08-25
fix: include symlink even outside

### LINKED ISSUE #1852: Support for packaging external symlinks via recipe configuration
## Problem Description

Currently, `rattler-build` automatically excludes symlinks that point to files outside the package prefix from `paths.json`, even when explicitly included via `always_include_files`. This behavior is hardcoded in the metadata generation phase and cannot be overridden through recipe configuration.

## Current Behavior

When a symlink points to a target outside the package:

1. **Relative symlinks**: If the resolved target doesn't exist, the symlink is skipped with a warning
2. **Absolute symlinks**: A warning is logged, but the behavior is inconsistent
3. **`always_include_files`**: Has no effect on this limitation as the filtering happens after file discovery

```rust
// In src/packaging/metadata.rs
if !resolved_path.exists() {
    tracing::warn!(
        "symlink target not part of this package: {:?} -> {:?}",
        &p,
        &link_target
    );
    continue; // Always skips, ignoring recipe configuration
}
```

## Impact on Real Packages

This limitation significantly affects packages like **PyTorch** and other ML frameworks that commonly use symlinks for `libtorch_cuda.so`

These packages often need to include symlinks that point to system libraries or external dependencies that are expected to be present at runtime.

## Proposed Solution

Add recipe-level configuration options to control symlink packaging behavior:

```yaml
build:
  # Allow packaging symlinks that point outside the prefix
  allow_external_symlinks: true
  
  # Or more granular control
  symlink_policy:
    external_symlinks: "allow"  # allow, warn, error
    broken_symlinks: "warn"     # allow, warn, error
    
  # Force include specific symlinks regardless of target
  force_include_symlinks:
    - "lib/libcuda.so"
    - "bin/python"
```

## Alternative Workarounds

Currently, package maintainers must:

1. **Copy files instead of symlinking**: Increases package size
2. **Create relative symlinks**: Requires restructuring package layout
3. **Modify build scripts**: Complex logic to handle different scenarios

## Expected Behavior

With the proposed configuration:

- Recipe authors can explicitly opt-in to packaging external symlinks
- `always_include_files` should respect symlink inclusion decisions
- Clear warnings/errors when symlinks cannot be resolved
- Backward compatibility with current default behavior

This would enable proper packaging of complex software like PyTorch while maintaining safety for packages that don't need this functionality.

fecet:
close #1852

### DIFF
--- a/src/packaging/metadata.rs
+++ b/src/packaging/metadata.rs
@@ -428,7 +428,7 @@ impl Output {
                                         &p,
                                         &link_target
                                     );
-                                    return Ok(None);
+                                    // Continue processing the symlink instead of skipping it
                                 }
                             } else {
                                 tracing::warn!(
--- a/src/packaging/metadata.rs
+++ b/src/packaging/metadata.rs
@@ -468,17 +468,24 @@ impl Output {
                         &content_type,
                         self.recipe.build().prefix_detection(),
                     )?;
-                    let digest = compute_file_digest::<sha2::Sha256>(p)?;
+                    let file_size = meta.len();
+                    // Compute SHA256 for files - empty files get empty hash
+                    let digest = if file_size > 0 {
+                        Some(compute_file_digest::<sha2::Sha256>(p)?)
+                    } else {
+                        Some(compute_bytes_digest::<sha2::Sha256>(&[]))
+                    };
                     let no_link = always_copy_files.is_match(&relative_path);
                     return Ok(Some(PathsEntry {
-                        sha256: Some(digest),
+                        sha256: digest,
                         relative_path,
                         path_type: PathType::HardLink,
                         prefix_placeholder,
                         no_link,
-                        size_in_bytes: Some(meta.len()),
+                        size_in_bytes: Some(file_size),
                     }));
                 } else if meta.is_symlink() {
+                    // For symlinks, compute hash of the target file content if it exists and is within package, otherwise empty digest
                     let digest = if p.is_file() {
                         compute_file_digest::<sha2::Sha256>(p)?
                     } else {
--- a/src/packaging/metadata.rs
+++ b/src/packaging/metadata.rs
@@ -27,6 +27,16 @@ use url::Url;
 use super::{PackagingError, TempFiles};
 use crate::{hash::HashInput, metadata::Output, recipe::parser::PrefixDetection};
 
+/// Safely check if a symlink resolves to a regular file, with basic loop protection
+fn is_symlink_to_file(path: &Path) -> bool {
+    // Simple approach: try to canonicalize the path, which handles cycles gracefully
+    // If canonicalization fails (due to cycles or missing targets), assume it's not a file
+    match path.canonicalize() {
+        Ok(canonical_path) => canonical_path.is_file(),
+        Err(_) => false, // Could be cycle, missing target, or permission issue
+    }
+}
+
 /// Detect if the file contains the prefix in binary mode.
 #[allow(unused_variables)]
 pub fn contains_prefix_binary(file_path: &Path, prefix: &Path) -> Result<bool, PackagingError> {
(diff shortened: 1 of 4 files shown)


wolfv:
I am a little bit unsure what we should do with the SHA256 hash in this case. If it points outside the prefix / or not to a file coming from the package, it doesn't really make sense to compute it at all. I think for broken symlinks we use the empty SHA256 string instead. Maybe that would make sense here as well? WDYT?

zelosleone:
> Quick question about the broken symlink SHA256 handling - I see that PathsEntry.sha256 is `Option<Hash>`, not `Option<String>` https://github.com/conda/rattler/blob/b3eff3438a28e39923c011414dff17320f4f8132/crates/rattler_conda_types/src/package/paths.rs#L219.
> 
> When you said "use the empty SHA256 string", did you mean:
> 
>     * None (omits the field in JSON)
> 
>     * Or the hash of empty bytes (e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855)?
> 
> 
> Can't directly set it to "" without changing the type. Which approach do you prefer?

Hash of empty bytes should be fine in my opinion. By the way what i am worried is infinite looping, maybe setting a limitation or some rules to symlink lookup would be better?

src/packaging/metadata.rs
```diff
@@ -468,17 +468,24 @@ impl Output {
...
                         self.recipe.build().prefix_detection(),
                     )?;
-                    let digest = compute_file_digest::<sha2::Sha256>(p)?;
+                    let file_size = meta.len();
+                    // Compute SHA256 for files - empty files get empty hash
+                    let digest = if file_size > 0 {
+                        Some(compute_file_digest::<sha2::Sha256>(p)?)
+                    } else {
+                        Some(compute_bytes_digest::<sha2::Sha256>(&[]))
+                    };
```
zelosleone, line 487: I think it would be better to make this configurable rather than directly continuing for all cases? It might be not ideal to do this for all packages, since this is might not be intended behavior. Plus issue mentioned it as being configurable so i am unsure if this is intended.

src/packaging/metadata.rs
```diff
@@ -468,17 +468,24 @@ impl Output {
                         &content_type,
                         self.recipe.build().prefix_detection(),
                     )?;
-                    let digest = compute_file_digest::<sha2::Sha256>(p)?;
+                    let file_size = meta.len();
+                    // Compute SHA256 for files - empty files get empty hash
+                    let digest = if file_size > 0 {
+                        Some(compute_file_digest::<sha2::Sha256>(p)?)
+                    } else {
+                        Some(compute_bytes_digest::<sha2::Sha256>(&[]))
```
zelosleone, line 486: You can also make a new type for empty hash with empty bytes and directly use for example `EmptyHash`

src/packaging/metadata.rs
fecet: This is actually conda-build's default behavior; users can also manually remove those symlinks to get the desired result. As a user, I only need one option; however, the discussion with @wolfv  inspired me to implement this quick fix and submit it along the way :)

zelosleone: Oh if it's conda-build's default behavior then that's completely alright

fecet:
@zelosleone Thanks for the EmptyHash suggestion and the infinite loop concern!

Infinite loop protection: ✅ Implemented using canonicalize() which safely handles symlink cycles without manual recursion tracking.
EmptyHash evaluation: Only 2 calls to compute_bytes_digest::<sha2::Sha256>(&[]) in the entire codebase. Implementation options would add 5-20 lines of complexity for negligible performance gain (hashing empty bytes is already very fast).

Current approach keeps the code simple while addressing the safety concern. Happy to revisit EmptyHash optimization if usage increases in the future. WDYT?

PR #1860, opened 2025-08-28
refactor: source cache into a separate crate

wolfv:
This PR splits up `rattler-build` to add additional crates:

- `rattler_build_source_cache` contains a module to fetch and cache sources (URL and Git).
- `rattler_git` is a copy from `pixi_git` (which comes from `uv_git`) and takes care of checking out git repositories. This should ideally – if things work out – make it's way to `rattler`.


crates/rattler_build_source_cache/Cargo.toml
```diff
@@ -0,0 +1,50 @@
+[package]
+name = "rattler_build_source_cache"
+version = "0.1.0"
+edition.workspace = true
+authors.workspace = true
+license.workspace = true
+repository.workspace = true
+description = "Source cache management for rattler-build"
```
nichmor, line 8: only for GIT and url right?

crates/rattler_build_source_cache/README.md
```diff
@@ -0,0 +1,103 @@
...
+- **Content-Addressable Storage**: Sources are indexed by content hash for deduplication
+- **Concurrent Access Control**: File locking ensures safe concurrent access from multiple processes
+- **Automatic Extraction**: Archives (.tar.gz, .zip, .7z, etc.) are automatically extracted
+- **Cache Management**: Automatic cleanup of old entries, configurable TTL
+- **Progress Tracking**: Optional progress callbacks for downloads and extractions
+- **Checksum Validation**: SHA256 and MD5 checksum support for URL downloads
+
+## Usage
+
+```rust
```
nichmor, line 17: should we make it as a doctest for example in lib.rs?

crates/rattler_build_source_cache/Cargo.toml
wolfv: yeah but we don't cache "path" sources (as they are on disk already) :)

PR #1866, opened 2025-09-02
fix: patch application with `.orig` files

wolfv:


src/source/patch.rs
```diff
@@ -634,7 +827,7 @@ mod tests {
                     (Some(o), None) => {
                         patch_results.push((patch, PatchResult::Deleted(!o.exists())))
                     }
-                    (Some(o), Some(m)) => {
+                    (Some(_), Some(m)) => {
```
wolfv, line 830: This was the problem I believe :)

PR #1872, opened 2025-09-03
feat(bindings): add recipe parsing in python bindings

zelosleone:
will close https://github.com/prefix-dev/rattler-build/issues/1871

part of https://github.com/prefix-dev/rattler-build/issues/1851

### DIFF
--- a/py-rattler-build/tests/unit/test_recipe_oop.py
+++ b/py-rattler-build/tests/unit/test_recipe_oop.py
@@ -29,6 +29,7 @@ def test_recipe_all_sections() -> None:
     assert build.number == 0
     assert not build.noarch
 
+    assert build.script is not None
     assert build.script.strip() == "pip install . --no-deps -vv"
 
     # Requirements (including selectors)


py-rattler-build/src/lib.rs
```diff
@@ -90,6 +92,65 @@ fn generate_luarocks_recipe_string_py(rock: String) -> PyResult<String> {
...
+/// Parse a recipe YAML string and return the parsed recipe as a Python dictionary.
+#[pyfunction]
+#[pyo3(signature = (yaml_content, target_platform=None, host_platform=None, build_platform=None, experimental=None, allow_undefined=None))]
+fn parse_recipe_py(
+    yaml_content: String,
+    target_platform: Option<String>,
+    host_platform: Option<String>,
+    build_platform: Option<String>,
+    experimental: Option<bool>,
+    allow_undefined: Option<bool>,
```
wolfv: I think we are at least missing the variant configuration here. 

However, this should be given as a proper SelectorConfig IMO, so you should make another class for Selector Config that can be created properly from Python.

We can also think about using hte `rattler` classes e.g. for Platform as we could use `py-rattler` (at least in the Python interface because translating them through to Rust might be tricky).

zelosleone: Added a `selector` class for us to pytonize variant parts correctly without defaulting them as well, but yeah this only for parsing since as we talked on discord, this is only for the parsing part after the resolving the if/then/else selectors correctly. So in the future this could be better when we get the code from pixi-buld to rattler-build.

py-rattler-build/src/lib.rs
```diff
@@ -37,6 +45,174 @@ where
...
+        build_platform: Option<String>,
+        variant: Option<HashMap<String, Py<PyAny>>>,
+        experimental: Option<bool>,
+        allow_undefined: Option<bool>,
+        recipe_path: Option<PathBuf>,
+    ) -> PyResult<Self> {
+        let target_platform = target_platform
+            .map(|p| Platform::from_str(&p))
+            .transpose()
+            .map_err(|e| PyRuntimeError::new_err(e.to_string()))?
```
nichmor: I think it is better to pack them in a more concrete error, for example, PyRattlerBuildError and make this into exception rather than a RuntimeError.

You can take a look how we have done it in py-rattler:
https://github.com/conda/rattler/blob/main/py-rattler/src/error.rs#L102

py-rattler-build/src/lib.rs
```diff
@@ -37,6 +45,174 @@ where
...
+
+        let build_platform = build_platform
+            .map(|p| Platform::from_str(&p))
+            .transpose()
+            .map_err(|e| PyRuntimeError::new_err(e.to_string()))?
+            .unwrap_or_else(Platform::current);
+
+        // Convert variant from Python dict to BTreeMap<NormalizedKey, Variable>
+        let variant_map = if let Some(variant_dict) = variant {
+            Python::attach(|py| {
```
nichmor: you can receive py: Python as an argument in any `#[pymethods]` so you don't need anymore to `attach` or request for gil token

py-rattler-build/src/lib.rs
```diff
@@ -37,6 +45,174 @@ where
...
+                let mut map = BTreeMap::new();
+                for (key, value) in variant_dict {
+                    let normalized_key = NormalizedKey::from(key);
+                    // Convert Python object to JSON Value then to Variable
+                    let json_val: serde_json::Value = pythonize::depythonize(value.bind(py))
+                        .map_err(|e| {
+                            PyRuntimeError::new_err(format!(
```
nichmor: same goes for the errors here

py-rattler-build/src/lib.rs
```diff
@@ -37,6 +45,174 @@ where
...
+                                        Variable::from_string(&n.to_string())
+                                    }),
+                                    _ => Err(PyRuntimeError::new_err(
+                                        "Complex array elements not supported",
+                                    )),
+                                })
+                                .collect();
+                            Variable::from(vars?)
+                        }
+                        JsonValue::Object(_) => {
```
nichmor: you can omit other variants by using `_ => ..`

py-rattler-build/src/lib.rs
```diff
@@ -37,6 +45,174 @@ where
...
+            })?
+        } else {
+            BTreeMap::new()
+        };
+
+        let selector_config = SelectorConfig {
+            target_platform,
+            host_platform,
+            build_platform,
+            hash: None,
```
nichmor: Should we also receive `hash` as an argument to the constructor?

py-rattler-build/src/lib.rs
```diff
@@ -37,6 +45,174 @@ where
...
+    fn target_platform(&self) -> String {
+        self.inner.target_platform.to_string()
+    }
+
+    #[getter]
+    fn host_platform(&self) -> String {
+        self.inner.host_platform.to_string()
+    }
+
+    #[getter]
```
nichmor, line 186: I think we also need to provide setters for all these fields

py-rattler-build/src/lib.rs
```diff
@@ -37,6 +45,174 @@ where
...
+        self.inner.allow_undefined
+    }
+
+    #[getter]
+    fn recipe_path(&self) -> Option<PathBuf> {
+        self.inner.recipe_path.clone()
+    }
+
+    #[getter]
+    fn variant(&self) -> PyResult<Py<PyAny>> {
```
nichmor: same here - you can just receive py: Python and omit attach part

py-rattler-build/src/lib.rs
```diff
@@ -37,6 +45,174 @@ where
...
+        Python::attach(|py| {
+            let mut dict = HashMap::new();
+            for (key, value) in &self.inner.variant {
+                let json_value = serde_json::to_value(value).map_err(|e| {
+                    PyRuntimeError::new_err(format!("Failed to serialize variant: {}", e))
+                })?;
+                dict.insert(key.normalize(), json_value);
+            }
+            pythonize::pythonize(py, &dict)
```
nichmor: Why do you return `PyAny` and not `HashMap`?

py-rattler-build/src/lib.rs
```diff
@@ -37,6 +45,174 @@ where
...
+                    PyRuntimeError::new_err(format!("Failed to convert variant to Python: {}", e))
+                })
+        })
+    }
+}
+
+impl PySelectorConfig {
+    /// Internal method to get the underlying SelectorConfig
+    /// This is used internally when passing the config to Rust functions
+    pub(crate) fn inner(&self) -> &SelectorConfig {
```
nichmor: I think you can omit having this function and just mark the inner field directly as pub(crate)

py-rattler-build/src/lib.rs
```diff
@@ -90,6 +266,34 @@ fn generate_luarocks_recipe_string_py(rock: String) -> PyResult<String> {
...
+    yaml_content: String,
+    selector_config: &PySelectorConfig,
+) -> PyResult<Py<PyAny>> {
+    match Recipe::from_yaml(yaml_content.as_str(), selector_config.inner().clone()) {
+        Ok(recipe) => {
+            let json_value = serde_json::to_value(recipe).map_err(|e| {
+                PyRuntimeError::new_err(format!("Failed to serialize recipe: {}", e))
+            })?;
+
+            Python::attach(|py| {
```
nichmor, line 370: same here with `py` token

py-rattler-build/tests/unit/test_recipe_oop.py
```diff
@@ -0,0 +1,130 @@
...
+    # Other component reprs
+    assert "Build(" in repr(recipe.build)
+    assert "Requirements(" in repr(recipe.requirements)
+    assert "About(" in repr(recipe.about)
+    assert "Source(" in repr(recipe.source[0])
+
+
+def test_selector_config_with_variants() -> None:
+    """Test SelectorConfig with variant configuration"""
+    config = PySelectorConfig(target_platform="linux-64", variant={"python": "3.11", "build_number": 1})
```
nichmor: you are testing rust structure, but we need to double wrap it into a python class

nichmor (changes requested):
We also need to double wrap `PySelectorConfig` into a separate python class `SelectorConfig`

py-rattler-build/src/lib.rs
zelosleone: Imo just returning HashMap would require us to convert each type correctly and would mean more code/more complexity. `Variable` in `BTreeMap<NormalizedKey, Variable>` takes `bool`, `string `or `vec<variable>` this why we can just pythonize them correctly as pythonzie just makes them python objects automatically.

py-rattler-build/rattler_build/__init__.py
```diff
@@ -26,6 +39,16 @@
...
+    "Recipe",
+    "Package",
+    "Build",
+    "Requirements",
+    "About",
+    "Source",
+    "TestType",
+    "parse_recipe_py",
+    "SelectorConfig",
+    "PySelectorConfig",
```
nichmor: we don't need to export Py_ thingy. end-users should use python version of our structs

py-rattler-build/rattler_build/recipe.py
```diff
@@ -0,0 +1,489 @@
...
+from pathlib import Path
+from typing import Any, Dict, List, Optional, Union
+from .rattler_build import parse_recipe_py, PySelectorConfig
+
+
+class Build:
+    """Build configuration for a recipe."""
+
+    def __init__(self, data: Dict[str, Any]):
+        self._data = data
```
nichmor, line 28: I would suggest to also put this field in the class itself.
like:

```python
class Build:
    """Build configuration for a recipe."""
        _data: Dict[str, Any]
```

py-rattler-build/rattler_build/recipe.py
```diff
@@ -0,0 +1,489 @@
...
+        if isinstance(noarch_data, dict):
+            return noarch_data.get("type", "generic")
+        return "generic"
+
+    @property
+    def noarch_type(self) -> Optional[str]:
+        """Get the noarch type if any (alias for noarch property)."""
+        return self.noarch
+
+    def is_noarch(self) -> bool:
```
nichmor, line 58: you can also make them properties

py-rattler-build/rattler_build/recipe.py
```diff
@@ -0,0 +1,489 @@
...
+    """Test type for a recipe."""
+
+    def __init__(self, data: Dict[str, Any]):
+        self._data = data
+
+    @property
+    def test_type(self) -> str:
+        """Get the test type name."""
+        if "Command" in self._data:
+            return "command"
```
nichmor: I think we can maybe create an enum type for this values?

py-rattler-build/rattler_build/recipe.py
```diff
@@ -0,0 +1,489 @@
...
+    def __init__(
+        self,
+        target_platform: Optional[str] = None,
+        host_platform: Optional[str] = None,
+        build_platform: Optional[str] = None,
+        experimental: Optional[bool] = None,
+        allow_undefined: Optional[bool] = None,
+        variant: Optional[Dict[str, Any]] = None,
+    ):
+        self._config = PySelectorConfig(
```
nichmor, line 337: sames goes here, lets put self._config also as a explicit class field

py-rattler-build/rattler_build/recipe.py
```diff
@@ -0,0 +1,489 @@
...
+        """Get the build platform."""
+        return self._config.build_platform
+
+    @property
+    def experimental(self) -> Optional[bool]:
+        """Get whether experimental features are enabled."""
+        return self._config.experimental
+
+    @property
+    def allow_undefined(self) -> Optional[bool]:
```
nichmor, line 387: we are also missing setters for these properties. Let's also add them in the tests

py-rattler-build/src/error.rs
```diff
@@ -0,0 +1,70 @@
+use pyo3::exceptions::PyRuntimeError;
+use pyo3::prelude::*;
+use std::fmt;
+
+pyo3::create_exception!(rattler_build, PyRattlerBuildError, PyRuntimeError);
+
+#[derive(Debug)]
+pub enum RattlerBuildError {
+    PlatformParse(String),
```
nichmor: You need to use `thiserror #[from]` to create the error directly

and also register exceptions for every error, as we do in py-rattler

py-rattler-build/rattler_build/recipe.py
```diff
@@ -0,0 +1,553 @@
...
+        """Get the build number."""
+        return self._data.get("number", 0)
+
+    @property
+    def string(self) -> Optional[str]:
+        """Get the build string."""
+        return self._data.get("string")
+
+    @property
+    def script(self) -> Optional[str]:
```
nichmor: we have duplicated property here

py-rattler-build/src/lib.rs
```diff
@@ -149,33 +446,34 @@ fn build_recipes_py(
...
     let host_platform = host_platform
         .map(|p| Platform::from_str(&p))
         .transpose()
-        .map_err(|e| PyRuntimeError::new_err(e.to_string()))?;
+        .map_err(RattlerBuildError::from)?;
     let package_format = package_format
         .map(|p| PackageFormatAndCompression::from_str(&p))
         .transpose()
-        .map_err(|e| PyRuntimeError::new_err(e.to_string()))?;
+        .map_err(|e| RattlerBuildError::PackageFormat(e.to_string()))?;
```
nichmor, line 461: can we use directly error here?

py-rattler-build/src/lib.rs
```diff
@@ -149,33 +446,34 @@ fn build_recipes_py(
...
+        .map_err(RattlerBuildError::from)?;
     let channel = match channel {
         None => None,
         Some(channel) => Some(
             channel
                 .iter()
                 .map(|c| {
                     NamedChannelOrUrl::from_str(c)
-                        .map_err(|e| PyRuntimeError::new_err(e.to_string()))
+                        .map_err(|e| RattlerBuildError::ChannelPriority(e.to_string()))
```
nichmor, line 475: same here

py-rattler-build/src/lib.rs
```diff
@@ -311,9 +610,9 @@ fn upload_package_to_artifactory_py(
     auth_file: Option<PathBuf>,
 ) -> PyResult<()> {
     let store = tool_configuration::get_auth_store(auth_file)
-        .map_err(|e| PyRuntimeError::new_err(e.to_string()))?;
+        .map_err(|e| RattlerBuildError::Auth(e.to_string()))?;
```
nichmor: same here

nichmor (changes requested):

py-rattler-build/rattler_build/recipe.py
```diff
@@ -0,0 +1,553 @@
...
+
+    @property
+    def is_noarch(self) -> bool:
+        """Check if this is a noarch build."""
+        return self.noarch is not None
+
+    @property
+    def has_script(self) -> bool:
+        """Check if this build has a script defined."""
+        return self.script is not None
```
nichmor, line 70: after some testing it seems to lack this field. should we read it from _data or set it in the struct? @zelosleone

nichmor (changes requested):

py-rattler-build/rattler_build/recipe.py
zelosleone: i think reading from _data is gonna be better, i dont think we need a new struct for this and it would require more complex work that will eventually go beyond of this pr, maybe in another pr we can go for that? lets discuss this further, for now i pushed a new commit for property

nichmor (approved):

Issue #1880, opened 2025-09-05
do not download packages when only rendering w/ running the solver


beckermr:
rattler-build is downloading all of the packages for a given environment when it is only rendering a recipe and checking if the envs are solvable. The actual packages should not be needed and these downloads should be skipped.

PR #1883, opened 2025-09-10
fix: disable run_exports extraction in render-only solves

zelosleone:
closes https://github.com/prefix-dev/rattler-build/issues/1880

i also think we shouldnt download these packages while running `--render-only --with-solve` as it makes sense and our current behavior would be too confusing. as to provide solvability we dont need to download the packages, we only need the index metadata anyways.

the only change was done to rattler-build build for `--render-only --with-solve` flag.

### DIFF
--- a/src/render/resolved_dependencies.rs
+++ b/src/render/resolved_dependencies.rs
@@ -693,6 +693,7 @@ pub(crate) async fn resolve_dependencies(
     output: &Output,
     channels: &[ChannelUrl],
     tool_configuration: &tool_configuration::Configuration,
+    download_missing_run_exports: bool,
 ) -> Result<FinalizedDependencies, ResolveError> {
     let merge_build_host = output.recipe.build().merge_build_and_host_envs();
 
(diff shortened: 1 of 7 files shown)


src/render/resolved_dependencies.rs
```diff
@@ -693,6 +693,7 @@ pub(crate) async fn resolve_dependencies(
     output: &Output,
     channels: &[ChannelUrl],
     tool_configuration: &tool_configuration::Configuration,
+    download_missing_run_exports: bool,
```
remimimimimi: I think it would be better to have a separate enum instead of a bool here.

src/render/solver.rs
```diff
@@ -186,7 +186,7 @@ impl rattler_repodata_gateway::DownloadReporter for GatewayReporter {
             .multi_progress
             .add(ProgressBar::new(1))
             .with_finish(indicatif::ProgressFinish::AndLeave)
-            .with_prefix("Downloading");
+            .with_prefix("Fetching repodata");
```
remimimimimi: I think it is more inconsistent than it was before.

remimimimimi (changes requested):

src/render/solver.rs
zelosleone: Went with `Downloading repodata` Hopefully thats clearer and better

src/render/resolved_dependencies.rs
zelosleone: I had to debate with myself for this, i think a boolean was fine enough too but went with your suggestion for readability and future usage

remimimimimi (approved):

PR #1909, opened 2025-09-30
feat: move variant to rattler

Hofer-Julian:
This moves variant file parsing to a new crate in rattler called `rattler_variants`. Most of it is directly moved from rattler-build.
The only bigger duplication is `VariantValue`.
In `rattler-build` we reused `Variable` that is used for variables in the `recipe.yaml`. Therefore, we have to add conversion functions on the rattler-build side.


wolfv:
Should we close this for now?

PR #1925, opened 2025-10-08
feat: use `pixi-build` to build the Rust - Python bindings

wolfv:
Get's rid of some manual tasks in place of `pixi-build` compiling / linking things for us :)

### DIFF
--- a/py-rattler-build/pixi.toml
+++ b/py-rattler-build/pixi.toml
@@ -6,7 +6,7 @@ channels = ["conda-forge"]
 platforms = ["win-64", "linux-64", "osx-64", "osx-arm64"]
 license = "BSD-3-Clause"
 preview = ["pixi-build"]
-# This does not seem to work yet?
+# We are using the oldest supported python version as the build variant here
 build-variants = { python = ["3.8"] }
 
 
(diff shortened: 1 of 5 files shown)


py-rattler-build/pixi.toml
```diff
@@ -5,22 +5,30 @@ authors = ["Wolf Vollprecht <wolf@prefix.dev>"]
...
+# build-variants = { python = ["3.8"] }
 
-[tasks]
 
-[feature.build.dependencies]
-maturin = "~=1.2.2"
-pip = "~=23.2.1"
-rust = ">=1.86.0,<1.87"
+[package]
+name = "py-rattler-build"
```
ruben-arts:
```suggestion
```
We read the metadata from the `pyproject.toml`
I would get rid of the `dynamic = ["version"]` so you don't have to have multiple location for this value.

py-rattler-build/pixi.toml
```diff
@@ -5,22 +5,30 @@ authors = ["Wolf Vollprecht <wolf@prefix.dev>"]
 channels = ["conda-forge"]
 platforms = ["win-64", "linux-64", "osx-64", "osx-arm64"]
 license = "BSD-3-Clause"
+preview = ["pixi-build"]
+# This does not seem to work yet?
+# build-variants = { python = ["3.8"] }
```
ruben-arts: Best to ask @Hofer-Julian

Hofer-Julian: This should work

py-rattler-build/pixi.toml
wolfv: yeah currently the version is read from the `Cargo.toml`. But since we are using `tbump` it doesn't really matter in how many places we keep the static version around :)

PR #1933, opened 2025-10-11
feat: new parser

wolfv:
This adds a new parser that has some advantages. 

It splits into a "stage0" and "stage1" parser.

Stage 0: completely un-evaluated recipe that still contains If/Then/Else and Jinja. This is heavily inspired by the "intermediate recipe" work in pixi-build backends and hopefully replaces it. 
Stage 1: All Jinja and conditionals resolved.

TODO:

- [ ] move globvec to another crate (e.g. rattler_build_types).

### DIFF
--- a/crates/rattler_build_recipe/src/stage0/parser/snapshot_tests.rs
+++ b/crates/rattler_build_recipe/src/stage0/parser/snapshot_tests.rs
@@ -4,10 +4,10 @@
 
 use crate::stage0::parser::{parse_recipe_from_source, parse_recipe_or_multi_from_source};
 
-const TEST_DATA_DIR: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/test-data");
-
+#[cfg(not(target_os = "windows"))]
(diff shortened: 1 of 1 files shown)


crates/rattler_build_recipe/src/stage0/evaluate.rs
```diff
@@ -0,0 +1,1611 @@
...
+    context: &EvaluationContext,
+) -> Result<String, ParseError> {
+    match content {
+        ScriptContent::Command(cmd) => Ok(cmd.clone()),
+        ScriptContent::Inline(inline) => {
+            // For inline scripts, we need to evaluate the content or file
+            if let Some(content_list) = &inline.content {
+                // Evaluate the conditional list to get a Vec<String> of commands
+                let commands = evaluate_string_list(content_list, context)?;
+                // Join the commands with newlines to create a single script
```
wolfv, line 476: This should stay a list (especially on Windows) because we add some checks in between the elements.

PR #1966, opened 2025-10-29
Py rattler build refactor

wolfv:


py-rattler-build/rattler_build/stage1.py
```diff
@@ -0,0 +1,288 @@
...
+
+if TYPE_CHECKING:
+    # For type checking, use Any as placeholder since we don't have stubs
+    _Stage1Recipe = Any
+    _Stage1Package = Any
+    _Stage1Build = Any
+    _Stage1Requirements = Any
+    _Stage1About = Any
+    _Stage1Source = Any
+    _Stage1StagingCache = Any
```
Hofer-Julian, line 19: Why would this be necessary?

py-rattler-build/rattler_build/stage1.py
```diff
@@ -0,0 +1,288 @@
...
+
+
+class About:
+    """About metadata at stage1 (fully evaluated)."""
+
+    def __init__(self, inner: _Stage1About):
+        self._inner = inner
+
+    @property
+    def homepage(self) -> Optional[str]:
```
Hofer-Julian, line 214: Let's remember to use `ruff` here to migrate all use cases of `Optional[str]` to the newer `str | None`  syntax

py-rattler-build/src/build_types.rs
```diff
@@ -0,0 +1,152 @@
...
+    #[getter]
+    fn archive_type(&self) -> String {
+        match self.inner.archive_type {
+            ArchiveType::Conda => "conda".to_string(),
+            ArchiveType::TarBz2 => "tar-bz2".to_string(),
+        }
+    }
+
+    /// Get the compression level
+    #[getter]
```
Hofer-Julian, line 108: I am curious about the heavy use of implicit getters and setters in this API.

My impression is that it's more common to use `PyPackagingSettings.set_compression_level(..)` instead of `PyPackagingSettings.compression_level = ..`, especially if there's logic happening behind the scenes

PR #1975, opened 2025-11-06
fix: skip overdepending check for packages without binaries

### LINKED ISSUE #1949: Overlinking against packages without shared objects
When building a recipe which contains a package without shared objects rattler-build will still complain that the package is overlinked.

The following recipe shows this problem

```yaml
package:
  name: test
  version: "0.1.0"

requirements:
  run:
    - pixi-build-api-version ==2
```

`pixi-build-api-version` is an empty package (its a mutex package). 

When building this package with:

```shell
rattler-build build --recipe recipe.yaml
```

you will get the warning:

```
warning Overdepending against pixi-build-api-version
```

baszalmstra:
Fixes issue #1949 where rattler-build incorrectly warns about overdepending for packages that contain no linkable binaries.

The overdepending check looks for unused library dependencies by checking which libraries the package's binaries link against. For packages without any binaries, this check is meaningless - there's nothing that could link to libraries.

This prevents false positives when building packages with only run dependencies like `pixi-build-api-version` (version constraint packages).

Fixes #1949

## AI Disclosure

Claude Code Web wrote this.

### DIFF
--- a/src/post_process/checks.rs
+++ b/src/post_process/checks.rs
@@ -372,6 +372,12 @@ pub fn perform_linking_checks(
         tracing::info!("{linked_package}");
     });
 
+    // If there are no binaries, skip the overdepending check.
+    // Packages without binaries can't link against libraries.
+    if package_files.is_empty() {
+        return Ok(());
+    }
+
     // If there are any unused run dependencies then it is "overdepending".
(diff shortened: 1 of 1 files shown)


wolfv:
We have a better fix in #2181 :)

PR #1978, opened 2025-11-06
feat: add `expected-commit` verification to git sources

### LINKED ISSUE #1895: Add `expected-commit` verification to git sources
Wolfi/Melange has a neat feature to test the "expected commit" when checking out a tag or a branch from a git repository. We could also add a new field to the `git` source to have an `expected-commit` and fail if the commit that the branch is pointing to doesn't match (anymore).

Could look like this:

```yaml
source:
  - git: git://anongit.gentoo.org/proj/pax-utils.git
    tag: v${{package.version}}
    expected-commit: 9ef54b472e42ba2c5479fbd86b8be2275724b064
```

baszalmstra:
This PR implements the feature requested in issue #1895, adding an optional `expected-commit` field to git source configurations.

Changes:
- Added `expected_commit` field to GitSource struct in source.rs
- Added GitCommitMismatch error variant to SourceError enum for structured error handling with expected, actual, and rev fields
- Implemented verification logic in git_source.rs to compare actual commit against expected commit after checkout
- Added comprehensive unit tests for serialization/deserialization
- Added integration tests for both success and failure cases
- Created test recipe demonstrating the feature
- Updated documentation in recipe_file.md with usage examples

The feature helps ensure security and reproducibility by detecting when a git tag or branch has been moved to point to a different commit than expected.

Closes #1895

### DIFF
--- a/test-data/recipes/git_source_expected_commit/recipe.yaml
+++ b/test-data/recipes/git_source_expected_commit/recipe.yaml
@@ -5,7 +5,7 @@ package:
 source:
   - git: https://github.com/prefix-dev/rattler-build
     tag: v0.1.0
-    expected-commit: df83c1edf287a756b8fc995e03e9632af0344777
+    expected_commit: df83c1edf287a756b8fc995e03e9632af0344777
 
 build:
   script:
--- a/test-data/recipes/git_source_expected_commit/recipe.yaml
+++ b/test-data/recipes/git_source_expected_commit/recipe.yaml
@@ -1,0 +1,12 @@
+package:
+  name: "git_source_expected_commit"
+  version: "1"
+
+source:
+  - git: https://github.com/prefix-dev/rattler-build
+    tag: v0.1.0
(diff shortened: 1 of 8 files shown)


test-data/recipes/git_source_expected_commit/recipe.yaml
```diff
@@ -0,0 +1,12 @@
+package:
+  name: "git_source_expected_commit"
+  version: "1"
+
+source:
+  - git: https://github.com/prefix-dev/rattler-build
+    tag: v0.1.0
+    expected-commit: df83c1edf287a756b8fc995e03e9632af0344777
```
wolfv: Great stuff! I think so far we use _ mostly, so I think renaming to `expected_commit` would be better. Also we should probably make a cep amendment.

wolfv (changes requested):

wolfv (approved):

PR #1985, opened 2025-11-14
fix: determenistic build order

remimimimimi:


src/main.rs
```diff
@@ -113,6 +112,9 @@ async fn async_main() -> miette::Result<()> {
                 }
             }
 
+            // Sort recipes to get determenistic build order.
+            recipe_paths.sort();
```
wolfv, line 116: This will certainly improve determinism, but we are also doing a topological sort in other places. Maybe we can figure out how to make that toplogicial sort deterministic (so that the order of the outputs is deterministic)?

mwiebe: Not sure if this helps, but I tried to define a stable topological sort once here: https://github.com/OpenJobDescription/openjd-model-for-python/blob/f517fbf927deef02b4ea4e9f0603b3b6423b8c16/src/openjd/model/_step_dependency_graph.py#L124

PR #1987, opened 2025-11-14
fix: avoid using corrupted git cache dirs

andrewjcg:
Try to detect cases where the git cache dir gets corrupted and needs to be
re-cloned (e.g. can happen if builds get killed).


wolfv:
Thanks!

PR #2006, opened 2025-11-25
feat: add `package extract` subcommand

baszalmstra:
Add a new `rattler-build package extract` command that allows users to extract conda packages (.conda or .tar.bz2) to a directory. This command supports:

- Extracting local package files
- Downloading and extracting packages from URLs (with authentication support)
- Custom destination directory via -d/--dest flag
- Automatic destination naming based on package filename

This functionality mirrors the `rattler extract` command from rattler (added in conda/rattler#1832) but integrates it into rattler-build's `package` subcommand alongside the existing `inspect` command.

Documentation for package inspect/extract has been added to the debugging builds guide.

### DIFF
--- a/src/package_info.rs
+++ b/src/package_info.rs
@@ -6,7 +6,9 @@ use std::sync::Arc;
 use fs_err as fs;
 use indicatif::HumanBytes;
 use miette::{Context, IntoDiagnostic};
-use rattler_conda_types::package::{AboutJson, IndexJson, PathType, PathsJson, RunExportsJson};
+use rattler_conda_types::package::{
+    AboutJson, ArchiveType, IndexJson, PathType, PathsJson, RunExportsJson,
+};
 use rattler_networking::{AuthenticationMiddleware, AuthenticationStorage};
 use rattler_package_streaming::seek::read_package_file;
 use reqwest::Client;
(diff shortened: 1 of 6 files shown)


src/package_info.rs
```diff
@@ -320,3 +327,140 @@ fn output_human_readable(
 
     Ok(())
 }
+
+/// Strips package extensions (.tar.bz2 or .conda) from a filename
```
wolfv, line 333: I think there should be a function in rattler that can do this better! :)

src/package_info.rs
```diff
@@ -320,3 +327,140 @@ fn output_human_readable(
...
+    if let Some(stripped) = filename.strip_suffix(".tar.bz2") {
+        stripped.to_string()
+    } else if let Some(stripped) = filename.strip_suffix(".conda") {
+        stripped.to_string()
+    } else {
+        filename.to_string()
+    }
+}
+
+/// Creates an HTTP client with authentication middleware
```
wolfv, line 340: Can you add a TODO comment to use the refactored `rattler-build-networking` once we merged the refactor?

wolfv (approved):

src/package_info.rs
baszalmstra:
```suggestion
/// Creates an HTTP client with authentication middleware
/// TODO: Refactor to use rattler-build-networking when its available.
```

PR #2074, opened 2025-12-30
support for powershell as interpreter

paperchalice:
This pull request implement part of `windows_quirks.md`, i.e. using `.ps1` file as build script.
This feature should be experimental but I don't now how to implement it.

Future works:
1. ~~It should use its own preamble, so we can inject preference variables like `ErrorActionPreference` and `PSNativeCommandUseErrorActionPreference` (pwsh 7.4+).~~
2. ~~Provide non environment variable aliases for standard environment variables, e.g. `PREFIX`.~~
3. ~~Detect interpreter directly from build script suffix, as described in `windows_quirks.md`.~~

### DIFF
--- a/src/script/interpreter/powershell.rs
+++ b/src/script/interpreter/powershell.rs
@@ -32,7 +32,7 @@ impl Interpreter for PowerShellInterpreter {
             let ver = out
                 .trim()
                 .split(' ')
-                .last()?
+                .next_back()?
                 .split('.')
                 .collect::<Vec<&str>>();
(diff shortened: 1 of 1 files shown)


Glatzel:
Powershell before 7.4 without `PSNativeCommandUseErrorActionPreference` will successfully and slilently exit even if third part cli tool failed, it also causes some unexpected behaviours. I suggest  to add doc about the error behaviour in different poweshell version, and add warning if the powershell version lower than 7.4.

PR #2136, opened 2026-02-10
feat: update default windows compiler

### LINKED ISSUE #2340: Change default windows compiler to vs2022
Currently, rattler-build resolves vs2017 as the default MSVC version for windows from the JNinja helper `compiler('c')` or 'compiler('cxx')`. Although this works, current GitHub actions runners don't include VisualStudio 2017 build tools calling the setup scripts of vs2017_win-64 to fail. This can be difficult to debug as someone setting up a CI/CD run for the first time with rattler-build and requires them to resolve vs2017 as a dependency before invoking rattler-build.

I propose rattler-build changes the default VisualStudio build tools version for windows to vs2022 to follow in step with [conda-forge](https://conda-forge.org/news/2025/06/11/moving-to-vs2022/) and help simplify this CI/CD pain point.

Also, given Microsoft [no longer provides distributions](https://learn.microsoft.com/en-us/visualstudio/releases/2022/release-notes?tabs=allfeatures#17.14.11) of MSVC 14.1, the only way to get the default to work is to download a version distributed from [a third party within the action runner](https://community.chocolatey.org/packages/visualstudio2017buildtools) then manually mangling the setup script to detect it

Alternative solution:
- just pin vs2022_win-64 if on windows but I feel this goes away from the "sane default" approach the JNinja option offers.

Glatzel:
references: 
- https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/main/recipe/conda_build_config.yaml

fix: #2340


Glatzel:
The error comes from upstream `vc` package from `vs2019`. The expansion of temp(prefix) path in test in vc activate.bat exceeds 8192 characters.

https://github.com/conda-forge/vc-feedstock/blob/41025d6f35dbe5134294f7ad2cbca95a25b8ec65/recipe/activate.bat#L155-L159

Glatzel:
The doc is generated by ai, but I have reviewed and revised it.

Hofer-Julian:
Will check this out @Glatzel as soon as CI is green. Please ping me sometime next week if I don't get to it :)

test/end-to-end/conftest.py
```diff
@@ -1,11 +1,35 @@
...
+    else:
+        yield
+
+
+def pytest_configure(config):
+    # On Windows, use a short absolute path to avoid hitting the cmd.exe
+    # line-length limit during VS compiler activation (vcvars64.bat).
+    if sys.platform == "win32":
+        worker_id = os.environ.get("PYTEST_XDIST_WORKER", "bld").replace("gw", "")
+        config.option.basetemp = Path(f"C:/{worker_id}")
```
Hofer-Julian: @baszalmstra can you please take a look if this makes sense to you?

pytest.ini
```diff
@@ -1,5 +1,4 @@
 [pytest]
-addopts = --basetemp=pytest-temp
```
Hofer-Julian, line 2: This should not be removed

Hofer-Julian (changes requested):

pytest.ini
Glatzel: @wolfv changed this and add the `pytest_configure`. Then I add the `clean_path_on_win32`, I do not know if `clean_path_on_win32` can make the path short enough itself, but we can try it. Or if you guys have better apporch to shorten the path, just change the code.

Glatzel:
@Hofer-Julian can you have a look?

Hofer-Julian (approved):
Thank you @Glatzel!

PR #2140, opened 2026-02-13
chore: Make rattler-build build with latest rattler main branch

hunger:
This is probably not relevant till we bump the rattler version, but I needed it and thought I'd push it for later reference:-)


hunger:
I did not commit the various cargo dependency overrides needed ;-)

PR #2141, opened 2026-02-13
fix: resolve 'Is a directory' error when using file_name with patches

wolfv:
- Guard against empty paths in custom_patch_stripped_paths that caused work_dir.join("") to resolve to the work directory itself
- Fix convert_url_source and convert_path_checksum to handle SHA256 checksums (previously only MD5 was converted)
- Add regression test for the file_name + patches interaction


wolfv:
- [ ] For the checksum, I think we should add _both_ md5 and sha256 (if we have them) and also validate both so that no invalid data can slip through.

PR #2142, opened 2026-02-13
fix: always set `SP_DIR` during script execution

wolfv:
We did not always set the `SP_DIR` variable that points to the `site-packages` folder. This was sometimes unset because for noarch packages, we cannot statically derive it (as we do not know the Python version chosen ahead of time). 

However, this leads to increased friction when trying to build packages cross-platform. The idea of this PR is to always set the SP_DIR if `python` is installed in the `host` environment.


wolfv:
Added more description. However, before merging, let me double check the logic tomorrow!

PR #2148, opened 2026-02-13
feat: implement sigstore publisher verification of URL sources

wolfv:
This adds a `publisher` field to the URL source.


wolfv:
Documentation added, and also moved behind the `--experimental` flag for the time being. We might iterate a bit more with how the identifiers are specified etc.

Issue #2097, opened 2026-01-19
python bindings: crashes when building recipes without passing recipe_path


jsmolic:
When using updated python bindings from https://github.com/prefix-dev/rattler-build/pull/2057 to build a package, if `recipe_path` is not passed to the `build_rendered_variants()`, the interface will crash without giving immediate clue as to what was incorrect.

For example, running the following commands in `iregi-feedstock` source directory will results in a `PyRattlerBuildError: Patch file not found` error:

```
>>> from rattler_build.stage0 import Stage0Recipe
>>> from rattler_build.variant_config import VariantConfig
>>> from rattler_build.render import build_rendered_variants
>>> recipe=Stage0Recipe.from_file("./recipe/recipe.yaml")
>>> config=VariantConfig.from_file(".ci_support/linux_64_.yaml")
>>> rendered=recipe.render(config)
>>> results=build_rendered_variants(rendered, output_dir="./output")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/jakov/git-conda/rattler-build/py-rattler-build/src/rattler_build/render.py", line 416, in build_rendered_variants
    result = variant.run_build(
             ^^^^^^^^^^^^^^^^^^
  File "/home/jakov/git-conda/rattler-build/py-rattler-build/src/rattler_build/render.py", line 334, in run_build
    rust_result = build_rendered_variant_py(
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^
rattler_build.PyRattlerBuildError: PyRattlerBuildError: Patch file not found: output/_no_recipe/makefile_ML5_lib.patch

Build log:
Starting build of 1 outputs
Fetching source from url: https://github.com/hep-packaging-coordination/mg5amcnlo-release-mirror/releases/download/3.5.7/MG5_aMC_v3.5.7.tar.gz
Downloading from: https://github.com/hep-packaging-coordination/mg5amcnlo-release-mirror/releases/download/3.5.7/MG5_aMC_v3.5.7.tar.gz
Copying source from cache: /home/jakov/git-conda/iregi-feedstock/output/src_cache/e78747b165ae2184_extracted to /home/jakov/git-conda/iregi-feedstock/output/bld/rattler-build_iregi-static_1768809657/work
Applying patch: output/_no_recipe/makefile_ML5_lib.patch
```

while `results=build_rendered_variants(rendered, output_dir="./output", recipe_path="recipe/recipe.yaml")` will succesfully build the package.

PR #2166, opened 2026-02-16
feat(py): infer recipe path output

Hofer-Julian:
Fixes https://github.com/prefix-dev/rattler-build/issues/2097

With this PR, we can infer `recipe_path` and `output` path in many places.
The starting point is always a `Stage0Recipe` which either:
- infers the path, since you open a file from disk,
- is given one by you, or
- it simply uses a temporary directory

In all other instances but `JinjaConfig`, we then infer the `recipe_path` from the recipe. The advantage can be seen nicely in the examples and notebooks (unfortunately the diff of the latter don't render nicely on GitHub).

We don't infer for `JinjaConfig`, since this is pretty much decoupled from a recipe and can also only apply to a variant file


crates/rattler_build_script/src/execution.rs
```diff
@@ -273,7 +273,10 @@ impl Script {
                 if let Some(recipe_file) = recipe_file {
                     match fs_err::read_to_string(&recipe_file) {
                         Err(e) => Err(e),
-                        Ok(content) => Ok(ResolvedScriptContents::Path(recipe_file, content)),
+                        Ok(content) => Ok(ResolvedScriptContents::Path(
+                            recipe_file,
+                            content.replace("\r\n", "\n"),
```
wolfv: Is this related?

test-data/recipes/test-rendering/test.py
```diff
@@ -1 +1 @@
-print("i am the test")
\ No newline at end of file
+print("i am the test")
```
wolfv: I saw this being done by ruff a few times as well, which changes the snapshot.

I think we can either do this, or keep excluding the file from linting/formatting.

Hofer-Julian:
It should be fine now

test-data/recipes/test-rendering/test.py
Hofer-Julian: Done

crates/rattler_build_script/src/execution.rs
Hofer-Julian: Removed

Issue #2039, opened 2025-12-08
Implement CEP-28 (customizable linking checks on Windows)


wolfv:
We should follow the implementation in conda-build to implement CEP-28: https://github.com/conda/conda-build/pull/5806

CEP: https://github.com/conda/ceps/blob/main/cep-0028.md

PR #2173, opened 2026-02-17
Improve Windows PE file detection and system library allowlisting (CEP 28)

wolfv:
## Summary
This PR improves the robustness of Windows PE file handling and system library allowlisting by adding file extension-based filtering and expanding the system library detection strategy.

## Key Changes

- **Add PE file extension filtering**: Introduced `has_pe_extension()` function that checks for known PE file extensions (dll, exe, pyd, sys, ocx, drv, cpl, scr, efi) before attempting to parse files as PE binaries. This avoids unnecessary parsing attempts on non-PE files.

- **Add SHLWAPI.dll to allowlist**: Extended `WIN_ALLOWLIST` to include SHLWAPI.dll, a commonly used Windows system library.

- **Improve system library detection**: Enhanced `add_windows_system_libs()` to use glob patterns for well-known Windows system directories (C:/Windows/system32/** and C:/Windows/SysWOW64/**) in addition to the hardcoded allowlist. This provides better coverage for system libraries while maintaining backward compatibility for cross-compiled builds.

- **Expand shared object detection**: Added `provides_shared_objects()` method to `PackageNature` that returns true for packages providing shared objects (DSO libraries, run-exports, interpreters like Python, and plugin libraries). Updated linking checks to use this method instead of only checking for `DSOLibrary`, allowing interpreters and other legitimate shared object providers to be linked against.

- **Improve logging and comments**: Changed malformed PE file warnings to debug level (since files like .pyc and .npz can partially match PE signatures), fixed grammar in comments, and added detailed documentation explaining the fallback strategy for system library detection.

- The PE extension check uses case-insensitive comparison to handle various filename conventions
- Glob patterns use forward slashes which are normalized by globset on Windows to match backslash paths
- The fallback strategy maintains two-tier allowlisting: directory-based patterns for normal builds and bare DLL names for cross-compiled builds

https://claude.ai/code/session_01DCjm7BriTknbio1fFvMK8A

Fixes: https://github.com/prefix-dev/rattler-build/issues/2039

### DIFF
--- a/src/post_process/checks.rs
+++ b/src/post_process/checks.rs
@@ -54,10 +54,11 @@ struct PackageLinkInfo {
 
 impl fmt::Display for PackageLinkInfo {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let file = self.file.display().to_string().replace('\\', "/");
         writeln!(
             f,
             "\n[{}] links against:",
-            console::style(self.file.display()).white().bold()
+            console::style(&file).white().bold()
         )?;
         for (i, package) in self.linked_packages.iter().enumerate() {
             let connector = if i != self.linked_packages.len() - 1 {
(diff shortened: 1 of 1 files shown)


src/post_process/checks.rs
```diff
@@ -409,24 +421,38 @@ fn add_windows_system_libs(
...
+    let host_result = load_dsolists(output.prefix(), &subdir)?;
+
+    // If no dsolist files matched at all, fall back to the hardcoded allowlist.
+    // If files *were* found (even with empty allow/deny), respect them – an
+    // explicit empty allowlist means "nothing is allowed".
+    if build_result.is_none() && host_result.is_none() {
+        // 1. Allow any DLL under well-known Windows system directories.
+        //    Globset normalises backslashes on Windows, so forward-slash
+        //    patterns match paths like "C:\Windows\system32\SHLWAPI.dll".
+        for dir in &["C:/Windows/system32/**", "C:/Windows/SysWOW64/**"] {
```
wolfv: double check this vs. conda-build

src/post_process/checks.rs
```diff
@@ -244,18 +244,24 @@ pub fn validate_dsolist_files(package_dir: &Path) -> Result<(), LinkingCheckErro
...
-    prefix: &Path,
-    subdir: &str,
-) -> Result<(Vec<String>, Vec<String>), LinkingCheckError> {
+///
+/// Returns `None` when no dsolist files matching `subdir` were found (the
+/// caller should fall back to a hardcoded allowlist).  Returns
+/// `Some((allow, deny))` when at least one matching file was processed –
+/// even if the resulting lists are empty, because an explicit empty
+/// allowlist means "nothing is allowed".
+fn load_dsolists(prefix: &Path, subdir: &str) -> Result<Option<DsoAllowDeny>, LinkingCheckError> {
```
wolfv: If possible we should use rattler_conda_types::Platform here, not `subdir: str`.

PR #2188, opened 2026-02-19
fix: prevent post-link scripts from leaking files into downstream packages

chrisburr:
- When a host dependency has a post-link script that generates files (e.g. cache files), those files are not recorded in `conda-meta/`. `Files::from_prefix()` then incorrectly attributes them to the downstream package being built.
- This fix snapshots the host prefix before and after `install_environments()` to compute the delta of files added during installation (including post-link artifacts), then excludes them from new-file detection. The delta approach avoids interfering with files restored from a staging cache.
- Adds an end-to-end test with two recipes (`post-link-leaker` and `post-link-downstream`) that verifies leaked files are excluded.

Related: conda-forge/gtk3-feedstock#81

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -11,6 +11,7 @@ use crate::{
     apply_patch_custom,
     metadata::{Output, build_reindexed_channels},
     package_test::PackageContentsTestExt as _,
+    packaging::record_files,
     render::{resolved_dependencies::RunExportsDownload, solver::load_repodatas},
     tool_configuration,
 };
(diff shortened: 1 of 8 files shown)


chrisburr:
This PR looks ready to me now.

It's out of scope for this PR but I wonder if it is worth tracking files added by post-link scripts in conda-meta. I also wonder if it's worth rattler-builds's testing step uninstalling the package at the end to help check if the post-unlink script is correct.

PR #2195, opened 2026-02-20
docs: pin `click` dependency

Hofer-Julian:
See also https://github.com/mkdocs/mkdocs/issues/4032


Hofer-Julian:
> Should we do a repodata patch for conda-forge?

Good idea, we should

PR #2196, opened 2026-02-21
Support relative paths in git functions for local repositories

wolfv:
## Summary
This PR adds support for resolving relative paths in git Jinja functions, allowing recipes to reference local git repositories using paths relative to the recipe directory.

## Key Changes
- **Pass recipe directory to Git object**: The `recipe_dir` is now extracted from `recipe_path` and passed to the `Git` struct, enabling path resolution relative to the recipe location.

- **Add local path resolution**: Implemented `resolve_local_path()` method that attempts to resolve paths in this order:
  1. Relative to the recipe directory (if set)
  2. As an absolute path
  3. Relative to the current working directory

- **Update git functions for local repositories**: Modified `head_rev()`, `latest_tag()`, and `latest_tag_rev()` to:
  - Check if the source is a local directory using the new resolution logic
  - Execute git commands directly in that directory using `git_command_in_dir()`
  - Fall back to remote `git ls-remote` calls for non-local sources

- **Add helper function**: Created `git_command_in_dir()` to execute git commands in a specific directory with proper error handling.

- **Add comprehensive test**: Included `eval_git_relative_path()` test that verifies all three git functions work correctly with relative paths from a recipe directory, with appropriate skip conditions for older git versions in cross-compilation environments.

## Implementation Details
The solution maintains backward compatibility by falling back to remote git operations when a path cannot be resolved locally. The recipe directory context is now available throughout the git operations, enabling more flexible recipe configurations that can reference sibling repositories.

https://claude.ai/code/session_01DwoqNVZte3ncWrKupSfQcV

### DIFF
--- a/crates/rattler_build_jinja/src/git.rs
+++ b/crates/rattler_build_jinja/src/git.rs
@@ -60,22 +60,28 @@ fn git_command_in_dir(dir: &Path, args: &[&str]) -> Result<String, minijinja::Er
 }
 
 impl Git {
-    /// Try to resolve `src` as a local path. First tries relative to the
-    /// recipe directory (if set), then as an absolute path, then relative to
-    /// the current working directory.
+    /// Try to resolve `src` as a local path. Relative paths are resolved
+    /// against the recipe directory only. Absolute paths are checked directly.
+    /// Returns `None` if `src` does not point to a local directory (i.e. it is
+    /// likely a remote URL).
     fn resolve_local_path(&self, src: &str) -> Option<PathBuf> {
-        // Try relative to recipe directory first
+        let path = Path::new(src);
+
+        // Absolute paths are checked directly
+        if path.is_absolute() {
+            return if path.is_dir() {
+                Some(path.to_path_buf())
+            } else {
+                None
+            };
+        }
+
+        // Relative paths are resolved against the recipe directory
         if let Some(recipe_dir) = &self.recipe_dir {
             let resolved = recipe_dir.join(src);
             if resolved.is_dir() {
                 return Some(resolved);
             }
-        }
-
-        // Try as-is (absolute or relative to cwd)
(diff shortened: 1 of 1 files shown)


crates/rattler_build_jinja/src/git.rs
```diff
@@ -33,8 +38,57 @@ fn get_command_output(command: &str, args: &[&str]) -> Result<String, minijinja:
...
+        Ok(String::from_utf8(output.stdout).map_err(|e| {
+            minijinja::Error::new(minijinja::ErrorKind::InvalidOperation, e.to_string())
+        })?)
+    }
+}
+
 impl Git {
+    /// Try to resolve `src` as a local path. First tries relative to the
+    /// recipe directory (if set), then as an absolute path, then relative to
+    /// the current working directory.
```
Hofer-Julian: What is the use case for also trying the current working directory? Do we have other instances where rattler-build behaves differently depending on your working directory?

crates/rattler_build_jinja/src/git.rs
```diff
@@ -8,6 +12,7 @@ use minijinja::{
 #[derive(Debug)]
 pub(crate) struct Git {
     pub(crate) experimental: bool,
+    pub(crate) recipe_dir: Option<PathBuf>,
```
Hofer-Julian, line 15: Is this optional because Jinja could also apply to variants?

wolfv: It would not work for an "in-memory" recipe (that doesn't actually have a recipe_dir).

crates/rattler_build_jinja/src/git.rs
wolfv: I think this is a good point. We shoudl not fall back to the current working directory, indeed.

PR #2205, opened 2026-02-23
feat: python recipe from pyproject.toml

wolfv:


crates/rattler_build_recipe_generator/src/pyproject.rs
```diff
@@ -0,0 +1,1013 @@
...
+                .requirements
+                .host
+                .push(format!("python {}", python_req));
+        } else {
+            recipe.requirements.host.push("python".to_string());
+        }
+        for req in &metadata.build_requires {
+            let mapped = map_requirement(req, mapping, use_mapping).await;
+            recipe.requirements.host.push(mapped);
+        }
```
pb01ka, line 434: Here, `build_requires` entries are mapped via `map_requirement` and pushed directly into `host`. However, for `metadata.dependencies` entries go through `format_requirement` as well, and the `"- "` prefix is stripped before being pushed into `run`,

```rust
        for req in &metadata.dependencies {
            let mapped = map_requirement(req, mapping, use_mapping).await;
            let formatted = format_requirement(&mapped);
            recipe
                .requirements
                .run
                .push(formatted.trim_start_matches("- ").to_string());
        }
```

To the best of my knowledge, `format_requirement` normalises PEP 508 specifiers to conda syntax (e.g. something like `requests>=2.20` will become `requests >=2.20`). Skipping it for host means that `setuptools>=64` ends up verbatim in the recipe rather than `setuptools >=64`. Therefore I think, the same `format_requirement` call (with the `"- "` strip) should be applied to `build_requires` entries pushed into host.

crates/rattler_build_recipe_generator/src/pyproject.rs
```diff
@@ -0,0 +1,1013 @@
...
+    #[cfg_attr(feature = "cli", arg(default_value = "pyproject.toml"))]
+    pub path: PathBuf,
+
+    /// Whether to write the recipe to a folder
+    #[cfg_attr(feature = "cli", arg(short, long))]
+    pub write: bool,
+
+    /// Whether to use the conda-forge PyPI name mapping
+    #[cfg_attr(feature = "cli", arg(short, long, default_value = "true"))]
+    pub use_mapping: bool,
```
pb01ka, line 28: I tested on a small example and found that, with `clap`, a `bool` field annotated as `arg(short, long)` is treated as a presence flag. `default_value = "true"` sets the default but provides no mechanism for a user to pass `false` (**is this intentional?**). Passing `--use-mapping=false` on the command line will fail because `false` is parsed as a positional argument rather than the flag's value.

One fix is to use `ArgAction::Set` along with a default value,

```suggestion
    /// Whether to use the conda-forge PyPI name mapping
    #[cfg_attr(feature = "cli", arg(short, long, default_value = "true", action = clap::ArgAction::Set))]
    pub use_mapping: bool,
```

Here's a small example,

```rust
use clap::Parser;

#[derive(Parser, Debug)]
struct Opts {
    /// Original (broken): bool flag with default_value = "true"
    /// Passing --use-mapping=false will produce an error.
    #[arg(short, long, default_value = "true")]
    first: bool,

    /// invert to --no-mapping (idiomatic for default-on flags).
    #[arg(short, long, default_value = "true", action = clap::ArgAction::Set)]
    second: bool,
}

fn main() {
    let opts = Opts::parse();
    println!("first  (original) = {}", opts.first);
    println!("second (ArgAction::Set) = {}", opts.second);
}
```

**Outputs**

```zsh
21:07:44:/tmp/claptest % ./target/debug/claptest
first  (original) = true
second (ArgAction::Set) = true
21:07:45:/tmp/claptest % ./target/debug/claptest --first=false
error: unexpected value 'false' for '--first' found; no more were expected

Usage: claptest --first

For more information, try '--help'.
21:07:53:/tmp/claptest % ./target/debug/claptest --second=false
first  (original) = true
second (ArgAction::Set) = false
```

crates/rattler_build_recipe_generator/src/pyproject.rs
```diff
@@ -0,0 +1,1013 @@
...
+        .and_then(|v| v.as_str())
+        .unwrap_or("");
+
+    if backend.contains("setuptools_scm") {
+        "${{ environ.get('SETUPTOOLS_SCM_PRETEND_VERSION', '0.0.0') }}".to_string()
+    } else if backend.contains("hatch") {
+        "${{ environ.get('HATCH_BUILD_VERSION', '0.0.0') }}".to_string()
+    } else {
+        "0.0.0".to_string()
+    }
```
pb01ka, line 335: How about,

```suggestion
    } else {
          tracing::warn!(
                "Unknown dynamic version backend '{}'; defaulting to 0.0.0. \
                  Set [project.version] or override via [tool.rattler-build.context].",
                  backend
           );
           "0.0.0".to_string()
    }
```

This is because there might be some backends like `poetry` (or `meson`? or anything else for that matter) which the end user might want to use. For those, they get a syntactically valid recipe that silently packages version `0.0.0`. May be, giving a warning would be helpful in that case?

**Just a suggestion, might be an overstatement to give warnings**

crates/rattler_build_recipe_generator/src/pyproject.rs
```diff
@@ -0,0 +1,1013 @@
...
+        );
+    } else if let Some(url) = &ov.source_url {
+        recipe.source.push(
+            UrlSourceElement {
+                url: vec![url.clone()],
+                sha256: ov.source_sha256.clone(),
+                md5: None,
+            }
+            .into(),
+        );
```
pb01ka, line 399: Here, if a user puts both `url` and `git` in `[tool.rattler-build.source]`, the `git` branch silently wins and the `url` value is discarded.

In practice, this cannot happen in a well-formed config - `url` and `git` are semantically mutually exclusive and no real `pyproject.toml` would set both intentionally. The behaviour is also deterministic (git always wins), so a user who accidentally sets both will see the expected output. 

**A validation error would be a nice touch for a misconfigured file, but this is genuinely low priority and can be left as-is.**

pb01ka (commented):
I tried my best to review the PR. We can also work on the documentation. All in all everything else looks good to me.

Please feel free to correct me where ever you think I stated facts wrongly or misinterpreted your intentions.

pb01ka:
@wolfv If you want I can work on addressing the reviews (those which seem valid to you). I will send a PR to your fork or send a new one on this repository with the reviews addressed.

PR #2208, opened 2026-02-24
feat: add code signing support for macOS and Windows binaries

wolfv:
## Motivation

We want to replace `cargo-dist` by building the Pixi (and rattler-build) binaries with rattler-build, including proper signing. We would then extract the files from the conda package for the release.

## Claude comments

Add a new `build.signing` section to the recipe schema that allows configuring native code signing for macOS (codesign) and Windows (signtool or Azure Trusted Signing).

Key changes:
- Recipe schema: `build.signing.macos` (identity, keychain, entitlements, options) and `build.signing.windows` (local certificate via signtool OR Azure Trusted Signing via azure_endpoint/account/profile)
- Stage0 parsing + Stage1 evaluation for signing configuration
- New `post_process::signing` module that signs all Mach-O/PE binaries after relinking but before packaging
- Signature verification after signing
- Safety check: errors if signed binaries contain the build prefix (which would be corrupted by conda's prefix replacement at install time)
- Added `Signtool` to system_tools enum

The signing step integrates into the packaging pipeline after relinking and post-processing but before metadata creation, ensuring archives contain properly signed binaries.

https://claude.ai/code/session_017eLvzM9YezN2oJ6Q7etqHu

### DIFF
--- a/docs/code_signing.md
+++ b/docs/code_signing.md
@@ -45,6 +45,45 @@ Because relinking invalidates existing signatures, rattler-build first applies
 an ad-hoc signature (`codesign -s -`) on macOS during relinking, then overwrites
 it with the real identity during the signing step using `--force`.
 
+## External signing config (`--signing-config-file`)
+
+Signing is often a CI/build-environment concern -- not something that belongs
+in the recipe itself. The `--signing-config-file` flag lets you keep your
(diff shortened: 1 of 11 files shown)


wolfv:
Another motivating blog post: https://d34dl0ck.me/cargo-codesign/index.html

PR #2218, opened 2026-02-25
feat: WASM playground

wolfv:
This deploys a WebAssembly playground straight to Cloudflare pages: https://wasm-support.rattler-build-playground.pages.dev/


wolfv:
A small bug with `noarch` packages: outputs that produce the same hash aren't de-duplicated in the right hand side view.

Issue #2139, opened 2026-02-11
CEP amendments (recipe format)


wolfv:
We have a few things already merged and proposed that require us to make additional changes to the recipe spec CEP:

### `git.expected_commit` field

We added a `expected_commit` field to the Git source, so that you can do:

```yaml
source:
  git: https://github.com/prefix-dev/rattler-build
  tag: v0.52.1
  expected_commit: a78s8d798a7s87987a
```

This validates that the tag points to the commit that you would expect.

### `git.submodules` field

We added a boolean `submodules` field to the `git` object to be able to disable fetching of submodules:

```yaml
source:
  git: https://github.com/prefix-dev/rattler-build
  tag: v0.52.1
  submodules: false
```

### `url.sigstore` dictionary

For URL sources, we want to add a `sigstore` field. Not entirely spec'd out yet, but should look something like:

```yaml
package:
  name: flask
  version: 2.0.1

source:
  url: https://files.pythonhosted.org/packages/dc/6d/flask-${{ version }}.tar.gz
  sha256: 9a0cbdcbbd1e7b8e5f1a4b2c3e5d8f1c9e6a7b8c9d0e1f2a3b4c5d6e7f8
  sigstore:
    # This will check to ensure that the sdist was created by the git pipeline / tag as expected
    # And that the cryptographic signature is valid
    publisher: github:pallets/flask/tree/refs/tags/${{ version }}
```

### `app` section

This is a section that we "forgot" about. 

```yaml
app:
  entry: myapp launch
  icon: myapp.png
  summary: My application description
  type: web  # or "desk" for desktop apps
```

https://github.com/prefix-dev/rattler-build/issues/2099

### `post_process` section in `build`

Regex based replacements of text

```yaml
build:
  post_process:
    - files:
        - "*.txt"
      regex: "regex-post-process"
      replacement: "regex-post-process-replaced"
    - files:
        - "*.pc"
      regex: (?:-L|-I)?"?([^;\s]+/sysroot/)
      replacement: "$(CONDA_BUILD_SYSROOT_S)"

```

PR #2254, opened 2026-03-03
docs: add build.post_process to recipe specification

mohitdebian:
This PR documents the `build.post_process` section in the recipe specification.

The feature is already implemented in rattler-build (Stage0 → Stage1 evaluation and packaging),
but it was not described in the reference documentation.

This update adds:

- Field documentation for `files`, `regex`, and `replacement`
- Support for `if` / `then` / `else` conditionals (including nesting)
- Execution order and packaging phase behavior
- UTF-8 text requirement
- Clarification that Rust's `regex` crate semantics are used (`replace_all`)
- Output inheritance behavior in multi-output recipes

Related to #2139.

No code changes.
Documentation-only update.

### DIFF
--- a/docs/reference/recipe_file.md
+++ b/docs/reference/recipe_file.md
@@ -534,6 +534,73 @@ build:
     version_independent: true  # defaults to false
 ```
 
+### Post-processing
+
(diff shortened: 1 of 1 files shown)


mohitdebian:
Thanks. I'll add the `post_process` documentation to the recipe format repository as well :)

PR #2266, opened 2026-03-05
feat(test): show executed commands for failing test scripts

### LINKED ISSUE #2264: no visibility into failing `script:` calls during test
In https://github.com/conda-forge/tzdata-feedstock/pull/38, the recipe contains
```yaml
tests:
  - script:
      - test -f "${PREFIX}/share/zoneinfo/leapseconds"
      - test -f "${PREFIX}/share/zoneinfo/leap-seconds.list"
      - test -f "${PREFIX}/share/zoneinfo/iso3166.tab"
      - test -f "${PREFIX}/share/zoneinfo/zone1970.tab"
      - test -f "${PREFIX}/share/zoneinfo/zone.tab"
      - test -f "${PREFIX}/share/zoneinfo/tzdata.zi"
    # Make sure we only package zoneinfo and nothing else.
      - |
        dirs="$(
          find "${PREFIX}" -mindepth 1 -maxdepth 2 \
          \! -path "${PREFIX}/share" \! -path "${PREFIX}/conda-meta*"
        )"
        test "${dirs}" = "${PREFIX}/share/zoneinfo"
    # Make sure we only package timezone information files.
      - |
        heads="$(
          find "${PREFIX}/share/zoneinfo" -type f \
            \! -name \*.zi \! -name \*.tab \! -name leapseconds \! -name leap-seconds.list \
            -exec sh -c 'head -c4 $1 && echo' sh {} \; \
            | uniq
        )"
        test "${heads}" = TZif
```

_Some_ aspect of this is failing, only the logs tell me absolutely nothing
```
 │ ╭─────────┬─────────┬────────────┬────────────┬────────────╮
 │ │ Package ┆ Version ┆ Build      ┆ Channel    ┆       Size │
 │ ╞═════════╪═════════╪════════════╪════════════╪════════════╡
 │ │ tzdata  ┆ 2026a   ┆ h151e31d_0 ┆ .tmpVGZn3E ┆ 116.66 KiB │
 │ ╰─────────┴─────────┴────────────┴────────────┴────────────╯
 │ 
 │ Installing test environment
 │ ✔ Successfully updated the test environment
 │ Testing commands:
 │ × error Script failed with status 1
 │ × error 
 │ × error Script execution failed.
 │ × error 
 │ × error   Work directory: /Users/runner/miniforge3/conda-bld/test/test_tzdataZKh9aj/test
 │ × error   Prefix: /Users/runner/miniforge3/conda-bld/test/test_tzdataZKh9aj/test_run_env
 │ × error   Build prefix: None
 │ × error 
 │ × error To run the script manually, use the following command:
 │ × error 
 │ × error   cd "/Users/runner/miniforge3/conda-bld/test/test_tzdataZKh9aj/test" && ./conda_build.sh
 │ × error 
 │ × error To run commands interactively in the build environment:
 │ × error 
 │ × error   cd "/Users/runner/miniforge3/conda-bld/test/test_tzdataZKh9aj/test" && source build_env.sh
 │
Error:   × Test failed: failed to run test: IO Error: Script failed
```

At the very least, show me the equivalent of `set -x` in bash, i.e. the calls as they happen.

mohitdebian:
Fixes #2264

Currently rattler-build only reports "Script failed with status 1" when a
`tests.script` command fails, which makes it impossible to tell which command
caused the failure.

This change emits `set -x` in the bash preamble, immediately after the
activation script is sourced, so each command is printed before it runs:

```
Testing commands:
+ echo 'first command'
+ test -f $PREFIX/this-file-does-not-exist
first command
× error Script failed with status 1
```

Because the wrapper script (`conda_build.sh`) is shared by build and test
scripts, this applies to both — matching the request to "always use `-x`".

Placing `set -x` *after* activation (rather than running the whole wrapper with
`bash -x`) keeps the sourced environment setup untraced, so activation that
expands sensitive values does not leak into logs.


mohitdebian:
> Can we change this to always use `-x`?

Yes updated the implementation accordingly instead of injecting set -x into the script the bash interpreter now runs with -x when executing test scripts.

crates/rattler_build_core/src/package_test/run_test.rs
```diff
@@ -967,8 +967,27 @@ async fn run_commands_test(
     })?;
 
     tracing::info!("Testing commands:");
-    commands_test
-        .script
+
+    // Clone the script and prepend `set -x` to enable bash tracing for test scripts.
+    // Only inject for bash/sh scripts (interpreter unset defaults to bash on unix).
+    let mut test_script = commands_test.script.clone();
```
wolfv: this can all be removed now, right?

wolfv (dismissed):

crates/rattler_build_core/src/package_test/run_test.rs
mohitdebian: yes that block has been removed in the updated pr since tracing is now handled at the interpreter level via `bash -x` the `set -x` injection logic in `run_commands_test()` is no longer needed

h-vetinari:
Can we move this PR forward?

PR #2268, opened 2026-03-05
Fix: enable topological sorting of build outputs

wolfv:
We got reports that `--up-to` is not working properly anymore. This should fix it! It looks like the code path around topological sorting the outputs was accidentally commented out.

## Summary

This PR re-enables the topological sorting functionality for build outputs that was previously commented out. The function `sort_build_outputs_topologically` is now active and integrated into the build pipeline to ensure outputs are built in the correct dependency order.

https://claude.ai/code/session_01VDg4JbE8eC1LifBb4Zvbp5


wolfv:
Adding one more test for `--up-to` behavior and then we can merge this IMO

PR #2332, opened 2026-03-17
ci: preserve Unix permissions (executable bit) in zip archive for macOS

### LINKED ISSUE #2328: rattler-build doesn't set `x` attr in zip
Image

trim21:
Fixes #2328

When creating the macOS zip archive, `zipfile.ZipFile.write()` silently drops Unix metadata including the executable bit. This replaces it with a `ZipInfo`-based approach that reads `st_mode` from each file on disk and stores it in `external_attr` (upper 16 bits, the standard ZIP convention for Unix attributes).

- The binary gets its actual permissions (e.g. `0o755`) → executable bit preserved
- README/LICENSE get their actual permissions (e.g. `0o644`) → unchanged


Hofer-Julian (approved):
Thanks!

PR #2352, opened 2026-03-20
ci: `sign_macos.py` should preserve zip attr

trim21:
fix https://github.com/prefix-dev/rattler-build/issues/2328 again

the scripts/sign_macos.py unpack and pack zip artifact again so we need to also fix it。

Funny enough. in `sign_macos.py` we are using the right API `zf.write` and it do preserve the permission. But we are using `zf.extractall`, and it drop the permission...

https://github.com/python/cpython/issues/59999

### DIFF
--- a/scripts/sign_macos.py
+++ b/scripts/sign_macos.py
@@ -116,7 +116,7 @@ def sign_zip(archive: Path, identity: str) -> None:
                 ]
             )
 
-        # Re-create zip (preserving Unix permissions including executable bit)
+        # Re-create zip
         archive.unlink()
         with zipfile.ZipFile(archive, "w", zipfile.ZIP_DEFLATED) as zf:
             for item in sorted(tmp.rglob("*")):


scripts/package_binary.py
```diff
@@ -70,11 +70,7 @@ def main() -> None:
     else:
         with zipfile.ZipFile(archive_path, "w", zipfile.ZIP_DEFLATED) as zf:
             for item in sorted(archive_dir.iterdir()):
-                zi = zipfile.ZipInfo(f"{pkg_basename}/{item.name}")
-                zi.compress_type = zipfile.ZIP_DEFLATED
-                # Preserve Unix permissions (including executable bit) in ZIP external attrs
-                zi.external_attr = item.stat().st_mode << 16
```
wolfv, line 76: You are using the `external_attr` in the other script, but `write` will set it correctly, yeah?

trim21: the `zf.write(...)` handle external_attr but `zf.writestr` doesn't, so we do not need to handle `external_attr` by ourself

trim21: this is actually  what we are doing in sign_macos.py, which work as we expected

Image

PR #2401, opened 2026-04-02
fix: reset origin URL and sync submodules before update

Hofer-Julian:
The local checkout created by `rattler_git` points origin at the bare
cache database. Submodules with relative URLs resolve against origin,
so they fail to clone from non-existent local paths.

Fix by resolving submodule URLs against source URL instead of mutating origin.

Fixes prefix-dev/rattler-build#2379

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4694,9 +4694,9 @@ dependencies = [
 
 [[package]]
 name = "rattler"
-version = "0.40.3"
+version = "0.40.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5ba7b9c2977ddc74fecb7e68dd6b3f40958416220b9643d94db5b7387fd0cbcd"
+checksum = "d5e004d0371d9671efc0c41c9d5311a21479db551fd68236a3299b7d7b8447f0"
 dependencies = [
  "anyhow",
  "clap",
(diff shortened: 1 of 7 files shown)


wolfv:
Is this something that we should move to `rattler_git`?

Hofer-Julian:
> Is this something that we should move to `rattler_git`?

Good point, what do you say @baszalmstra? In rattler-build we currently handle git submodules. Would this make sense for `rattler_git` as well? That is currently only used by rattler-build but we plan to use it for Pixi in the future

baszalmstra:
Yeah that makes sense, what does cargo do and uv?

Hofer-Julian:
> Yeah that makes sense, what does cargo do and uv?

Good question. Cargo calls libgit2 which has a function for that `absolute_submodule_url`. `uv` doesn't handle it (yet): https://github.com/astral-sh/uv/issues/9822

baszalmstra:
Ideally we make it configurable whether submodules should be there or not. Wdyt?

Hofer-Julian:
> Ideally we make it configurable whether submodules should be there or not. Wdyt?



> Ideally we make it configurable whether submodules should be there or not. Wdyt?

That is already the case. Per default we sync submodules, but you can turn it off in the `recipe.yaml`

PR #2440, opened 2026-04-21
test: add test of channel prefixes for run w/ tests

beckermr:
I am trying to debug an issue I am finding over in https://github.com/conda-forge/r-nmf-feedstock/pull/27 where rattler-build cannot seem to resolve channel-prefixed run requirements when a package is being tested.


beckermr:
@wolfv are there other tests that are supposed to run here?

wolfv:
I think the issue might be that we cannot express this in the repodata - that uses still the simple format. Something we'll change in v3

PR #2442, opened 2026-04-22
test: add case: specific channel specification under requirements, w/ channel_sources and tests specified

dlaehnemann:
This comes from [an attempt to build a `conda-forge` package that references dependencies from the `bioconda` channel](https://github.com/conda-forge/r-nmf-feedstock/pull/27). The syntax seems to work for building, but the channel then isn't found during testing. Thus, this test case includes a `tests:` section.

Generally, if this channel specification works out with both building and testing, we can use this to consistently fix a number of `conda-forge` packages that added `bioconda` package dependencies at some point (usually `r-` packages from `CRAN` that added `bioconductor-` package dependencies).

### DIFF
--- a/test-data/recipes/channel_specific_with_sources/conda_build_config.yaml
+++ b/test-data/recipes/channel_specific_with_sources/conda_build_config.yaml
@@ -1,0 +1,2 @@
+channel_sources:
+  - conda-forge,bioconda
--- a/test-data/recipes/channel_specific_with_sources/recipe.yaml
+++ b/test-data/recipes/channel_specific_with_sources/recipe.yaml
@@ -1,0 +1,11 @@
+package:
+  name: channel_specific_with_sources
+  version: 1.0.0
+
+requirements:
+  run:
+    - bioconda::r-dwls
+
+tests:
+  - script:
+      - R -e "library('DWLS')"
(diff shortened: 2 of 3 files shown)


wolfv:
Just to note - this could/might work in the `test` section. It will _NOT_ work as part of the run dependnecies of a package as channel-specific run dependencies need the `::` or `[ ... ]` syntax, whcih is currently not emitted to the `repodata.json`. 

That will likely change with repodata v3, which is in a CEP that should be soon voted on.

dlaehnemann:
Would this work in the `build:` and `host:` sections under `requirements:`, as these are only used when building the package, and should have pulled the respective packages in at that time? That might already solve some of our use cases.

Also, do you have a cross-link to the CEP you mention? I had a quick look at some [open PRs on the CEPS repo](https://github.com/conda/ceps/pulls) and couldn't identify it.

dlaehnemann:
@wolfv I now found the recently approved CEP for `repodata v3`:
https://conda.org/learn/ceps/cep-0048

However, from looking through it, I am not sure if the channel-specific `host:` or `run:` dependency specification syntax is included there, and how it would become available. Do you have any pointers to where and how that kind of info might be included into `repodata.json` and then become available to be used on `conda-forge`, eventually?

wolfv:
@dlaehnemann that is a good question that I actually have to bring up ...

PR #2449, opened 2026-04-27
fix: remove PKG_* environment variables in staging cache, set variant values

wolfv:
## Summary
This change fixes environment variable handling in staging cache contexts by removing misleading PKG_* variables that would otherwise inherit from the parent output's identity.

## Key Changes
- Added variant-specific environment variables to the staging cache environment setup
- Removed PKG_NAME, PKG_VERSION, PKG_BUILDNUM, PKG_BUILD_STRING, and PKG_HASH from the environment since staging caches don't produce packages and these variables would be misleading in that context

## Implementation Details
The fix ensures that when setting up environment variables for a staging cache, we:
1. First extend with variant-specific environment variables
2. Then explicitly remove the PKG_* variables that carry package identity information, since a staging cache is not a package output and shouldn't inherit these values from the parent output

https://claude.ai/code/session_01Jy4Cz8QvigYixpjzvwzHLK

### DIFF
--- a/crates/rattler_build_core/src/staging.rs
+++ b/crates/rattler_build_core/src/staging.rs
@@ -281,9 +281,12 @@ impl Output {
         ));
         env_vars.extend(env_vars::env_vars_from_variant(self.variant()));
 
-        // Override PKG_NAME with the staging cache name so the script sees the
-        // staging cache identity rather than the inheriting output's name.
-        env_vars.insert("PKG_NAME".to_string(), Some(staging.name.clone()));
+        // A staging cache does not produce a package, so the PKG_* vars
+        // (which would otherwise carry the inheriting output's identity) are
+        // misleading here.
+        for key in ["PKG_NAME", "PKG_VERSION", "PKG_BUILDNUM", "PKG_BUILD_STRING", "PKG_HASH"] {
+            env_vars.remove(key);
+        }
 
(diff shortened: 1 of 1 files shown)


beenje:
This actually broke my recipe :-)

I'm using `PKG_VERSION` in the [build.sh script](https://github.com/beenje/cpptango-feedstock/blob/3edd75a474b6352146c5a1001d5b14c589806d81/recipe/build.sh#L23) to separate the debugging symbols in different files. This is done in the [staging output](https://github.com/beenje/cpptango-feedstock/blob/3edd75a474b6352146c5a1001d5b14c589806d81/recipe/recipe.yaml#L20).

Do you have a suggestion on how to do that instead?

wolfv:
You can use 

```yaml
script:
  env:
    MY_VERSION_VARIABLE: ${{ version }}  # or what you define in context
  file: build
```

And the env var should be set accordingly.

duncanmmacleod:
@wolfv, sorry to revive a closed thread (happy to repost in a new issue if that's more appropriate) but is it reasonable to add new `RECIPE_*` versions of the same variables so that the information from the `recipe` section of the YAML is automatically available in the build script? I guess (i.e. without evidence) that there are a lot of recipes that use the upstream package version in the build script for a multi-output build.

PR #2454, opened 2026-04-28
fix: skip unpacking symlinks on Windows instead of failing

pb01ka:
## Problem

Refer - https://github.com/prefix-dev/rattler-build/issues/940#issuecomment-4319905923

When rattler-build tries to extract a source archive (e.g. a `.tar.gz`) that contains symbolic links on Windows **without Developer Mode enabled**, the extraction fails hard with no actionable information:

```
 ╭─ Running build for recipe: podman-desktop-1.26.2-hf0b5a29_0
 │
 │ ╭─ Fetching source code
 │ │ Fetching source from url: https://github.com/podman-desktop/podman-desktop/archive/refs/tags/v1.26.2.tar.gz
 │ │ Downloading from: https://github.com/podman-desktop/podman-desktop/archive/refs/tags/v1.26.2.tar.gz
 │ │ ⚠ warning Failed to fetch from https://github.com/podman-desktop/podman-desktop/archive/refs/tags/v1.26.2.tar.gz: Failed to extract archive: Failed to extract tar
 │ │ ⚠ warning .gz: failed to unpack `\\?\C:\Users\Gagan\staged-recipes\build_artifacts\src_cache\.tmpyweool\podman-desktop-1.26.2\CLAUDE.md`
 │ │
 │ ╰─────────────────── (took 8 seconds)
 │
 ╰─────────────────── (took 8 seconds)
Error:   x Failed to extract archive: Failed to extract tar.gz: failed to unpack `\\?\C:\Users\Gagan\staged-recipes\build_artifacts\src_cache\.tmpyweool\podman-desktop-
  | 1.26.2\CLAUDE.md`
```

## Solution

Instead of failing the entire extraction when a symlink cannot be created, this PR iterates tar entries manually and **skips failing symlink entries with a warning** on Windows. The approach mirrors what rattler/pixi already do for their own extraction path.

1. **`unpack_tar_entries` helper** - replaces the single `archive.unpack(target)` call across all compression formats (`.tar.gz`, `.tar.bz2`, `.tar.xz`, `.tar.zst`, plain `.tar`) with a manual entry loop. On Windows, if unpacking a symlink entry fails, a `tracing::warn!` is emitted and extraction continues. Any other error still propagates as before.

2. **No registry access or new dependencies** - no `winreg`, no `windows-sys`, no `unsafe`. The fix is pure platform-conditional Rust.

3. **Test coverage** - a Windows-only test (`test_symlink_extraction_skipped_on_windows`) creates a minimal `.tar.gz` with a symlink entry and asserts that `extract_tar` succeeds regardless of Developer Mode status.

## Notes

- No behaviour change on Linux/macOS - the `#[cfg(target_os = "windows")]` block is a no-op on other platforms and all other errors still fail extraction.
- A missing symlink will only surface as a build error if the symlink is actually needed, giving users a much clearer failure point.

### DIFF
--- a/crates/rattler_build_source_cache/src/cache.rs
+++ b/crates/rattler_build_source_cache/src/cache.rs
@@ -717,7 +717,54 @@ pub fn is_tarball(file_name: &str) -> bool {
     .any(|ext| file_name.ends_with(ext))
 }
 
-fn extract_tar(archive: &Path, target: &Path) -> Result<(), CacheError> {
+/// Iterates over the entries of a tar archive and unpacks each one into `target`.
+///
+/// On Windows, symlink entries that fail to unpack are skipped with a warning instead of
+/// aborting the whole extraction. Creating symlinks on Windows requires Developer Mode or
+/// administrator privileges; skipping them gracefully means the rest of the archive is
+/// still extracted and the build can proceed (a missing symlink will surface as a build
+/// error only if it is actually needed).
+fn unpack_tar_entries<R: std::io::Read>(
+    archive: &mut tar::Archive<R>,
+    target: &Path,
+) -> Result<(), CacheError> {
+    let entries = archive.entries().map_err(|e| {
+        CacheError::ExtractionError(format!("Failed to read archive entries: {}", e))
+    })?;
+
+    for entry in entries {
+        let mut entry = entry.map_err(|e| {
+            CacheError::ExtractionError(format!("Failed to read archive entry: {}", e))
+        })?;
+
+        match entry.unpack_in(target) {
+            Ok(_) => {}
+            Err(e) => {
+                // On Windows, creating symlinks requires Developer Mode or admin
+                // privileges. If unpacking a symlink fails, warn and skip the entry
+                // rather than failing the whole extraction.
+                #[cfg(target_os = "windows")]
+                if entry.header().entry_type().is_symlink() {
+                    tracing::warn!(
+                        "skipping symlink in tar archive: {}: {e}",
+                        entry
+                            .path()
+                            .map(|p| p.display().to_string())
+                            .unwrap_or_default()
+                    );
+                    continue;
+                }
+                return Err(CacheError::ExtractionError(format!(
+                    "Failed to unpack archive entry: {e}"
+                )));
+            }
+        }
+    }
+
+    Ok(())
+}
+
+pub(crate) fn extract_tar(archive: &Path, target: &Path) -> Result<(), CacheError> {
     let file = fs_err::File::open(archive)
         .map_err(|e| CacheError::ExtractionError(format!("Failed to open archive: {}", e)))?;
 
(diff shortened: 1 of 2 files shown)


crates/rattler_build_source_cache/src/cache.rs
```diff
@@ -711,7 +711,41 @@ pub fn is_tarball(file_name: &str) -> bool {
...
+            .open_subkey(r"SOFTWARE\Microsoft\Windows\CurrentVersion\AppModelUnlock")?
+            .get_value("AllowDevelopmentWithoutDevLicense")?;
+        Ok(value == 1)
+    })()
+    .unwrap_or(false)
+}
+
+/// Wraps a tar unpack error, appending a Windows Developer Mode hint when on Windows and
+/// Developer Mode is detected as disabled.
+fn tar_unpack_error(context: &str, e: std::io::Error) -> CacheError {
```
baszalmstra: I assume this is also a specific io error that we are looking for? Permission denied or something?

baszalmstra:
Having a clear error is nice, but maybe even just continuing would be even nicer.

In rattler/pixi we try to unpack symlinks, and if they fail, we simply ignore them and issue a warning. I think that behavior would also be fine here. If it's a problem that they are missing, you will find out soon enough during the build itself.

Another option that we could pursue instead is to create a symlink, and if that fails, we simply materialize/create a copy of the file.

crates/rattler_build_source_cache/src/cache.rs
pb01ka: Not reliably. The error code when symlink creation fails on Windows with Developer Mode disabled is not consistent. `ERROR_PRIVILEGE_NOT_HELD` (1314) is the most common (as per my observation), but it can't reliably help in identifying a "symlink failed" error from the error alone, we instead gate the hint on the Developer Mode registry check: if Developer Mode is off and tar extraction fails for *any* reason, we surface the hint. The hint copy uses "may contain symbolic links" to reflect this uncertainty.

pb01ka:
@baszalmstra I think both alternatives are better UX than a hard failure.

1. **Warn and skip**

> In rattler/pixi we try to unpack symlinks, and if they fail, we simply ignore them and issue a warning. I think that behavior would also be fine here. If it's a problem that they are missing, you will find out soon enough during the build itself.

It is simpler to implement - the `tar` crate surfaces individual entry errors, so we could iterate entries manually instead of calling `unpack()`, and skip symlink entries that fail. The downside is the extracted tree is incomplete, which may cause silent build failures later.


2. **Fall back to copying the target file**

> Another option that we could pursue instead is to create a symlink, and if that fails, we simply materialize/create a copy of the file.


This approach is more robust - if the symlink target exists in the archive, we can materialise it as a regular file. The extracted tree is functionally complete even without actual symlinks.

I'd lean towards the **copy fallback** since it's transparent to the build, but it does require iterating tar entries manually and tracking regular files as they're extracted so we can copy them when a symlink to them is encountered. Happy to implement whichever approach you prefer.

pb01ka:
@baszalmstra Looking forward to your thoughts/opinions on my ideas above. 

Thank you. <3.

pb01ka:
Hi @baszalmstra. Waiting for your response. Thank you.

baszalmstra:
I like the copy fallback the most as well but it might be slightly tricky if there are symlinks to files that do not (yet) exist, because the order of extraction now matters. Id say we go with warn and skip first.

baszalmstra:
See this part in rattler where we implement this behavior for extraction:

https://github.com/conda/rattler/blob/f89ba4a5741af11e94547c8d6f0808ac2f8cb31c/crates/rattler_package_streaming/src/tokio/shared.rs#L18

pb01ka:
@baszalmstra I have implemented the warn and skip approach. Please take a look whenever you get a chance. Thank you.

pb01ka:
Hi @baszalmstra. I have addressed your previous reviews. Please let me know if any other change needs to be done.

Thank you.

pb01ka:
Hi @baszalmstra. A gentle reminder for reviewing my latest changes. Thank you.

PR #2456, opened 2026-04-30
perf: lazily initialize the `BaseClient`

baszalmstra:
This PR makes the initialization of the reqwest client more lazy. This ensure we only pay a startup cost when the client is actually used, and we avoid accessing the keychain until we actually need it. 

The downside is that `rattler_networking` is no longer an optional dependency, but I feel thats fine. 

Also removes a bunch of duplicate code.


wolfv (approved):
Thanks!

PR #2483, opened 2026-05-08
feat: compute and store content hash of extracted source archives

### LINKED ISSUE #261: Compute and store content hash
We are currently using the SHA256 of the tar archive. 
This is all fine, except that the SHA256 hash of the tar archive is dependent on how the archive was packed, and could change (without the contents being changed). For example, at some point Github changed the way they produced these archives and that changed all the hashes (that was reverted shortly after because it broke many package managers etc.).

However, we can also compute a content hash (a combination of all the hashes of all the files). This kind of hash is only dependent on the actual files in the archive. I believe Nix computes / uses such a hash. That would mean, at a later time, even if the hash of the .tar.gz archive doesn't match any more we could still verify the content hash, if we store it as part of the metadata.

pb01ka:
Closes https://github.com/prefix-dev/rattler-build/issues/261

### Motivation

The current source cache stores the SHA-256 of the downloaded tar archive. That hash is sensitive to *how* the archive was packed, not just what files it contains. This has caused breakage before (refer https://github.com/prefix-dev/rattler-build/issues/261): when GitHub changed their archive-generation pipeline, all hashes changed overnight even though the source code was identical.

A *content hash* - derived from the actual files inside the archive - is immune to this problem. Even if the upstream tarball is re-packed, the content hash stays the same as long as the files are unchanged. This is the approach used by conda-build.

### What this PR does

- Adds a new `content_hash.rs` module implementing `compute_content_hash`, which matches **conda-build's algorithm** for cross-tool compatibility:
  1. Walk the extracted directory recursively without following symlinks.
  2. Sort all entries lexicographically (backslashes normalised to `/`).
  3. Feed `relative_path | type_tag | payload | "-"` into a SHA-256 hasher for every entry, where the type tag is `L` (symlink), `D` (directory), or `F` (regular file), and text-file payloads have `\r\n` normalised to `\n` for cross-platform consistency.
  4. Return the lower-case hex digest.

- Adds a `content_sha256: Option<String>` field to `CacheEntry`.

- Exports `compute_content_hash` from the crate's public API.

### Tests

Seven unit tests cover the new module:

| Test | What it checks |
|---|---|
| `test_empty_directory` | Empty dir produces SHA-256 of zero bytes |
| `test_deterministic_across_calls` | Same directory hashes identically on repeated calls |
| `test_crlf_and_lf_are_equal` | `\r\n` and `\n` text files hash the same |
| `test_different_content_differs` | Different file content produces different hashes |
| `test_binary_file_not_crlf_normalised` | `\r\n` in binary files is **not** normalised |
| `test_binary_file_known_hash` | Binary file hash matches independently computed constant |
| `test_hash_extracted_archive_with_mixed_content` | End-to-end: tar.gz with dirs, text, binary, and symlinks; hash pinned to known-correct value |

The `test_cache_entry_persistence` integration test is updated to assert that `content_sha256`.

### Notes

- The content hash is stored as *additional* metadata alongside the existing archive checksum - no existing behaviour is changed.

### DIFF
--- a/crates/rattler_build_source_cache/src/cache.rs
+++ b/crates/rattler_build_source_cache/src/cache.rs
@@ -6,6 +6,7 @@ use tokio::io::AsyncWriteExt;
 
 use crate::{
     builder::ProgressHandler,
+    content_hash::compute_content_hash,
     error::CacheError,
     index::{CacheEntry, CacheIndex, SourceType},
     lock::LockManager,
@@ -153,6 +154,7 @@ impl SourceCache {
             created: chrono::Utc::now(),
             lock_file: Some(_lock.path().to_path_buf()),
             attestation_verified: false,
+            content_sha256: None,
         };
 
         self.index.insert(key, entry).await?;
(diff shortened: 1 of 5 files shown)


pb01ka:
@wolfv This PR is ready for review.

wolfv:
@jaimergp would you be able to have a look here as you implemented content hashes in conda-build?

pb01ka:
Note - I discussed with @jaimergp. Seems like https://github.com/conda/ceps/issues/150 is related to this PR. So I have taken it up. I will keep this thread updated. 

Thank you.

baszalmstra:
I would also prefer if you implement the algorithm itself in rattler and use it here.

baszalmstra:
Can be a seperate crate.

PR #2484, opened 2026-05-08
feat: `git.latest_tag_distance`

casperdcl:
Implements the equivalent of [conda-build's `GIT_DESCRIBE_NUMBER`](https://docs.conda.io/projects/conda-build/en/latest/user-guide/environment-variables.html#git-environment-variables)

### DIFF
--- a/crates/rattler_build_jinja/src/git.rs
+++ b/crates/rattler_build_jinja/src/git.rs
@@ -163,12 +163,44 @@ impl Git {
         Ok(Value::from(result))
     }
 
+    fn latest_tag_distance(&self, src: &str) -> Result<Value, minijinja::Error> {
+        if let Some(local_path) = self.resolve_local_path(src) {
+            let result = git_command_in_dir(&local_path, &["describe", "--tags", "--long"])?
+                .trim()
+                .splitn(3, '-')
+                .collect::<Vec<&str>>()[1]
+                .to_string();
+            return Ok(Value::from(result));
+        }
+
+        let tag_rev =
+            get_command_output("git", &["ls-remote", "--tags", "--sort=-v:refname", src])?
+                .lines()
+                .next()
+                .and_then(|s| s.split_ascii_whitespace().nth(0))
+                .ok_or_else(|| {
+                    minijinja::Error::new(
+                        minijinja::ErrorKind::InvalidOperation,
+                        "Failed to get the latest tag".to_string(),
+                    )
+                })?
+                .to_string();
+        let distance = get_command_output(
+            "git",
+            &["rev-list", "--count", &format!("{}..HEAD", tag_rev)],
+        )?
+        .trim()
+        .to_string();
+        Ok(Value::from(distance))
+    }
+
     fn dispatch(&self, name: &str, args: &[Value]) -> Result<Value, minijinja::Error> {
         let (src,) = from_args(args)?;
         match name {
             "head_rev" => self.head_rev(src),
             "latest_tag_rev" => self.latest_tag_rev(src),
             "latest_tag" => self.latest_tag(src),
+            "latest_tag_distance" => self.latest_tag_distance(src),
             name => Err(minijinja::Error::new(
                 minijinja::ErrorKind::UnknownMethod,
                 format!("object has no method named {name}"),
--- a/crates/rattler_build_jinja/src/jinja.rs
+++ b/crates/rattler_build_jinja/src/jinja.rs
@@ -1029,9 +1029,17 @@ mod tests {
                     .unwrap()
             );
             assert_eq!(
+                jinja
+                    .eval(&format!("git.latest_tag_distance({:?})", path))
+                    .expect("test 2")
+                    .as_str()
+                    .unwrap(),
+                "0"
+            );
+            assert_eq!(
                 jinja_wo_experimental
                     .eval(&format!("git.latest_tag({:?})", path))
-                    .expect_err("test 2")
+                    .expect_err("test 3")
                     .to_string(),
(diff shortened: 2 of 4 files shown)


wolfv:
Thanks! Looks good to me. Can you add a test?

casperdcl:
I think I need to get it building first :D
Apologies have just been desperately force-pushing to CI. Will try locally...

casperdcl:
ping?

crates/rattler_build_jinja/src/git.rs
```diff
@@ -163,12 +163,44 @@ impl Git {
...
 
+    fn latest_tag_distance(&self, src: &str) -> Result<Value, minijinja::Error> {
+        if let Some(local_path) = self.resolve_local_path(src) {
+            let result = git_command_in_dir(&local_path, &["describe", "--tags", "--long"])?
+                .trim()
+                .splitn(3, '-')
+                .collect::<Vec<&str>>()[1]
+                .to_string();
+            return Ok(Value::from(result));
+        }
```
pb01ka, line 174: I think this might be buggy for tag names with dashes. `git describe --tags --long` always emits `<tag>-<N>-g<hash>`, where `<N>` is the distance and `<hash>` is an abbreviated commit hash. The split must anchor from the **right**, not the left, because tag names routinely contain dashes (e.g. `v1.0-rc1`, `v2.3.0-beta.1`). Consider the following examples,

| Tag | `git describe` output | `splitn(3,'-')[1]` | Correct |
|---|---|---|---|
| `v0.1.0` | `v0.1.0-0-gabcdef` | `"0"` (correct) | `"0"` |
| `v1.0-rc1` | `v1.0-rc1-5-gabcdef` | `"rc1"` (incorrect) | `"5"` |
| `v0.18.0-1` | `v0.18.0-1-0-g0a09577fec` | `"1"` (integer but still incorrect) | `"0"` |

The fix is to use `rsplitn(3, '-')` and take index `[1]`,


```suggestion
            // rsplitn on "v1.0-rc1-5-gabcdef" -> ["gabcdef", "5", "v1.0-rc1"]
            let result = git_command_in_dir(&local_path, &["describe", "--tags", "--long"])?
                .trim()
                .rsplitn(3, '-')
                .collect::<Vec<&str>>()[1]
                .to_string();
            return Ok(Value::from(result));
        }
```

wolfv (approved):

wolfv:
Thank you!

pb01ka:
Hi. I was reviewing this PR and had some queries/suggestions? Can I still drop those in? The review session is in progress (not submitted yet from my end).

cc: @wolfv

crates/rattler_build_jinja/src/git.rs
```diff
@@ -163,12 +163,44 @@ impl Git {
...
+                    )
+                })?
+                .to_string();
+        let distance = get_command_output(
+            "git",
+            &["rev-list", "--count", &format!("{}..HEAD", tag_rev)],
+        )?
+        .trim()
+        .to_string();
+        Ok(Value::from(distance))
```
pb01ka, line 194: The remote path implementation resolves the latest tag's SHA via `ls-remote`, then calls, `git rev-list ...` command.

`git rev-list` operates on the **local** object store of whatever repo happens to be the shell's current working directory. It has no knowledge of the remote repo's commit graph unless those objects have already been fetched locally.

The common build scenario is building from a shallow clone or an unpacked `.tar.gz` source archive. In both cases the SHA returned by `ls-remote` is unknown to the local repo, so the command might fail, 

```
$ git rev-list --count 527eb7fd7953a1de..HEAD
fatal: ambiguous argument '527eb7fd7953a1de..HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

The other remote functions (`head_rev`, `latest_tag_rev`, `latest_tag`) only use `ls-remote`, which is a pure network operation - they never need local objects. `rev-list --count` cannot follow that pattern because counting commits inherently requires the full history.

casperdcl:
@pb01ka if you chose to open a follow-up PR, feel free to `@mention` me for review

crates/rattler_build_jinja/src/jinja.rs
```diff
@@ -1028,10 +1028,18 @@ mod tests {
...
                     .unwrap()
             );
+            assert_eq!(
+                jinja
+                    .eval(&format!("git.latest_tag_distance({:?})", path))
+                    .expect("test 2")
+                    .as_str()
+                    .unwrap(),
+                "0"
+            );
```
pb01ka, line 1038: I think this and the second assertion below, check that the distance is `"0"` - they are run immediately after tagging with no additional commits. 
I feel, a test that makes one or more commits after the tag and asserts `"1"` (or the exact count) would help in catching the `splitn` bug described above and give confidence that the counting logic works end-to-end.

wolfv:
Sure, feel free to comment here or make a follow up PR! Thank you !

pb01ka (commented):
Lastly, a minor,

The remote-path implementation duplicates (used thrice) the `ls-remote --tags --sort=-v:refname` call verbatim from `latest_tag_rev`. A small private helper would eliminate the duplication.

I know this PR is merged. However, if my suggestions look good then I can send a PR addressing these reviews. Please let me know if I can work on these.

I am happy to discuss more. Please feel free to correct me if I misinterpreted anything.

cc: @casperdcl @wolfv

casperdcl:
certainly looks worthy of a follow-up PR :)

pb01ka:
Sure. @casperdcl. I will send a follow up in couple of hours.

pb01ka:
The PR #2491 is opened as follow up for this. It is ready for review.

PR #2488, opened 2026-05-11
fix: respect run dependencies on sibling outputs for build ordering

esteve:
This PR fixes a regression in the topological sort that was introduced during the v0.58.0 refactor. When multiple recipes are built together, the dependency graph used for ordering builds only considered `pin_subpackage` references from run dependencies, but ignored plain `run:` specs. This caused solver failures when a package needed a plain runtime dependency that hadn't been built yet.

The fix `replaces extract_dependency_names` with `extract_dependency_names_with_sibling_context`, which also collects plain run/run_constraints dependencies whose names match other outputs in the build set.

This issue is causing ROS Humble to fail a full rebuild for the RoboStack project https://github.com/RoboStack/ros-humble/issues/400 (`ros-humble-uncrustify-vendor` which is a plain run dependency was scheduled to build after its consumer `ros-humble-ament-uncrustify`), with this fix I was able to do a full rebuild of ROS Humble.

I found this fix as part of an existing Claude session https://github.com/prefix-dev/rattler-build/commit/780ce8023c33a15463b06c62656f2a51025d0e97, I don't know what's the policy regarding AI-generated contributions, so I kept the author (Claude) and removed the skip filtering test as it's not part of this fix, but part of https://github.com/prefix-dev/rattler-build/commit/8f5cb03c9d43fa236c951d36a50374587176498c

### DIFF
--- a/crates/rattler_build_recipe/src/variant_render.rs
+++ b/crates/rattler_build_recipe/src/variant_render.rs
@@ -429,7 +429,8 @@ fn build_recipe_graph(
     for (idx, recipe) in recipes.iter().enumerate() {
         let current_name = &recipe.package.name;
 
-        for dep_name in extract_dependency_names(recipe) {
+        for dep_name in extract_dependency_names_with_sibling_context(recipe, Some(name_to_indices))
+        {
             if &dep_name == current_name {
                 continue;
             }
@@ -612,8 +613,11 @@ fn stable_topological_sort(
             let current_name = &variants[idx].recipe.package.name;
             let current_combination = &variants[idx].full_combination;
 
-            // Check all dependencies in requirements
-            for dep_name in extract_dependency_names(&variants[idx].recipe) {
+            // Check all dependencies in requirements (including run deps on sibling outputs)
+            for dep_name in extract_dependency_names_with_sibling_context(
+                &variants[idx].recipe,
+                Some(name_to_indices),
+            ) {
                 // Skip self-dependencies
                 if &dep_name == current_name {
                     continue;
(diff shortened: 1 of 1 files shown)


crates/rattler_build_recipe/src/variant_render.rs
```diff
@@ -1079,15 +1083,17 @@ fn finalize_build_string_single(
...
-///
-/// Note: May contain duplicates, which is acceptable for dependency graph construction.
-fn extract_dependency_names(recipe: &Stage1Recipe) -> Vec<rattler_conda_types::PackageName> {
+/// When `sibling_names` is provided, plain run/run_constraints dependencies whose names
+/// match sibling output names are included as build-order dependencies. This ensures that
+/// when output B has a run dependency on output A (e.g., `polars` depends on
+/// `polars-runtime-32`), A is built before B so that B's test environment can install A.
+fn extract_dependency_names_with_sibling_context(
+    recipe: &Stage1Recipe,
+    sibling_names: Option<&BTreeMap<rattler_conda_types::PackageName, Vec<usize>>>,
```
pb01ka:
```suggestion
    sibling_names: &BTreeMap<rattler_conda_types::PackageName, Vec<usize>>,
```

To the best of my understanding, there are two call sites for this new replacement,

1. `for dep_name in extract_dependency_names_with_sibling_context(recipe, Some(name_to_indices))`
2. `for dep_name in extract_dependency_names_with_sibling_context(&variants[idx].recipe, Some(name_to_indices))`

As far as I can tell, the `Vec::new()` branch isn't executed by both of these calls. Therefore, as of now, the `Option<>` wrapper adds cognitive overhead and makes it harder to see that sibling context is always required. In future if we ever need to make `sibling_names` optional we can, however for now let's make it compulsory.

pb01ka (commented):
Overall I found the fix correct and the tests also seem to be meaningful. 

The main actionable item from my end is removing the `Option<>` wrapper from `sibling_names` since the other path is never executed as of today. Hence, it would make the contract clearer and removes a branch that can never execute.

crates/rattler_build_recipe/src/variant_render.rs
esteve: Addressed in https://github.com/prefix-dev/rattler-build/pull/2488/commits/aee88d56e2e1d144926223a8529518c06e4f20ca

pb01ka (approved):
Submitting approval. 

@wolfv I think this can be merged.

Thank you.

esteve:
@pb01ka @wolfv I've rebased this branch on top of the latest main, is there anything else I should address? Thanks.

traversaro:
fyi @ruben-arts @baszalmstra

baszalmstra:
I dont think this is the correct approach. Let me whip up a better fix.

traversaro:
> I dont think this is the correct approach. Let me whip up a better fix.

Thanks!

esteve:
@baszalmstra thanks. Shall I close this PR? Or feel free to push here if you want, either is fine by me.

Hofer-Julian:
Thank you for your contribution, this PR is superseded by https://github.com/prefix-dev/rattler-build/pull/2796

PR #2490, opened 2026-05-11
fix: sanitize forbidden filename chars in build_string_prefix

hunger:
Characters that are invalid in file names on Windows or Unix (/ \ : * ? " < > | NUL) are now replaced with '_' when set via RenderConfig::with_build_string_prefix, so the resulting package file name is valid across platforms.


wolfv:
That should never pass build string validation though ... or do we not validate the build string?

hunger:
I looked into that a bit more: Rattler validates build strings, but only when validating MatchSpecs. When building packages, emojis and such pass through all the way into the package name written onto disk.

The latest iteration of this change adds a BuildStringPrefix type to rattler-build incl. validation. That is used in the clap contexts, so this should trigger early when users provide nonsense -- once pixi depends on this.

PR #2491, opened 2026-05-12
fix: correct `git.latest_tag_distance` parsing for dashed tag names

pb01ka:
Follow-up to https://github.com/prefix-dev/rattler-build/pull/2484.
Addresses, https://github.com/prefix-dev/rattler-build/pull/2484#pullrequestreview-4271687826, https://github.com/prefix-dev/rattler-build/pull/2484#discussion_r3225891184 and https://github.com/prefix-dev/rattler-build/pull/2484#discussion_r3226425675

## Problem

`git describe --tags --long` always emits output in the form `<tag>-<N>-g<hash>`, where `<N>` is the number of commits since the tag and `<hash>` is an abbreviated commit SHA. The original implementation extracted `<N>` using `splitn(3, '-')`, which splits from the **left**. This works for simple tag names like `v0.1.0`, but silently returns the wrong value for tag names that themselves contain dashes:

| Tag | `git describe` output | old result | correct |
|---|---|---|---|
| `v0.1.0` | `v0.1.0-0-gabcdef` | `"0"` (correct) | `"0"` |
| `v1.0-rc1` | `v1.0-rc1-5-gabcdef` | `"rc1"` (incorrect) | `"5"` |
| `v0.18.0-1` | `v0.18.0-1-0-g0a09577fec` | `"1"` (incorrect) | `"0"` |

The `v0.18.0-1` case is particularly subtle - the wrong value is still an integer, so the bug is invisible until a version comparison goes wrong.

### Bug fix

Switch from `splitn(3, '-')` to `rsplitn(3, '-')` in `latest_tag_distance`. Anchoring the split from the right is correct because the trailing `-<N>-g<hash>` suffix has a fixed structure regardless of how many dashes appear in the tag name.

### Refactor

Extract a private `latest_tag_sha_remote` helper that encapsulates the `git ls-remote --tags --sort=-v:refname` call. Both `latest_tag_rev` and `latest_tag_distance` duplicated this 8-line block verbatim; now they delegate to the helper instead.

### Tests

- **`eval_git_latest_tag_distance_nonzero`** - creates a repo, tags it, then adds one commit and asserts the distance returns `"1"`. This is the minimal test that would have caught the `splitn` bug, since the wrong field is extracted as soon as any commits exist after the tag.

- **`eval_git_latest_tag_distance_dashed_tags`** - exercises tags with dashes in their names (`v0.18.0-1`, `v1.0-rc1`, `v2.3.0-beta.1`) to prevent the regression from reappearing.

cc: @wolfv @casperdcl

### DIFF
--- a/crates/rattler_build_jinja/src/git.rs
+++ b/crates/rattler_build_jinja/src/git.rs
@@ -129,9 +129,6 @@ impl Git {
         Ok(Value::from(result))
     }
 
-    /// Returns the SHA of the most recent tag in `src` (a remote URL) by
-    /// querying `git ls-remote`. Used as a shared helper by `latest_tag_rev`,
-    /// `latest_tag`, and `latest_tag_distance`.
     fn latest_tag_sha_remote(&self, src: &str) -> Result<String, minijinja::Error> {
         get_command_output("git", &["ls-remote", "--tags", "--sort=-v:refname", src])?
             .lines()
@@ -172,9 +169,6 @@ impl Git {
 
     fn latest_tag_distance(&self, src: &str) -> Result<Value, minijinja::Error> {
         if let Some(local_path) = self.resolve_local_path(src) {
-            // rsplitn on "v1.0-rc1-5-gabcdef" -> ["gabcdef", "5", "v1.0-rc1"]
-            // Splitting from the right ensures tag names containing dashes
-            // (e.g. "v1.0-rc1", "v0.18.0-1") don't corrupt the distance field.
             let result = git_command_in_dir(&local_path, &["describe", "--tags", "--long"])?
                 .trim()
                 .rsplitn(3, '-')
--- a/crates/rattler_build_jinja/src/git.rs
+++ b/crates/rattler_build_jinja/src/git.rs
@@ -125,7 +125,15 @@ impl Git {
             return Ok(Value::from(result));
         }
 
-        let result = get_command_output("git", &["ls-remote", "--tags", "--sort=-v:refname", src])?
+        let result = self.latest_tag_sha_remote(src)?;
+        Ok(Value::from(result))
+    }
+
+    /// Returns the SHA of the most recent tag in `src` (a remote URL) by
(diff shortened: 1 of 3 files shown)


pb01ka:
This PR is ready for review. @wolfv @casperdcl

crates/rattler_build_jinja/src/git.rs
```diff
@@ -125,7 +125,15 @@ impl Git {
...
         }
 
-        let result = get_command_output("git", &["ls-remote", "--tags", "--sort=-v:refname", src])?
+        let result = self.latest_tag_sha_remote(src)?;
+        Ok(Value::from(result))
+    }
+
+    /// Returns the SHA of the most recent tag in `src` (a remote URL) by
+    /// querying `git ls-remote`. Used as a shared helper by `latest_tag_rev`,
+    /// `latest_tag`, and `latest_tag_distance`.
```
casperdcl:
```suggestion
```

crates/rattler_build_jinja/src/git.rs
```diff
@@ -165,26 +172,18 @@ impl Git {
 
     fn latest_tag_distance(&self, src: &str) -> Result<Value, minijinja::Error> {
         if let Some(local_path) = self.resolve_local_path(src) {
+            // rsplitn on "v1.0-rc1-5-gabcdef" -> ["gabcdef", "5", "v1.0-rc1"]
+            // Splitting from the right ensures tag names containing dashes
+            // (e.g. "v1.0-rc1", "v0.18.0-1") don't corrupt the distance field.
```
casperdcl:
```suggestion
```

casperdcl (approved):
good spot!

Personally don't really find the comments in the code useful; the function names are already self-explanatory, and edge cases are covered by the tests so don't need to be restated.

pb01ka:
@casperdcl I have addressed your review to remove the redundant code comments. 

Thanks for pointing out.

wolfv:
Thanks!

PR #2500, opened 2026-05-18
fix(py-rattler-build): Append hash string to specified ouput_dir when creating debug session

jsmolic:
Currently, the randomized string will be included in the temporary directory only if output_dir does not have a value, but we still want to have it included even if it's specified to have a unique directory for each debug attempt.

Note: originally reported in https://github.com/conda/conda-build/pull/5950#issuecomment-4475280119

### DIFF
--- a/py-rattler-build/rust/src/debug.rs
+++ b/py-rattler-build/rust/src/debug.rs
@@ -256,11 +256,10 @@ pub fn create_debug_session_py(
         .map(|tc| tc.inner)
         .unwrap_or_else(|| Configuration::builder().finish());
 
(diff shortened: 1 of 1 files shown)


wolfv:
Hey @jsmolic thanks for the PR! Do you really want this though? Isn't it better to be declarative in the API and use exactly the output dir? Should be easy to put somethiing "random" together from Python, right?

jsmolic:
Hey @wolfv, I thought of this as `rattler-build` CLI also does this when output directory is specified (though I didn't dig if that's handled in the rust API or the cli code), but if you want to keep the API unchanged that's fine with me too, we can come up with something in conda-build.

PR #2527, opened 2026-06-03
feat: Add brush interpreter support

baszalmstra:
Adds support for [`brush`](https://github.com/reubeno/brush), a bash-compatible shell written in Rust, as a build script interpreter.

Recipes can now opt in with `interpreter: brush` and an ordinary bash script will run under brush instead of the system bash. This works the same way on Linux, macOS, and Windows.

```yaml
build:
  script:
    interpreter: brush
    content: |
      echo "Hello from brush!"
requirements:
  build:
    - brush
```

`brush` is taken from the build environment, so it must be listed in `requirements/build`. A `brush` found on the system is intentionally not used, which keeps builds reproducible. I explicitly used this instead of embedding brush itself into rattler-build (which is also possible) because I think this approach makes the recipe more reproducible and if we ever run things inside a container in the future this will work out of the box (e.g. if the build script itself is not run on the host). 

## Notes

- Documentation is updated with a "Using `brush`" section.
- An end-to-end test which also runs on windows builds a brush-based recipe and was verified against conda-forge.

### DIFF
--- a/crates/rattler_build_script/src/interpreter/nushell.rs
+++ b/crates/rattler_build_script/src/interpreter/nushell.rs
@@ -9,7 +9,7 @@ use rattler_shell::{
 
 use crate::execution::{ExecutionArgs, run_process_with_replacements};
 
-use super::{Interpreter, InterpreterError, find_interpreter};
+use super::{Interpreter, InterpreterError, InterpreterSearchScope, find_interpreter};
 
 pub struct NuShellInterpreter;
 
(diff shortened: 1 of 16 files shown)


crates/rattler_build_script/src/interpreter/nushell.rs
```diff
@@ -155,6 +159,11 @@ impl Interpreter for NuShellInterpreter {
         build_prefix: Option<&PathBuf>,
         platform: &Platform,
     ) -> Result<Option<PathBuf>, which::Error> {
-        find_interpreter("nu", build_prefix, platform)
+        find_interpreter(
+            "nu",
+            build_prefix,
+            platform,
+            InterpreterSearchScope::PrefixThenSystemPath,
```
Hofer-Julian, line 166: Everything but bash and cmd should be only searched in the prefix I feel

baszalmstra: I agree but that would break the current behavior.

wolfv (approved):
Cool!

crates/rattler_build_script/src/interpreter/nushell.rs
baszalmstra: I think if we want to change the nushell support we should do the same thing as what we do with brush. Currently the implementation runs the activation and then directly invokes nushell. But with brush (and python) the normal bash activation runs and simply calls `brush script.sh` as part of the build instead. This has the advantage that the activation happens natively.

Hofer-Julian (approved):
Works like a charm. This bug I found, predates your PR: https://github.com/prefix-dev/rattler-build/issues/2530

PR #2528, opened 2026-06-04
chore: bump rattler crates to latest versions

baszalmstra:
## Description

Bumps the rattler crates to their latest versions. Major and minor updates are applied in `Cargo.toml`, while patch-level updates are picked up in the lockfile only.

Updated crates include rattler, rattler_conda_types, rattler_cache, rattler_config, rattler_networking, rattler_digest, and rattler_virtual_packages, along with patch bumps to the rest of the rattler family.

Keeping up with the latest releases also required a few adaptations:

- Raised the Rust toolchain to 1.95, since the new rattler set requires it.
- Migrated timestamp handling from chrono to jiff, matching the change in rattler_conda_types. The chrono dependency is now removed.
- Updated for the digest 0.11 upgrade pulled in by rattler_digest (bumped sha2 and md-5 to 0.11).
- Renamed the `experimental_extra_depends` field to `extra_depends` to match `IndexJson`.
- Fixed two clippy lints newly reported under the updated toolchain.


Hofer-Julian (approved):
Thanks!

PR #2535, opened 2026-06-08
refactor: streamline script interpreters through a native shell wrapper

baszalmstra:
## Description

Streamlines how build and test scripts run across all interpreters. Every script now runs through a native shell wrapper (bash on Unix, cmd.exe on Windows) that activates the build and host prefixes first and then invokes the chosen interpreter. This gives python, perl, ruby, node, R, nushell, brush and powershell the same activation behavior, instead of each interpreter handling activation in its own way.

The script crate's public interface was also tightened to a small, functional surface. The process environment variables and platform are now injected as a single value instead of read from global system state, which makes script generation deterministic and testable without serial_test workarounds.

## Behavior change

Interpreter resolution is unchanged for existing recipes: python, perl, ruby, node, rscript and nushell are still taken from the build or host prefix when present, and otherwise fall back to the system PATH. Declaring the interpreter in requirements.build is recommended for reproducible builds. brush is the exception and is always taken from the build environment.

Nushell now inherits its activated environment from the native wrapper like the other interpreters, rather than running its own activation.

## Testing

- Added unit coverage for interpreter resolution, environment isolation, activation, and the generated scripts.
- Existing interpreter end-to-end tests (python, ruby, perl, R, nushell, brush, powershell) still pass.

### DIFF
--- a/crates/rattler_build_script/src/interpreter/cmd_exe.rs
+++ b/crates/rattler_build_script/src/interpreter/cmd_exe.rs
@@ -14,9 +14,9 @@ impl InterpreterInvocation for CmdExeInvocation {
 
     fn search_scope(&self, build_platform: &Platform) -> InterpreterSearchScope {
         if build_platform.is_windows() {
-            InterpreterSearchScope::PrefixThenSystemPath
+            InterpreterSearchScope::build_and_host_with_system_fallback()
         } else {
-            InterpreterSearchScope::BuildPrefixOnly
+            InterpreterSearchScope::build_only()
         }
     }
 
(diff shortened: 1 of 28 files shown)


Hofer-Julian:
Awesome 😍

baszalmstra:
The language interpreters (ruby, python, perl, etc), now default to reading the interpreter from the build environment but fallback to using the system interpreter. I think it makes more sense to force users to put them in the build environment but I explicitly didnt change this because some tests relied on this behavior. 

However, I wonder what we want the intended behavior to be? @wolfv @Hofer-Julian Wdyt?

baszalmstra:
This code now doesnt look at the `host` prefix, only at `build` (because thats the platform where the script will run).

But that means that code like [this](https://github.com/conda-forge/jsonschema-specifications-feedstock/blob/d2b572e9b9c9992214acf689bd3ee7fa4382610d/recipe/recipe.yaml#L22):

```
build:
  number: 0
  noarch: python
  script:
    file: builder.py
    interpreter: python

requirements:
  host:
    - hatch-vcs
    - hatchling
    - pip
    - python ${{ python_min }}.*
    - python-build
  run:
    - python >=${{ python_min }}
    - referencing >=0.31.0
```
    
Are technically wrong, because `python` should be in the build section. 

What do we think about this? The previous code also looking in the host prefix, but thats not really correct as it could be for another platform. We can also implement it such that if `host_platform` equals `build_platform` that only then it looks at the host prefix as well? I dont like that its not very explicit and just adds to the confusion of what host and build mean.

Thougths? @wolfv @Hofer-Julian

Hofer-Julian:
I would be fine to technically break backwards compatibility by requiring the interpreter to be part of the build environment

baszalmstra:
I now opted to reuse the current behavior. Language specific interpreters are searched in the host and build prefix. I think we can think about alternatives later. Didnt want to break that now.

test/end-to-end/__snapshots__/test_tests.ambr
```diff
@@ -16,13 +16,9 @@
 # name: test_win_errorlevel_injection
   '''
   - script:
-      content: |-
-        echo "This test is designed to fail."
-        if %errorlevel% neq 0 exit /b %errorlevel%
```
wolfv, line 21: The errorlevel pieces are not injected anymore? That would be a loss :)

baszalmstra: They are just in a different place.

crates/rattler_build_script/src/interpreter/cmd_exe.rs
```diff
@@ -1,106 +1,63 @@
...
         }
+    }
 
-        Ok(())
+    fn extension(&self) -> &'static str {
+        "bat"
     }
 
-    async fn find_interpreter(
+    fn join_commands(&self, commands: &[String]) -> String {
```
baszalmstra, line 27: Here @wolfv

Hofer-Julian (approved):

Hofer-Julian:
Works as expected, and the code is now a bit cleaner

crates/rattler_build_script/src/interpreter/cmd_exe.rs
wolfv: OK!

PR #2539, opened 2026-06-09
refactor(script): compose the build wrapper from scoped sections

baszalmstra:
## What

Generalizes the build-script generator so the platform-native wrapper (`conda_build.sh` / `conda_build.bat`) is composed from an ordered list of sections rather than a single script body.

Each section is an isolated unit of work. It can carry its own interpreter, its own local `env`, and an optional label, and it runs in its own shell scope so its environment and shell state don't leak into later sections. A failure in any section aborts the whole wrapper.

Today exactly one section is built (from `ExecutionArgs`), so the generated build behavior is unchanged for real recipes. This is a structural refactor that makes the wrapper generator able to emit multiple scoped sections.

## Scope

- Internal only, no public API change.
- Single-section recipes keep their existing wrapper filenames and structure.
- Adds cross-platform tests for section scoping and composition.

### DIFF
--- a/crates/rattler_build_script/src/execution.rs
+++ b/crates/rattler_build_script/src/execution.rs
@@ -427,14 +427,51 @@ impl Decoder for CrLfNormalizer {
     }
 }
 
+/// One unit of a generated build wrapper: content run in a single interpreter,
+/// with optional step-local `env` and a boundary-comment label.
+///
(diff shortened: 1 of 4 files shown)


Hofer-Julian:
Is this the preparation for build tasks?

baszalmstra:
You caught me!

PR #2542, opened 2026-06-11
fix(ci): rattler-build binary detection

Hofer-Julian:
We moved to the Pixi Build `dev` feature to build rattler-build, but that also means we got more environment variables from the rust activation package: https://github.com/conda-forge/rust-activation-feedstock/blob/790b767c19ceb4caa6a5454141f835fec32ac227/recipe/activate.sh#L11

This PR adapts our test code so it still works

### DIFF
--- a/test/end-to-end/conftest.py
+++ b/test/end-to-end/conftest.py
@@ -20,8 +20,10 @@ def rattler_build():
         possible_paths = [
             base_path / f"target/release/{executable_name}",
             base_path / f"target/debug/{executable_name}",
-            base_path / f"target-pixi/release/{executable_name}",
-            base_path / f"target-pixi/debug/{executable_name}",
+            base_path
+            / f"target-pixi/{os.environ.get('CARGO_BUILD_TARGET')}/release/{executable_name}",
+            base_path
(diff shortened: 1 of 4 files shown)


test/end-to-end/conftest.py
```diff
@@ -20,8 +20,8 @@ def rattler_build():
         possible_paths = [
             base_path / f"target/release/{executable_name}",
             base_path / f"target/debug/{executable_name}",
-            base_path / f"target-pixi/release/{executable_name}",
-            base_path / f"target-pixi/debug/{executable_name}",
+            base_path / f"target-pixi/{os.environ.get('CARGO_BUILD_TARGET', 'release')}/{executable_name}",
+            base_path / f"target-pixi/{os.environ.get('CARGO_BUILD_TARGET', 'debug')}/{executable_name}",
```
baszalmstra: I think we should add `debug` after CARGO_BUILD_TARGET no?

Hofer-Julian: Ah yes

baszalmstra (approved):

PR #2545, opened 2026-06-12
fix: don't resolve the native wrapper shell as a build-env interpreter

wolfv:
A `file: build` script resolves to `build.bat` on Windows, whose `.bat` extension makes `infer_interpreter()` return `cmd`. Since #2535 that sent script generation down the specialized-interpreter path, which resolves a `cmd.exe` from the build prefix / host prefix / system PATH and re-invokes it. But `cmd` is the native wrapper shell itself (the wrapper is already run by `cmd.exe`), not a conda-provided executable, so the lookup failed with "interpreter 'cmd' was not found in the build environment".

When the resolved interpreter equals the native runner's own `default_interpreter()` (`cmd` on Windows, `bash` on Unix), inline the script body into the wrapper instead — the historical conda-build behavior — rather than resolving and re-invoking the shell.

## Alternative

Never search for `cmd.exe` in the build environment?

### DIFF
--- a/crates/rattler_build_script/src/execution.rs
+++ b/crates/rattler_build_script/src/execution.rs
@@ -430,8 +430,10 @@ impl Decoder for CrLfNormalizer {
 /// Returns the path to the generated native build wrapper script.
 ///
 /// If `args.interpreter` is set, or if the resolved script path has a known
-/// interpreter extension, inline script text is written to a separate file and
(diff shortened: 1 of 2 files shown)


baszalmstra (approved):

jschueller:
I got this problem too:
Error:   × interpreter `cmd` was not found in the build environment

is there a new release planned ?

wolfv:
Yes I'll cut a release today

Issue #2551, opened 2026-06-15
Refer to patches contained in other sources


jaimergp:
I'm trying to port `conda-standalone`'s `meta.yaml` to `recipe.yaml`. It features this sources section:

```yaml
source:
  - url: https://github.com/conda/conda-standalone/archive/{{ version }}.tar.gz
    sha256: 12323fb53979e00243b87d9391b2da4b8eb05b7a0abc988501dd719df41d7b29

  - url: https://github.com/conda/conda/archive/{{ conda_version }}.tar.gz
    sha256: 599defcc081883ee23656e146878abc2df344296e152c459a0ef05553c47e1c9
    folder: conda_src
    patches:
      - "{{ SRC_DIR | replace('\\', '/') }}/src/conda_patches/0001-Rename-and-replace-entrypoint-stub-exe.patch"
      - "{{ SRC_DIR | replace('\\', '/') }}/src/conda_patches/0002-Manipulate-PATH-directly-instead-of-_call_ing-conda.patch"
      - "{{ SRC_DIR | replace('\\', '/') }}/src/conda_patches/0003-Restrict-search-paths.patch"
      - "{{ SRC_DIR | replace('\\', '/') }}/src/conda_patches/0004-Fix-conda-run.patch"
```

`conda/conda-standalone` ships some patches to be applied on top of `conda/conda`. However, there doesn't seem to be a way to refer to `SRC_DIR` from the `patches` context:

```
Error:   × Failed to render recipe with variants
  ╰─▶ Jinja template error: Template rendering failed: undefined value (in
      <string>:1) (template: ${{ SRC_DIR }}/src/conda_patches/0001-Rename-and-
      replace-entrypoint-stub-exe.patch)
    ╭─[/Users/jrodriguez/devel/conda-standalone-feedstock/recipe/recipe.yaml:23:9]
 22 │     patches:
 23 │       - ${{ SRC_DIR }}/src/conda_patches/0001-Rename-and-replace-entrypoint-stub-exe.patch
    ·         ─┬─
    ·          ╰── Template rendering failed: undefined value (in <string>:1) (template: ${{ SRC_DIR }}/src/conda_patches/0001-Rename-and-replace-entrypoint-stub-exe.patch)
 24 │       - ${{ SRC_DIR }}/src/conda_patches/0002-Manipulate-PATH-directly-instead-of-_call_ing-conda.patch
    ╰────
```

[...]

PR #2554, opened 2026-06-15
feat: support late-bound build directory variables in patches and license files

wolfv:
## Summary

This PR introduces support for late-bound build directory variables (e.g., `${{ SRC_DIR }}`, `${{ PREFIX }}`) in recipe fields that are evaluated before build directories exist. This allows patches and license files to reference directories that are only known once the build has started.

This would fix issues we had for R packages (license files needing to come from the `$PREFIX`, and would allow patches to come from other source packages prefix-dev/rattler-build#2551).

* Late-bound variables are registered as self-referential Jinja globals (e.g., `SRC_DIR` → `"${{ SRC_DIR }}"`) so they render back to their token form instead of erroring as undefined
* Accessed variables are tracked for variant computation, but synthetic late-bound globals are excluded to prevent leakag
* Glob validation is deferred for late-bound entries since resolved paths may not be valid glob patterns
* The `split_glob_base()` helper separates concrete path prefixes from glob patterns to enable glob matching within resolved late-bound paths
* Test recipes demonstrate practical use cases: applying patches from one source to another, and collecting license files installed into `${{ PREFIX }}`

[https://claude.ai/code/session_01RenhdoLPhvaZmUYq8tu2VW](<https://claude.ai/code/session_01RenhdoLPhvaZmUYq8tu2VW>)


Hofer-Julian (approved):
Nice improvement!

PR #2562, opened 2026-06-18
fix(test): build llamacpp recipe with Ninja generator on Windows

Hofer-Julian:
The llamacpp test recipe ran 'cmake -B build' with only cmake in its build requirements, relying on a system Visual Studio toolchain. When the windows-latest runner image moved to Visual Studio 18 2026, cmake selected that generator but could not locate a compiler, failing test_git_source.

Force the Ninja generator and provision conda-forge c/cxx compilers so the build no longer depends on whichever Visual Studio version the runner ships.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -8,6 +8,7 @@ on:
       - "crates/**"
       - "Cargo.*"
       - "test/end-to-end/**"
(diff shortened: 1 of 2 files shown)


wolfv (approved):

wolfv:
Makes sense!

PR #2586, opened 2026-06-25
fix: switch PyPI sources to use `files.pythonhosted.org`

mgorny:
Fetch PyPI sources via `files.pythonhosted.org` directly rather than `pypi.org`.  This is the canonical location per upstream documentation: https://docs.pypi.org/api/#predictable-urls

It saves a single unnecessary redirect and reduces load off `pypi.org`, though it seems that currently both hostnames are served off the same server.  It also follows suit with what other distributions are doing, e.g. Arch Linux and Gentoo:
https://wiki.archlinux.org/title/Python_package_guidelines#Source https://gitweb.gentoo.org/repo/gentoo.git/tree/eclass/pypi.eclass#n173


wolfv (approved):

mgorny:
Thanks!

PR #2597, opened 2026-06-29
feat: improve CRAN recipe generation for R packages

wolfv:
## Summary

This PR significantly enhances the CRAN (R) recipe generation to produce more robust and complete recipes that better match conda-forge conventions. The changes address cross-compilation support, license file handling, build requirements, and package metadata.

## Key Changes

- **License file handling**: Refactored `map_license()` to return a vector of license files instead of an optional single file. Now automatically references license files bundled by `r-base` (e.g., `${{ PREFIX }}/lib/R/share/licenses/GPL-3`) via a new `r_bundled_license()` mapping function, with package-local files appended after bundled ones.

- **Cross-compilation support**: Added `CROSS_R_BASE_MARKER` placeholder that gets expanded by `format_cran_recipe_with_suggests()` into a conditional selector (`if: build_platform != target_platform`) that pulls in `cross-r-base` for cross-compilation scenarios.

- **Build script enhancement**: Updated `R CMD INSTALL` invocation to include `${R_ARGS}` variable, allowing recipe authors to pass extra flags (e.g., `--configure-args`) matching conda-forge conventions.

- **Compilation detection**: Improved logic to determine when a package needs compilation—now correctly identifies packages with `LinkingTo` dependencies as requiring a compiler even if `NeedsCompilation` is not explicitly set.

- **Dynamic linking configuration**: Added `dynamic_linking` field with `rpaths` for compiled packages to ensure the linker can find R's shared libraries (`lib/R/lib/` and `lib/`).

- **Architecture-independent packages**: Pure-R packages (those without compilation needs) are now marked with `noarch: generic`.

- **R version constraints**: Fixed handling of R version constraints from `Depends: R (>= x.y.z)` entries so they properly pin `r-base` in both `host` and `run` requirements without duplication.

- **Requirement organization**: Refactored requirement building to be more explicit and maintainable, with clearer separation of `r-base`, build tools, and dependency handling.

- The `r_bundled_license()` function maps SPDX identifiers to license file names shipped by `r-base`, with special handling for `-or-later` variants that map to base version files.
- License files are collected in order: bundled licenses first (deduplicated), then package-local files.
- The cross-compilation marker uses a plain identifier to ensure `serde_yaml` emits it unquoted for simple post-processing.
- Build requirements are now only added when `needs_compilation` is true, keeping pure-R recipes minimal.

https://claude.ai/code/session_01M9MFHMC3psboGoePGiyWPC

### DIFF
--- a/docs/snippets/recipes/r-dplyr-generated.yaml
+++ b/docs/snippets/recipes/r-dplyr-generated.yaml
@@ -1,58 +1,58 @@
 context:
-  build_number: '0'
+  build_number: "0"
 
 package:
   name: r-dplyr
   version: 1.2.1
 
 source:
-- url:
-  - https://cran.r-project.org/src/contrib/dplyr_1.2.1.tar.gz
-  - https://cran.r-project.org/src/contrib/Archive/dplyr_1.2.1.tar.gz
-  sha256: 18d66fe2f24cda8b619c177326d39dd015229ab99695fa67d3fa31e4569390e0
+  - url:
+      - https://cran.r-project.org/src/contrib/dplyr_1.2.1.tar.gz
+      - https://cran.r-project.org/src/contrib/Archive/dplyr_1.2.1.tar.gz
+    sha256: 18d66fe2f24cda8b619c177326d39dd015229ab99695fa67d3fa31e4569390e0
 
 build:
   number: ${{ build_number }}
   script: R CMD INSTALL --build . ${R_ARGS}
   dynamic_linking:
     rpaths:
-    - lib/R/lib/
-    - lib/
+      - lib/R/lib/
+      - lib/
 
 requirements:
   build:
-  - if: build_platform != target_platform
-    then:
-      - cross-r-base ${{ r_base }}
-  - ${{ compiler('c') }}
-  - ${{ compiler('cxx') }}
-  - make
+    - if: build_platform != target_platform
+      then:
+        - cross-r-base ${{ r_base }}
+    - ${{ compiler('c') }}
+    - ${{ compiler('cxx') }}
+    - make
   host:
-  - r-base >=4.1.0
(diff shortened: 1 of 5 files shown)


docs/snippets/recipes/r-dplyr-generated.yaml
```diff
@@ -1,60 +1,87 @@
...
+  - url:
+      - https://cran.r-project.org/src/contrib/dplyr_1.2.1.tar.gz
+      - https://cran.r-project.org/src/contrib/Archive/dplyr_1.2.1.tar.gz
+    sha256: 18d66fe2f24cda8b619c177326d39dd015229ab99695fa67d3fa31e4569390e0
 
 build:
-  script: R CMD INSTALL --build .
-  python: {}
+  number: ${{ build_number }}
+  script: R CMD INSTALL --build . ${R_ARGS}
```
isuruf, line 16: For windows we have to use `%R_ARGS%`

docs/snippets/recipes/r-dplyr-generated.yaml
```diff
@@ -1,60 +1,87 @@
...
+  build:
+    - if: build_platform != target_platform
+      then:
+        - cross-r-base ${{ r_base }}
+    - ${{ compiler('c') }}
+    - ${{ compiler('cxx') }}
+    - make
   host:
-    - r-base >=3.5.0
+    - r-base >=4.1.0
```
isuruf, line 31: Use this
```suggestion
    - r-base
```
 and add a skip like `skip: match(r_base, "<4.1")`

wolfv:
Thanks @isuruf! I'll try to take a look tomorrow. Btw. you can now also try the recipe generation in the WASM playground (https://playground.rattler.build/)

Issue #2598, opened 2026-06-29
No license file were copied error when overriding `license_file`


pavelzw:
```
outputs:
  - package:
      name: py_rattler
      version: 0.1.0
    build:
      noarch: generic
    about:
      license_file: []

about:
  license_file:
    - LICENSE
```

```
❯ rattler-build build -r recipe/
 Processing recipe at path: /private/var/folders/rz/q1r4tv0n6ll38q18fbk419mc0000gn/T/tmp.9o3SIOUrpD/recipe/recipe.yaml

 ╭─ Finding outputs from recipe
 │ Found 1 variants
 │
 │ Build variant: py_rattler-0.1.0-h4616a5c_0
 │
 │ ╭─────────────────┬──────────╮
 │ │ Variant         ┆ Version  │
 │ ╞═════════════════╪══════════╡
 │ │ target_platform ┆ "noarch" │
 │ ╰─────────────────┴──────────╯
 │
 ╰─────────────────── (took 0 seconds)
 Starting build of 1 outputs

 ╭─ Running build for recipe: py_rattler-0.1.0-h4616a5c_0
 │
 │ ╭─ Fetching source code
 │ │ No sources to fetch
 │ │
 │ ╰─────────────────── (took 0 seconds)
 │
 │ Finalized run dependencies: this output has no run dependencies
 │
 │ ╭─ Installing build environment
 │ │ ✔ Successfully updated the build environment
 │ │
 │ ╰─────────────────── (took 0 seconds)
 │
 │ ╭─ Installing host environment
 │ │ ✔ Successfully updated the host environment
 │ │
 │ ╰─────────────────── (took 0 seconds)
 │
 │ ╭─ Packaging new files
 │ │ Copying done!
 │ │ Post-processing done!
 │ │ Writing test files
 │ │ Writing metadata for package
 │ │ Copying license files
 │ │
 │ ╰─────────────────── (took 0 seconds)
 │
 ╰─────────────────── (took 0 seconds)
Error:   × No license files were copied
```

tested with rattler-build 0.67.0, https://github.com/conda-forge/py-rattler-feedstock/pull/85

PR #2600, opened 2026-06-29
fix: preserve explicit empty license_file in output packages

wolfv:
## Summary
Fix handling of explicit empty `license_file: []` declarations in output packages so they properly override inherited top-level license files instead of falling back to them.

## Key Changes
- Modified `Stage0About` evaluation to distinguish between "key absent" (inherit top-level value) and "key explicitly set to empty" (override with empty set)
- When `license_file` evaluates to no globs and no late-bound entries, explicitly set `license_file_globs` to an empty `GlobVec` instead of leaving it as `None`
- Added regression test `test_output_can_override_license_file_with_empty_list` to verify the fix

## Implementation Details
The fix addresses issue #2598 where outputs with `license_file: []` were incorrectly inheriting the top-level `license_file` value during merge. By converting the `None` result to `Some(GlobVec::default())` when the key is explicitly present but empty, the merge logic can now properly distinguish between:
- **Absent key**: `None` → inherit top-level value
- **Explicit empty list**: `Some(empty)` → override and copy no license files

The test validates both scenarios: an output that explicitly clears license files and another that inherits them.

https://claude.ai/code/session_01QCuC5RWsV2guGNDbjWNWST

### DIFF
--- a/crates/rattler_build_recipe/src/stage0/evaluate.rs
+++ b/crates/rattler_build_recipe/src/stage0/evaluate.rs
@@ -1616,7 +1616,20 @@ impl Evaluate for Stage0About {
             license_file_globs,
             license_file_late_bound,
         } = match self.license_file.as_ref() {
-            Some(lf) => evaluate_license_files(lf, context)?,
+            Some(lf) => {
+                let mut files = evaluate_license_files(lf, context)?;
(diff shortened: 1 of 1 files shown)


pavelzw:
ah i guess you were faster... #2605

PR #2607, opened 2026-06-30
fix: follow symlinks when copying license files

wolfv:
## Summary
This change adds support for dereferencing symlinks when copying license files, ensuring that symlinked licenses are packaged as actual content rather than potentially dangling symlinks.

## Key Changes
- Added `dereference_symlinks` field to `CopyDir` struct with a new builder method to enable the behavior
- Modified the symlink handling logic in `CopyDir::run()` to skip symlink-specific processing when dereferencing is enabled, allowing the standard file/directory copy logic to follow and materialize symlink targets
- Updated `copy_license_files()` in `packaging.rs` to enable symlink dereferencing for all three license file copy operations (work directory, recipe directory, and source directories)
- Added comprehensive tests covering:
  - Successful dereferencing of a symlinked file with relative path resolution
  - Error handling for broken symlinks (now surfaces as a clear copy error instead of silently creating a dangling symlink)
- Updated CHANGELOG.md documenting the fix

## Implementation Details
When `dereference_symlinks(true)` is set, symlinks are no longer copied as symlinks. Instead, the code falls through to the regular file/directory branches which use `is_dir()` and `reflink_or_copy()` - both of which follow symlinks. This approach:
- Materializes symlinked files as real content in the package
- Handles relative symlinks correctly (e.g., `../LICENSE`)
- Surfaces broken symlinks as clear copy errors rather than silently creating dangling symlinks
- Works consistently across Unix and Windows platforms

The feature is disabled by default and only enabled for license file copying operations.

https://claude.ai/code/session_01Fcmrv6nQdGYURcZ3vB5vyh


pavelzw:
> Hit this too. I think a simpler (safer?) solution would be to reject symlinks entirely, and require the recipe author to specify the path to the actual file.

by @mgorny

but i agree

PR #2608, opened 2026-06-30
feat: support `source.filter` for URL and archive path sources

wolfv:
## Summary
This PR extends the `source.filter` field to support `url` sources and `path` sources that point to archives. Previously, filtering was only available for git sources. The filter is now applied to the contents of extracted archives, allowing large source tarballs to be trimmed down to only the parts needed for the build.

## Key Changes

- **Recipe Schema Updates**: Added `filter` field to `UrlSource` in both stage0 and stage1 recipe structures with proper serialization handling
- **Parser Enhancement**: Extended the URL source parser to recognize and parse the `filter` field alongside existing fields like `url`, `sha256`, and `patches`
- **Source Extraction Logic**: Modified `copy_from_cache()` to accept and apply glob filters when copying extracted archive contents to the destination directory
- **Path Source Archives**: When a `path` source points to an archive file, it's now converted to a temporary URL source with the filter applied during extraction
- **Variable Evaluation**: Added filter evaluation in the stage0 evaluation logic to support variable interpolation within glob patterns
- **Documentation**: Added comprehensive examples in `recipe_file.md` showing both simple list-based and include/exclude mapping-based filter syntax for URL sources
- **Tests**: Added unit tests for parsing URL sources with filters and an end-to-end test with a real archive demonstrating filter functionality

## Implementation Details

- The filter is applied at the extraction stage using the existing `copy_dir::CopyDir` builder with the new `with_globvec()` method
- Filters support both simple list syntax (`["cmake/*", "clang/*"]`) and structured include/exclude mappings
- The implementation reuses the existing `IncludeExclude` and `GlobVec` types already used for git source filtering
- Snapshot tests were updated to reflect the new `filter: []` field in serialized output

https://claude.ai/code/session_01RxSDwdYTY8h7T8VRHR5wmD


wolfv:
I also submitted a CEP update: https://github.com/conda/ceps/pull/176

wolfv:
Thanks for the pointer to [conda/ceps#176](https://github.com/conda/ceps/pull/176). Since CEP-0014 specifies `filter` on `url`, `path`, **and** `git` sources, I've pushed a follow-up commit (41f28d5) adding git source support too — the filter is applied to the checked-out tree, mirroring the url/path behavior. Parser tests and an end-to-end check against a local git repo confirm it works. The PR now covers all three source types from the CEP.

---
_Generated by [Claude Code](https://claude.ai/code)_

h-vetinari:
Super nice! 🤩

Issue #2563, opened 2026-06-18
feat(generate-recipe): close remaining gaps in R recipe generation


pb01ka:
> I may have made mistakes in my research - please correct me if anything below is wrong or incomplete.

## Background

`rattler-build generate-recipe` is the only R recipe generator in the conda ecosystem that works today, targets the v1 `recipe.yaml` format, and uses a reliable data source. The two alternatives have both effectively stalled:

- **`conda skeleton cran`** - broken in all recent conda-build releases; requires a downgrade to v3.27.0; users are being directed away from it; refer - https://github.com/bgruening/conda_r_skeleton_helper/issues/58 and specifically, https://github.com/bgruening/conda_r_skeleton_helper/issues/58#issuecomment-2899609421. Also note the correction pointed out by @isuruf https://github.com/prefix-dev/rattler-build/issues/2563#issuecomment-4755758238,

Quoting the correction,

> Yes, that's a misinterpretation. 99.9% of R packages are in CRAN so R packages in github is a niche problem. conda skeleton cran is still the best way to generate R recipes.

- **`grayskull`** - active, unresolved CRAN bugs ([#616](https://github.com/conda/grayskull/issues/616), [#615](https://github.com/conda/grayskull/issues/615)). AFAIK, it generates v0 `meta.yaml` by default; an extra `conda-recipe-manager` conversion step is required for v1 output. Scrapes CRAN DESCRIPTION files directly - fragile and CRAN-only. ([#532](https://github.com/conda/grayskull/issues/532))
- **`conda_r_skeleton_helper` (bgruening)** - a post-processor only; calls `conda skeleton cran` internally and inherits all of its brokenness.

`rattler-build generate-recipe` already has the right foundations:

- Queries the [R-universe JSON API](https://r-universe.dev) - structured, reliable, and covers CRAN, Bioconductor, and GitHub-hosted R packages via `--universe`. ([cran.rs:build_cran_recipe_and_deps()](https://github.com/prefix-dev/rattler-build/blob/1424726689ba43338d8b9b827630ec5088ebd737/crates/rattler_build_recipe_generator/src/cran.rs#L229))
[...]

PR #2622, opened 2026-07-02
feat: warn about R dependencies missing from conda-forge in generated recipes

pb01ka:
Refer https://github.com/prefix-dev/rattler-build/issues/2563

**⚠️ Breaking:** `generate-recipe cran --tree` now recurses only into dependencies missing from conda-forge by default, instead of the whole dependency tree. Pass the new `--full` flag alongside `--tree` to restore the old behavior.

## Summary

`cran.rs` already collected an R package's dependencies into a `remaining_deps` set while generating a recipe, but the set was discarded after use - there was no signal to the user about which of those dependencies weren't yet packaged on conda-forge. The recipe would only fail much later, at solve time, with no clear diagnostic pointing back to the missing R package.

This PR checks each collected dependency against conda-forge right after the recipe is generated, and logs a warning for any that are missing. The check goes through `rattler_repodata_gateway`'s `Gateway`, with every dependency batched into a single query rather than one request per package. This mirrors the existing behavior discussed for `pypi.rs`, minus the name-mapping step: CRAN package `foo` always maps to conda-forge `r-foo` by convention, so no lookup table is needed.

## Important Points

- Wired `warn_about_missing_conda_forge_deps` into `generate_r_recipe`, run once per generated recipe.
- Network errors are treated as "exists" (fail open) so a flaky connection doesn't block recipe generation or spam false warnings.
- Added unit tests, including a deterministic one for the network-failure branch ([using the RFC 2606 `.invalid` TLD, which is guaranteed never to resolve](https://datatracker.ietf.org/doc/html/rfc2606)).
- Changed `--tree`'s behavior: it now recurses only into dependencies missing from conda-forge by default (matching grayskull's `--recursive`), instead of blindly recursing into the whole dependency tree. (refer https://github.com/prefix-dev/rattler-build/pull/2622#issuecomment-5033337593) Generating boilerplate recipes for packages already available on conda-forge isn't useful, since they'd need further changes to be valid anyway.
- Added a `--full`/`-f` modifier to restore the old behavior when it's genuinely wanted (e.g. building a self-contained "forge" channel that doesn't want to depend on conda-forge for anything in the tree): `--tree --full` recurses into every dependency regardless of conda-forge status. Has no effect without `--tree`.
- This started out as a separate opt-in `--recursive` flag alongside the untouched `--tree`, but https://github.com/prefix-dev/rattler-build/pull/2622#issuecomment-5006275075 and https://github.com/prefix-dev/rattler-build/pull/2622#issuecomment-5033337593 pointed out the two were confusingly similar at a glance, so `--recursive` was folded into `--tree` (with `--full` as the escape hatch) instead of shipping both.

### DIFF
--- a/crates/rattler_build_recipe_generator/src/cran.rs
+++ b/crates/rattler_build_recipe_generator/src/cran.rs
@@ -912,7 +912,14 @@ mod tests {
     /// Build a `CranOpts` configured to check the given channel names via
     /// `--skip-available`.
     fn opts_with_skip_available(channels: &[&str]) -> CranOpts {
+        opts_with_tree_and_skip_available(false, channels)
+    }
+
+    /// Build a `CranOpts` with `tree` set as given, configured to check the
+    /// given channel names via `--skip-available`.
+    fn opts_with_tree_and_skip_available(tree: bool, channels: &[&str]) -> CranOpts {
         CranOpts {
+            tree,
             skip_available: channels.iter().map(|c| c.parse().unwrap()).collect(),
             ..dummy_cran_opts()
         }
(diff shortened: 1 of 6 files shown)


pb01ka:
This PR is ready for review. @wolfv Please feel free to take a look.

@isuruf You might be interested as well. 

TY.

isuruf:
> Should we also recurse specifically for the dependencies this PR flags as missing from conda-forge?

Yes, but not by default.

pb01ka:
Noted. Will update this PR tomorrow using my best judgement.

pb01ka:
> Yes, but not by default.

Addressed using my best possible judgement. I have added a `--recursive` flag along side `--tree` flag. Once my changes are reviewed, I would like some thoughts on the following question I have in my mind.

**Question: should `--recursive` just be folded into `--tree`?**

Now that both flags exist side by side, I want to sanity-check the design. At a high level `--recursive` and `--tree` look very similar - both walk the dependency graph and generate recipes for what they find. The only real difference is *which* dependencies they recurse into:

- `--tree` recurses into **every** dependency (**am I right?**), regardless of whether it's already on conda-forge.
- `--recursive` recurses only into the dependencies that are **missing from conda-forge**.

Given how close they are on the surface, would you prefer we:

1. Keep them as two separate flags (current state), or
2. Merge them into one - e.g. drop `--recursive` and instead have `--tree` skip dependencies already available on conda-forge (or add a modifier to `--tree` that toggles that behavior)?

I'm happy either way; just want to avoid shipping two options that most users will find hard to tell apart (including me :laughing:).

jsmolic:
>     * `--tree` recurses into **every** dependency (**am I right?**), regardless of whether it's already on conda-forge.
> 

I dont see this option in grayskull or conda-skeleton, only `--recursive`.


>     2. Merge them into one - e.g. drop `--recursive` and instead have `--tree` skip dependencies already available on conda-forge (or add a modifier to `--tree` that toggles that behavior)?

I think this makes most sense, having generated recipes for already available packages doesn't seem useful (especially given that boilerplate recipes may need further changes to be valid)

pb01ka:
@jsmolic I used my best possible judgement to implement your suggestion,

- `--tree` now **by default** recurses only into the dependencies that are **missing from conda-forge**.
- `--tree --full` recurses into **every dependency**, regardless of whether it's already on conda-forge. (the older default behaviour of `--tree`)

The [PR description](https://github.com/prefix-dev/rattler-build/pull/2622#issue-4797491398) is also updated to reflect the same. TY. 

Please let me know if anything else needs to be done.

pb01ka:
@wolfv Kindly take a look. This PR is passing the tests and waiting for reviews/decision. TY.

crates/rattler_build_recipe_generator/src/cran.rs
```diff
@@ -248,6 +257,66 @@ pub async fn fetch_package_sha256sum(url: &Url) -> Result<Sha256Hash, miette::Er
...
+        }
+    }
+}
+
+/// Check whether `conda_name` exists on the `conda-forge` channel.
+///
+/// Unlike PyPI, CRAN package names map to conda-forge by a fixed convention
+/// (`foo` -> `r-foo`, see [`format_r_package`]), so no name-mapping lookup is
+/// needed here, only an existence check via the anaconda.org API.
+async fn conda_forge_package_exists(conda_name: &str) -> bool {
```
wolfv: I think we should use the Repodata Gateway here and query conda-forge that way. Is it possible for you to try that?

wolfv:
Looks good except that one comment!

crates/rattler_build_recipe_generator/src/cran.rs
pb01ka: I implemented the repodata gateway approach using my best judgement. Please take a look and let me know your feedback. TY.

pb01ka:
@wolfv I addressed your comment to the best of my abilities and understanding. The tests are also passing. Please take a look. TY.

wolfv:
I think we should maybe flip this around. rattler-build is not a conda-forge specific tool. So instead of hard-coding conda-forge it might be better to have a flag that says: `--skip-available=conda-forge` (or any other channel) and then the recipe generator can skip over those. What do you think?

pb01ka:
> What do you think?

I agree @wolfv. Please note that, I have used `--skip-available` instead of `--skip-existing`. I hope that works. The tests are passing with my latest changes. TY.

pb01ka:
@wolfv A gentle reminder for [my above update](https://github.com/prefix-dev/rattler-build/pull/2622#issuecomment-5308394667). TY.

pb01ka:
@wolfv A gentle reminder on this. Review requested via Github. Please also see my comment [above](https://github.com/prefix-dev/rattler-build/pull/2622#issuecomment-5308394667). TY.

PR #2636, opened 2026-07-05
feat: add tool-specific config extension and default config discovery

wolfv:
## Description

Part of the effort to make rattler, rattler-build and pixi share one configuration (see conda/rattler#2557, now released as `rattler_config` 0.6.0). rattler-build already consumed `rattler_config`, but only as `ConfigBase<()>` (no extension point) and only when `--config-file` was passed explicitly — without it, no configuration was loaded at all.

Two changes, plus a dependency bump:

1. **Tool-specific extension point.** New `src/config.rs` defines `RattlerBuildConfig` — an (empty for now) extension struct implementing the `rattler_config::config::Config` trait (in 0.6 that's just `merge_config`) — and the alias `Config = ConfigBase<RattlerBuildConfig>`, used everywhere `ConfigBase<()>` was (opt.rs, main.rs, debug.rs, Python bindings). Future rattler-build-only configuration keys go into this struct; all shared keys (default channels, mirrors, S3 options, concurrency, …) keep coming from `rattler_config`.

2. **Default config discovery via the shared `locations` helper.** When `--config-file` is *not* given, rattler-build now discovers and merges configuration from the standard locations (only files that exist, later overriding earlier) using `rattler_config::locations::config_search_paths(&["pixi", "rattler-build"])` — the shared discovery shipped in 0.6. This replaces ~80 lines of hand-rolled path logic that previously duplicated pixi's. Precedence (lowest first): all **system-wide** files (`/etc/pixi/config.toml`, `/etc/rattler-build/config.toml`; `C:\ProgramData\…` on Windows), then all **per-user** files (platform config dir, then `$PIXI_HOME`/`$RATTLER_BUILD_HOME` or `~/.pixi` / `~/.rattler-build`) — pixi first, then rattler-build, so a user's existing pixi config (mirrors, S3, channels) applies out of the box and rattler-build's own config wins. `--config-file` still bypasses discovery entirely (loads only that file), and when no file exists the loader returns "no config" exactly as before.

   Note: this newly searches a system-wide `/etc/rattler-build/config.toml`, and the ordering is now system-before-user across both tools (previously it was all-pixi-then-all-rattler-build) — a deliberate consequence of using the shared helper.

### Dependency bump

All rattler crates bumped to the released set that includes `rattler_config` 0.6: rattler 0.46→0.47, rattler_conda_types 0.47→0.48, rattler_config 0.5.2→0.6, rattler_repodata_gateway 0.29→0.30, and the rest (cache, index, menuinst, networking, package_streaming, shell, solve, upload, virtual_packages) to their current releases. Non-config fallout from the bump, all behavior-preserving:
- `Gateway::query` now returns `RepoDataQueryOutput` (repodata_gateway 0.30) — `solver.rs` reads `result.repodata`.
- Shared `ConfigBase` keys moved behind `.common` (Deref) — one access in the Python bindings adjusted.
- `Channel::from_directory` was deprecated in conda_types 0.48 — 3 call sites migrated to `try_from_directory(..)`, preserving the prior panic-on-invalid-path behavior.

`cargo tree` confirms a single `rattler_conda_types 0.48.0` and `rattler_config 0.6.0`, both from crates.io. `docs/config.md` and the CLI reference were updated.

> **This PR unblocks pixi.** pixi (prefix-dev/pixi#6531) consumes `rattler_build_core` as a library and currently can't move to the released rattler 0.6 line because the published `rattler_build_core` (0.2.8) still pins rattler 0.46 / rattler_config 0.5.2. Merging **and publishing** a `rattler_build_core` release built on this branch is what lets pixi drop its temporary git patch and switch to the crates.io versions.

## Testing

- `cargo check --workspace --all-targets` clean (zero warnings); `cargo check -p py-rattler-build` clean.
- Config module tests pass (`cargo test -p rattler-build --lib`), including the discovery/ordering tests (rewritten to assert per-group precedence via the shared `locations` helpers, using temp dirs / env overrides, not the real home dir).
- `cargo fmt --check` and `cargo clippy` (`-D warnings`) clean on both workspaces.

## AI Disclosure

- [x] This PR contains AI-generated content.
  - [x] I have tested any AI-generated content in my PR.
  - [x] I take responsibility for any AI-generated content in my PR.

Tools: Claude Code

https://claude.ai/code/session_01EzYs4p6XJDj9QBcNB4ipR7

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,7 @@ async-trait = "0.1"
 bzip2 = "0.6.1"
 clap = { version = "4.6.0", features = ["derive"] }
 dashmap = "6.1"
+dirs = "6.0.0"
 dunce = "1.0.5"
 console = { version = "0.16.3", features = ["windows-console-colors"] }
 flate2 = "1.1.9"
@@ -133,39 +134,38 @@ tracing-subscriber = { version = "0.3", features = [
 ] }
 tracing-test = "0.2.6"
 
-# rattler crates
 # Rattler crates
-rattler_conda_types = { version = "0.47.2", default-features = false }
+rattler_conda_types = { version = "0.48", default-features = false }
 rattler_digest = { version = "1.3.1", default-features = false, features = [
   "serde",
 ] }
-rattler_networking = { version = "0.30.0", default-features = false }
-rattler_package_streaming = { version = "0.26.5", default-features = false }
-rattler_shell = { version = "0.27.7", default-features = false, features = [
+rattler_networking = { version = "0.30.1", default-features = false }
+rattler_package_streaming = { version = "0.26.7", default-features = false }
+rattler_shell = { version = "0.27.9", default-features = false, features = [
   "sysinfo",
 ] }
 rattler_git = { version = "0.1.3" }
 rattler_prefix_guard = { version = "0.1.1" }
 
-rattler_config = { version = "0.5.2" }
-rattler = { version = "0.46.0", default-features = false, features = [
+rattler_config = { version = "0.6" }
+rattler = { version = "0.47", default-features = false, features = [
   "cli-tools",
   "indicatif",
 ] }
-rattler_cache = { version = "0.10.0", default-features = false }
-rattler_index = { version = "0.30.5", default-features = false }
-rattler_upload = { version = "0.8.0", default-features = false }
+rattler_cache = { version = "0.10.2", default-features = false }
+rattler_index = { version = "0.30.8", default-features = false }
+rattler_upload = { version = "0.8.3", default-features = false }
 rattler_s3 = { version = "0.2.5" }
 rattler_redaction = { version = "0.2.1", default-features = false }
-rattler_repodata_gateway = { version = "0.29.6", default-features = false, features = [
+rattler_repodata_gateway = { version = "0.30", default-features = false, features = [
   "gateway",
 ] }
-rattler_solve = { version = "7.1.4", default-features = false, features = [
+rattler_solve = { version = "7.2", default-features = false, features = [
   "resolvo",
   "serde",
 ] }
-rattler_virtual_packages = { version = "3.0.2", default-features = false }
-rattler_menuinst = "0.2.67"
+rattler_virtual_packages = { version = "3.0.3", default-features = false }
+rattler_menuinst = "0.2.69"
 
 
 # Internal rattler-build crates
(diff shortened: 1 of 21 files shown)


crates/rattler_build_core/src/render/solver.rs
```diff
@@ -208,7 +208,9 @@ pub async fn load_repodatas(
         .clear()
         .unwrap();
 
-    Ok(result)
+    // `query` now returns a `RepoDataQueryOutput` (records plus non-fatal
```
wolfv: Can we emit the warnings here? Or elsewhere? That would be a nice improvement.

docs/config.md
```diff
@@ -1,7 +1,33 @@
...
+
+1. The system-wide configuration of each tool: `/etc/pixi/config.toml` followed by `/etc/rattler-build/config.toml` (on Windows: `C:\ProgramData\<tool>\config.toml`)
+2. The per-user configuration of each tool: the platform config directory (`$XDG_CONFIG_HOME/<tool>/config.toml`, e.g. `~/.config/pixi/config.toml`) and the tool home (`$PIXI_HOME` / `$RATTLER_BUILD_HOME`, defaulting to `~/.pixi/config.toml` / `~/.rattler-build/config.toml`), for `pixi` first and then `rattler-build`
+
+In other words all system-wide files are read before any per-user file, and within each group pixi's files are overridden by Rattler-Build's own files. This means that settings such as default channels, mirrors, or S3 options that you have configured for pixi are picked up by Rattler-Build automatically, and can be overridden in Rattler-Build's own configuration files.
+
+Alternatively, a single configuration file can be specified explicitly with `--config-file` (e.g. `--config-file ~/.pixi/config.toml`), which disables the automatic discovery and loads only that file.
+
+!!! note "Behavior change"
```
wolfv, line 14: Should we add a mode in which rattler-build doesn't load any configuration so users can preserve the old behavior?

wolfv (commented):
Looks good, some comments

crates/rattler_build_core/src/render/solver.rs
baszalmstra: Yeah it would be good to add them for sure! Now they are silently dropped.

Cargo.toml
```diff
@@ -133,39 +134,38 @@ tracing-subscriber = { version = "0.3", features = [
...
 ] }
-rattler_solve = { version = "7.1.4", default-features = false, features = [
+rattler_solve = { version = "7.2", default-features = false, features = [
   "resolvo",
   "serde",
 ] }
-rattler_virtual_packages = { version = "3.0.2", default-features = false }
-rattler_menuinst = "0.2.67"
+rattler_virtual_packages = { version = "3.0.3", default-features = false }
+rattler_menuinst = "0.2.69"
```
baszalmstra, line 168: Would be better to only bump minor patches and leave patch bumps in the lockfile only.

PR #2641, opened 2026-07-06
fix: inherit all top-level build settings for cache outputs

### LINKED ISSUE #2497: Top level build number gets lost in outputs when inheriting from a staging build
Similar to https://github.com/prefix-dev/rattler-build/issues/1918

If an output has a `build` section and inherits from a staging build, the build number gets lost.
I attempted a workaround in gdal-feedstock, but it gets overriden by the bot. (https://github.com/conda-forge/gdal-feedstock/pull/1206, I'll open a bot issue too)

Minimal reproducer

```yaml
recipe:
  name: libgdal-split
  version: 3.21.3

build:
  number: 1

outputs:
  - staging:
      name: core-build
    requirements:
      build:
    build:
      script: build

  - package:
      name: libgdal-core
    # delete this line for `libgdal-core` to get build number 1 as intended.
    inherit: core-build
    requirements:
      build:
    build:
      script: build
```

Hofer-Julian:
## Summary

Fixes prefix-dev/rattler-build#2497.

A package output that inherits from a staging cache and has its own `build` section only inherited a **hand-maintained subset** of top-level build fields. That subset was a partial copy of `merge_stage1_build`, and whenever a new build field was added but not mirrored into it, the field was silently dropped for cache-inheriting outputs.

This is what caused the top-level `build.number` to get lost for such outputs (the symptom reported in the issue). [#2460](<https://github.com/prefix-dev/rattler-build/issues/2460>) patched `number` and `string` reactively, but the underlying duplication remained — for example `build.files` and the v3 package `flags` were still dropped.

## Change

Reuse `merge_stage1_build` as the single source of truth for which build fields are inheritable, then restore the output's own script. Cache inheritance intentionally does **not** inherit the top-level script (the cache has its own script; a cache-inheriting output packages the restored files), so that one field is overridden back to the output's value after the merge.

This also incidentally fixes a latent bug in the old branch where `merge_build_and_host_envs` could be reset from `true` back to `false`.

## Reproducer (from the issue)

```yaml
recipe:
  name: libgdal-split
  version: 3.21.3
build:
  number: 1
  files: ["*.txt"]
outputs:
  - staging:
      name: core-build
    build:
      script: build
  - package:
      name: libgdal-core
    inherit: core-build
    build:
      script: build
```

Before: `libgdal-core` renders with build number `0` and no `files`.<br>After: it renders with build number `1` and inherits `files: ["*.txt"]`.

## Tests

* Added `test_cache_inherited_output_inherits_top_level_build_settings` in `variant_render.rs`, covering the issue reproducer and asserting that both the build number **and** `build.files` are inherited.
* The existing `test_cache_inherited_output_inherits_top_level_build_number_before_build_string` continues to pass.
* Full `rattler_build_recipe` test suite is green.

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)

[https://claude.ai/code/session_01128Z1F2koWdxC5VKYfW7wr](<https://claude.ai/code/session_01128Z1F2koWdxC5VKYfW7wr>)

---

*Generated by [Claude Code](<https://claude.ai/code/session_01128Z1F2koWdxC5VKYfW7wr>)*

### DIFF
--- a/crates/rattler_build_recipe/src/stage0/evaluate.rs
+++ b/crates/rattler_build_recipe/src/stage0/evaluate.rs
@@ -3118,48 +3118,22 @@ fn evaluate_package_output_to_recipe(
         let output_build = output.build.evaluate(context)?;
         merge_stage1_build(toplevel_build, output_build)
     } else {
-        // Cache inheritance: only merge non-script build settings
+        // Cache inheritance: inherit all top-level build settings that the
+        // output does not set itself, EXCEPT the script (the cache has its own
+        // script; a cache-inheriting output packages the restored files and
+        // does not need to re-run the top-level script).
+        //
+        // We reuse `merge_stage1_build` as the single source of truth for which
+        // build fields are inheritable, then restore the output's own script.
+        // Maintaining a separate hand-written field list here caused top-level
+        // settings (build number, `files`, ...) to be silently dropped whenever
+        // a new field was added but not mirrored into this branch (see #2497).
         let toplevel_build = recipe.build.evaluate(context)?;
-        let mut output_build = output.build.evaluate(context)?;
-
-        // Only inherit specific build settings from top-level, not the script
-        if output_build.dynamic_linking.is_default() {
-            output_build.dynamic_linking = toplevel_build.dynamic_linking;
-        }
-        if output_build.prefix_detection.is_default() {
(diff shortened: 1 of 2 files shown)


isuruf (approved):
Thanks for fixing this

PR #2646, opened 2026-07-10
feat: add experimental build steps

baszalmstra:
## What

Adds experimental `build.steps` as an alternative to `build.script` for package outputs.

Each step runs as an isolated section of the generated build wrapper and can define its own `if` condition, `interpreter`, `cwd`, and `env`. Environment and shell state from one step do not leak into later steps, and a failing step stops the build.

```yaml
build:
  steps:
    - if: unix
      run: make
    - run: make install
      env:
        DESTDIR: $PREFIX
```

`script` and `steps` are mutually exclusive. Existing recipes using `build.script` or the default `build.sh` / `build.bat` keep their current behavior.

The new syntax currently only supports `run` steps and requires `--experimental`. The stage 0 and stage 1 Rust and Python recipe APIs expose the build plan as either a script or steps.

## Testing

- `pixi run build-release`
- `pixi run pytest test/end-to-end/test_build_steps.py -v`

### DIFF
--- a/crates/rattler_build_core/src/script.rs
+++ b/crates/rattler_build_core/src/script.rs
@@ -9,14 +9,16 @@ use rattler_build_jinja::Jinja;
 
 // Re-export from rattler_build_script
 pub use rattler_build_script::{
-    ExecutionArgs, ExecutionContext, InterpreterError, ResolvedScriptContents, RuntimeEnv,
-    SandboxArguments, SandboxConfiguration, Script, ScriptContent, platform_script_extensions,
+    BuildScriptSection, ExecutionArgs, ExecutionContext, InterpreterError, ResolvedScriptContents,
+    RuntimeEnv, SandboxArguments, SandboxConfiguration, Script, ScriptContent,
+    platform_script_extensions,
 };
 
 use crate::{
     env_vars::{self},
     metadata::Output,
 };
+use rattler_build_recipe::stage1::build::BuildPlan;
 
 impl Output {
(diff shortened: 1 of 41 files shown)


wolfv:
Should we gate this behind --experimental?

baszalmstra:
It is!

crates/rattler_build_core/src/script.rs
```diff
@@ -26,11 +27,23 @@ impl Output {
...
+    ///
+    /// The build script is always expressed as an ordered list of sections: a
+    /// `build.script` is a single section, and `build.steps` are one section per
+    /// step. Both go through the same execution path.
     async fn prepare_build_script(&self) -> Result<ExecutionArgs, std::io::Error> {
         let host_prefix = self.build_configuration.directories.host_prefix.clone();
         let target_platform = self.build_configuration.target_platform;
         let env_isolation = self.build_configuration.env_isolation;
+        let build = self.recipe.build();
+        if matches!(&build.plan, BuildPlan::Steps(_)) && !self.build_configuration.experimental {
```
wolfv, line 42: We could also already error when parsing?

baszalmstra: Yes. This is already rejected during stage 0 evaluation, with coverage in `test_build_steps_requires_experimental`. I kept the execution-time check as a defensive guard for callers that construct a stage 1 `BuildPlan::Steps` programmatically and therefore bypass parsing/rendering.

PR #2655, opened 2026-07-15
fix: set `LIBRARY_PREFIX` and friends for noarch builds on Windows

### LINKED ISSUE #2475: Cannot compile on windows for noarch:generic
Since target_platform is set to noarch for `noarch: generic` packages, env variables like `ProgramFiles(x86)` are not passed through, resulting in build failures.

baszalmstra:
## What

For `noarch` outputs there is no concrete target platform, so `os_vars` gated the Windows-specific target variables on `target_platform.is_windows()`, which is `false` for `noarch`. As a result `LIBRARY_PREFIX`, `LIBRARY_BIN`, `SCRIPTS`, `LIB`, `INCLUDE`, ... were never set, and `noarch: generic` recipes that still compile native artifacts during the build failed on Windows.

`os_vars` now also receives the host platform and, when the target is `noarch`, uses it to decide which OS-specific variables to emit (path variable casing, `SHLIB_EXT`, and the target OS default env vars). Non-`noarch` builds and cross-compilation are unchanged.

Fixes #2475

### DIFF
--- a/py-rattler-build/rust/src/render.rs
+++ b/py-rattler-build/rust/src/render.rs
@@ -78,6 +78,7 @@ impl PyRenderConfig {
         let os_env_var_keys = rattler_build::env_vars::os_vars(
             &std::path::PathBuf::new(),
             &host_platform,
+            &host_platform,
             EnvironmentIsolation::default(),
             &std::path::PathBuf::new(),
         )
--- a/py-rattler-build/rust/src/render.rs
+++ b/py-rattler-build/rust/src/render.rs
@@ -77,7 +77,7 @@ impl PyRenderConfig {
         // We use an empty prefix path since we just need the keys, not the values
         let os_env_var_keys = rattler_build::env_vars::os_vars(
             &std::path::PathBuf::new(),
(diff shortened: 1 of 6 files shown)


py-rattler-build/rust/src/render.rs
```diff
@@ -78,6 +78,7 @@ impl PyRenderConfig {
         let os_env_var_keys = rattler_build::env_vars::os_vars(
             &std::path::PathBuf::new(),
             &host_platform,
+            &host_platform,
```
Hofer-Julian, line 81: Shouldn't one of them be target_platform?

src/lib.rs
```diff
@@ -393,6 +393,7 @@ pub async fn get_build_output(
     let os_env_var_keys = env_vars::os_vars(
         &std::path::PathBuf::new(),
         &build_data.host_platform,
+        &build_data.host_platform,
```
Hofer-Julian, line 396: Same here?

Hofer-Julian (approved):

PR #2663, opened 2026-07-16
feat: keep `target_platform` concrete for noarch packages and add `build.subdir`

baszalmstra:
For noarch outputs, `target_platform` used to be replaced with `noarch` everywhere: in the variant, the jinja context, and the `target_platform` environment variable in the build script. This does not match conda-build, where `target_platform` keeps the requested platform and only the subdir of the resulting package becomes `noarch` (see the Discord discussion with @isuruf).

With this PR:

```yaml
build:
  noarch: generic
  script:
    - echo $target_platform  # linux-64 (was: noarch)
    - echo $SUBDIR           # noarch (unchanged)
```

The package is still stamped as `noarch` and placed in the `noarch` folder.

Build strings do not change. The hash input still replaces `target_platform` with `noarch` for noarch packages (now inside `HashInput::from_variant`), so hashes remain platform independent and byte-identical to what rattler-build produces today. `test_build_string_py_noarch` and `test_noarch_variants` assert the exact same build strings as before, and a new test asserts that noarch hashes are identical across target platforms.

## Implementation

- `BuildConfiguration::target_platform` is now always a concrete platform. The subdir of a package is derived instead: `Output::subdir()` returns `build.subdir` if set, `noarch` for noarch packages, and the target platform otherwise.
- Everything that used `target_platform == NoArch` as an "is this noarch?" check (relinking, linking checks, the Windows symlink check, run-export selection, `--noarch-build-platform`, native test skipping) now uses `Output::subdir()`, so behavior there is unchanged.
- index.json `subdir`/`arch`/`platform`, the output folder, prefix placeholder detection, file mapping, and package content tests also go through `Output::subdir()` and behave as before.

Visible behavior changes for noarch builds besides `target_platform` itself:

- `CONDA_BUILD_CROSS_COMPILATION` is now `0` when host and build platform match. It was always `1` for noarch because `noarch != build_platform`.
- The platform default environment variables (e.g. `MACOSX_DEPLOYMENT_TARGET`) are now set in the build script because `os_vars` receives a concrete platform. This is what made the compiler activation workaround in `tzdata-feedstock` necessary before.
- `ARCH` is unchanged (it was already derived from the host platform).

## `build.subdir`

Conda-build has an output-level `target` key that only controls the subdir the package is stamped with. This PR ports it as `build.subdir` (renamed per @isuruf's suggestion):

```yaml
build:
  subdir: linux-aarch64
```

It only affects index.json and the output folder. It does not change `target_platform` or dependency resolution, and it cannot be combined with `noarch`. The key supports jinja templates and inherits from the top-level build section in multi-output recipes, like `noarch`.

## Open questions

- `SUBDIR` in the build script still refers to the package subdir (`noarch`), while conda-build populates it from `host_subdir` (a concrete platform). Should we follow conda-build here as well?
- Test scripts still see `target_platform` as the package subdir, because at test time the value is derived from the package (index.json / `hash_input.json`) rather than from the build.

## Compatibility with existing conda-forge recipes

<details><summary>Assessment of the impact on conda-forge recipes</summary>

### Summary

Changing rattler-build so that `target_platform` remains the requested platform for noarch outputs appears to have a small compatibility risk for conda-forge recipes.

The search did not find a current `recipe.yaml` that clearly and intentionally relies on `target_platform=noarch`. The strongest ecosystem evidence points in the opposite direction: `tzdata-feedstock` explicitly works around rattler-build setting `target_platform` to `noarch`.

One recipe, `m2w64-sysroot-feedstock`, has a build script whose behavior can change depending on the value. It should be included in regression testing, although its normal conda-forge noarch build is unlikely to change.

### Behavior in conda-build

Given a noarch recipe built with `target_platform=win-64`, conda-build keeps the requested target platform visible to the build script:

```yaml
package:
  name: foo
  version: "1.0.0"

build:
  number: 0
  noarch: generic
  script:
    - echo %target_platform%
```

Running:

```console
conda-build recipe --variants "{target_platform: win-64}"
```

prints:

```text
win-64
```

The resulting package is nevertheless placed in the `noarch` subdirectory.

Conda-build therefore distinguishes between:

- the requested target or host platform, exposed as `target_platform`;
- whether the package is noarch, stored as recipe metadata;
- the final package subdirectory.

`SUBDIR` is populated from `config.host_subdir`, not from the final package subdirectory ([source](https://github.com/conda/conda-build/blob/88e566afe3b903ec5ad9ef21cc7a2b383caf3911/conda_build/environ.py#L390-L415)). When package metadata is created, conda-build explicitly sets the package subdirectory to `noarch` when `metadata.noarch` is set ([source](https://github.com/conda/conda-build/blob/88e566afe3b903ec5ad9ef21cc7a2b383caf3911/conda_build/metadata.py#L1899-L1907)).

Conda-build also supports an output-level `target` key:

```yaml
outputs:
  - name: foo
    target: noarch
```

This is assigned to `output_metadata.config.target_subdir` ([source](https://github.com/conda/conda-build/blob/88e566afe3b903ec5ad9ef21cc7a2b383caf3911/conda_build/metadata.py#L2513-L2514)) and controls the final package subdirectory ([source](https://github.com/conda/conda-build/blob/88e566afe3b903ec5ad9ef21cc7a2b383caf3911/conda_build/build.py#L1981-L1993)). It remains separate from `target_platform`.

### Conda-forge search

The search focused on current `recipe.yaml` files because those are the recipes using the v1 recipe format and rattler-build. It covered:

- literal comparisons between `target_platform` and `noarch`;
- `${{ target_platform }}` interpolation in noarch outputs;
- `build_platform != target_platform` selectors in noarch outputs;
- inline build scripts referencing `target_platform`;
- external shell and batch scripts referencing the runtime `target_platform` environment variable;
- mixed multi-output recipes where only some outputs are noarch.

No current `recipe.yaml` was found containing a literal comparison such as:

```yaml
target_platform == "noarch"
target_platform != "noarch"
```

GitHub code search indexes default branches and can lag behind recent changes, so this is not a guarantee that no such recipe exists. It does, however, cover the most direct ways a recipe could intentionally depend on the behavior.

### Confirmed workaround: `tzdata-feedstock`

[`tzdata-feedstock`](https://github.com/conda-forge/tzdata-feedstock/blob/7861f3afa58b5b0bb7e7fc171358ca3a725ee570/recipe/recipe.yaml#L16-L23) explicitly overrides `target_platform`:

```yaml
build:
  number: 1
  noarch: generic
  script:
    file: build.sh
    # Workaround as rattler-build sets target_platform to noarch, such that the compiler-activation script would not be run
    # https://github.com/conda-forge/tzdata-feedstock/pull/34#issuecomment-3593550598
    env: { target_platform: "${{ host_platform }}" }
```

This is not a dependency on the current behavior. It documents a concrete problem caused by it: compiler activation does not run when `target_platform=noarch`.

Keeping `target_platform` concrete would make this workaround unnecessary and align rattler-build with conda-build.

### Potentially behavior-sensitive recipe: `m2w64-sysroot-feedstock`

[`m2w64-sysroot-feedstock`](https://github.com/conda-forge/m2w64-sysroot-feedstock/blob/c5ae60d3f8d6efa47a86be4db1aa5493ad2d8918/recipe/recipe.yaml#L192-L198) defines a noarch output using `copy.sh`:

```yaml
- package:
    name: m2w64-sysroot_win-64
  build:
    noarch: generic
    merge_build_and_host_envs: false
    script: copy.sh
```

The corresponding [`copy.sh`](https://github.com/conda-forge/m2w64-sysroot-feedstock/blob/c5ae60d3f8d6efa47a86be4db1aa5493ad2d8918/recipe/copy.sh#L10-L13) contains:

```bash
if [[ "$target_platform" != "win-64" ]]; then
  mkdir -p ${PREFIX}/${HOST}/sysroot
  ln -sf ${PREFIX}/${HOST}/sysroot/usr ${PREFIX}/${HOST}/sysroot/ucrt64
fi
```

With the current rattler-build behavior, the condition is true because `target_platform=noarch`. With a concrete `target_platform`, it would be false when the requested platform is `win-64`.

This deserves regression coverage. However, the noarch package is normally produced in a Linux conda-forge job. In that job the concrete target is `linux-64`, so the condition remains true and the resulting artifact should be unchanged.

### References that do not rely on `target_platform=noarch`

Several noarch outputs use expressions such as:

```yaml
- if: build_platform != target_platform
  then:
    - python
    - cross-python_${{ target_platform }}
```

Examples include:

- [`mnelab-feedstock`](https://github.com/conda-forge/mnelab-feedstock/blob/d5a8af9cb6f6d1f064f816e5b1a95aac543d8a7b/recipe/recipe.yaml#L16-L32)
- [`lightgbm-feedstock`](https://github.com/conda-forge/lightgbm-feedstock/blob/80d7982ebf3297c33671e172007016d46b18cd33/recipe/recipe.yaml#L87-L101)
- [`conda-global-feedstock`](https://github.com/conda-forge/conda-global-feedstock/blob/054a2af2c512c81e5cad238c66662c67e60456de/recipe/recipe.yaml#L61-L78)
- [`pybind11-feedstock`](https://github.com/conda-forge/pybind11-feedstock/blob/7d2ea6398560d09fad899862455721d27d89dc1f/recipe/recipe.yaml#L89-L105)
- [`pyb2d3-feedstock`](https://github.com/conda-forge/pyb2d3-feedstock/blob/6eae35ca06bf03d136d4adc74c26105acf3dca88/recipe/recipe.yaml#L63-L84)
- [`gz-msgs-feedstock`](https://github.com/conda-forge/gz-msgs-feedstock/blob/3181b50ec6970b009e92f705ed92fe50b1329bbf/recipe/recipe.yaml#L89-L115)

These are not evidence of reliance on `target_platform=noarch`. Recipe expressions are evaluated using the requested platform before rattler-build assigns the effective noarch output platform. If these expressions saw `noarch`, some would attempt to depend on a package such as `cross-python_noarch`, which is clearly not their intent.

Other recipes explicitly require selectors to see the requested platform:

- [`stim-split-feedstock`](https://github.com/conda-forge/stim-split-feedstock/blob/c6bb026b59abe85f06f3be6715f4cd1fcd6d04e9/recipe/recipe.yaml#L82-L87) skips its noarch output unless `target_platform` is `linux-64`.
- [`backports.zstd-feedstock`](https://github.com/conda-forge/backports.zstd-feedstock/blob/85e12f8c168b1c049db1996fc788ba638151b2bf/recipe/recipe.yaml#L55-L59) uses `target_platform == "linux-64"` when selecting its single noarch build.

These recipes support preserving the requested target platform rather than replacing it with `noarch`.

### Risk assessment

The direct recipe compatibility risk appears low:

- no recipe was found that intentionally checks for `target_platform=noarch`;
- one recipe explicitly works around the current behavior;
- one external script is potentially behavior-sensitive but should be unchanged in its normal noarch build job;
- many noarch recipes already expect recipe expressions to see the requested platform.

The less visible risk is indirect behavior in activation scripts supplied by build dependencies. Those scripts can inspect `target_platform` without the recipe mentioning it. The `tzdata` workaround demonstrates that this already causes incompatibilities when the value is `noarch`.

The intended invariant:

> `target_platform` describes the platform for which the build was requested. Whether the resulting package is noarch is separate output metadata and must not replace the requested target platform.

</details>

### DIFF
--- a/crates/rattler_build_core/src/package_test/content_test.rs
+++ b/crates/rattler_build_core/src/package_test/content_test.rs
@@ -395,7 +395,7 @@ impl PackageContentsTestExt for PackageContentsTest {
     fn run_test(&self, paths: &PathsJson, output: &Output) -> Result<(), TestError> {
         let span = tracing::info_span!("Package content test");
         let _enter = span.enter();
-        let target_platform = output.target_platform();
+        let target_platform = &output.subdir();
         let paths: Vec<&PathBuf> = paths.paths.iter().map(|p| &p.relative_path).collect();
 
         let mut collected_issues = Vec::new();
--- a/crates/rattler_build_core/src/package_test/content_test.rs
+++ b/crates/rattler_build_core/src/package_test/content_test.rs
@@ -395,7 +395,7 @@ impl PackageContentsTestExt for PackageContentsTest {
     fn run_test(&self, paths: &PathsJson, output: &Output) -> Result<(), TestError> {
         let span = tracing::info_span!("Package content test");
         let _enter = span.enter();
-        let target_platform = &output.subdir();
+        let subdir = &output.subdir();
         let paths: Vec<&PathBuf> = paths.paths.iter().map(|p| &p.relative_path).collect();
 
         let mut collected_issues = Vec::new();
@@ -416,7 +416,7 @@ impl PackageContentsTestExt for PackageContentsTest {
             let globs = self.get_globs_for_section(
                 section.clone(),
                 true,
-                target_platform,
+                subdir,
                 version_independent_override,
             )?;
             Self::check_globs(
@@ -432,7 +432,7 @@ impl PackageContentsTestExt for PackageContentsTest {
             let globs = self.get_globs_for_section(
                 section,
                 false,
-                target_platform,
+                subdir,
                 version_independent_override,
             )?;
             Self::check_globs(
--- a/crates/rattler_build_core/src/packaging/file_mapper.rs
+++ b/crates/rattler_build_core/src/packaging/file_mapper.rs
@@ -98,7 +98,7 @@ impl Output {
         prefix: &Path,
         dest_folder: &Path,
     ) -> Result<Option<PathBuf>, PackagingError> {
-        let target_platform = &self.build_configuration.target_platform;
+        let target_platform = &self.subdir();
         let entry_points = &self.recipe.build().python.entry_points;
 
         let path_rel = path.strip_prefix(prefix)?;
--- a/crates/rattler_build_core/src/packaging/file_mapper.rs
+++ b/crates/rattler_build_core/src/packaging/file_mapper.rs
@@ -98,7 +98,7 @@ impl Output {
         prefix: &Path,
         dest_folder: &Path,
     ) -> Result<Option<PathBuf>, PackagingError> {
-        let target_platform = &self.subdir();
+        let subdir = &self.subdir();
         let entry_points = &self.recipe.build().python.entry_points;
 
         let path_rel = path.strip_prefix(prefix)?;
@@ -187,7 +187,7 @@ impl Output {
                 // on Windows, if the file ends with -script.py, remove the -script.py suffix
                 if let Some(Component::Normal(name)) = new_parts.last_mut()
                     && let Some(name_str) = name.to_str()
-                    && target_platform.is_windows()
+                    && subdir.is_windows()
                     && let Some(stripped_suffix) = name_str.strip_suffix("-script.py")
                 {
                     *name = stripped_suffix.as_ref();
@@ -217,7 +217,7 @@ impl Output {
 
         // Handle symlinks: make absolute symlinks relative and copy the link
         if metadata.file_type().is_symlink() {
-            if target_platform.is_windows() {
+            if subdir.is_windows() {
                 tracing::warn!("Symlink creation on Windows requires administrator privileges");
             }
             // Read the link target
--- a/crates/rattler_build_core/src/build.rs
+++ b/crates/rattler_build_core/src/build.rs
@@ -234,9 +234,8 @@ pub async fn run_build(
     // Check for symlinks on Windows if not allowed
     // Skip the check for noarch packages that have __unix in run dependencies,
     // since they will never be installed on Windows.
-    if (output.build_configuration.target_platform.is_windows()
-        || (output.build_configuration.target_platform == Platform::NoArch
-            && !has_unix_virtual_package(&output)))
+    if (output.subdir().is_windows()
+        || (output.subdir() == Platform::NoArch && !has_unix_virtual_package(&output)))
         && !tool_configuration.allow_symlinks_on_windows
     {
         tracing::info!("Checking for symlinks ...");
--- a/crates/rattler_build_core/src/env_vars.rs
+++ b/crates/rattler_build_core/src/env_vars.rs
@@ -394,7 +394,7 @@ pub fn vars(output: &Output, build_state: &str) -> HashMap<String, Option<String
     } else {
         insert!(vars, "CONDA_BUILD_CROSS_COMPILATION", "0");
     }
-    insert!(vars, "SUBDIR", output.target_platform().to_string());
+    insert!(vars, "SUBDIR", output.subdir().to_string());
     insert!(
         vars,
         "build_platform",
--- a/crates/rattler_build_core/src/packaging.rs
+++ b/crates/rattler_build_core/src/packaging.rs
@@ -919,13 +919,12 @@ pub fn package_conda(
 
     print_enhanced_file_listing(&files, &tmp, output)?;
 
-    let output_folder =
-        local_channel_dir.join(output.build_configuration.target_platform.to_string());
+    let output_folder = local_channel_dir.join(output.subdir().to_string());
     tracing::info!("Creating target folder '{}'", output_folder.display());
 
     fs::create_dir_all(&output_folder)?;
 
-    if let Platform::NoArch = output.build_configuration.target_platform {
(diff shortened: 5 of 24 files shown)


crates/rattler_build_core/src/packaging/file_mapper.rs
```diff
@@ -98,7 +98,7 @@ impl Output {
         prefix: &Path,
         dest_folder: &Path,
     ) -> Result<Option<PathBuf>, PackagingError> {
-        let target_platform = &self.build_configuration.target_platform;
+        let target_platform = &self.subdir();
```
baszalmstra: Lets rename this variable.

crates/rattler_build_core/src/package_test/content_test.rs
```diff
@@ -395,7 +395,7 @@ impl PackageContentsTestExt for PackageContentsTest {
     fn run_test(&self, paths: &PathsJson, output: &Output) -> Result<(), TestError> {
         let span = tracing::info_span!("Package content test");
         let _enter = span.enter();
-        let target_platform = output.target_platform();
+        let target_platform = &output.subdir();
```
baszalmstra: Lets rename this variable to subdir.

crates/rattler_build_core/src/packaging/file_mapper.rs
baszalmstra: Renamed in 6ae582f6, also applied the same rename in relink.rs and checks.rs which had the identical pattern.

crates/rattler_build_core/src/package_test/content_test.rs
baszalmstra: Renamed in 6ae582f6, also applied the same rename in relink.rs and checks.rs which had the identical pattern.

PR #2664, opened 2026-07-17
feat: make CMD subprocesses match `build_platform` architecture when running emulated

baszalmstra:
Build scripts and package tests now run under an x64 or ARM64 `cmd.exe` that matches the prefix they execute from.

`rattler-build` can run under a different Windows architecture than the configured build or host platform. Script execution now carries explicit build and host prefixes with their platforms through activation and interpreter lookup, instead of passing untyped paths. The Windows runner switches `cmd.exe` with `start /machine` when needed and forwards the child exit code.

The Windows build environment now derives its architecture metadata from the configured build platform. Shared package-test prefixes run with the package host architecture.

This changes the script API to require an `ExecutionContext`, making the build and host roles explicit.

This is needed to support https://github.com/conda-forge/conda-smithy/pull/2622

## Tests

- `cargo fmt --all -- --check`
- `cargo test -p rattler_build_script --lib`
- `cargo test -p rattler_build_core --lib`
- `cargo check --workspace --all-targets`
- `cargo check --manifest-path py-rattler-build/rust/Cargo.toml`
- [One-time Windows ARM runtime validation](https://github.com/prefix-dev/rattler-build/actions/runs/29581413313): an x64 `rattler-build` binary successfully ran `win-arm64` and `win-32` build scripts on Windows ARM.

## Note

Cross-architecture execution requires Windows 11 `start /machine` support.


crates/rattler_build_script/src/native_runner/cmd_exe.rs
```diff
@@ -33,8 +34,43 @@ IF "%CONDA_BUILD%" == "" (
...
+                script_name,
+            );
+            CommandSpec::new(
+                "cmd.exe",
+                [
+                    "/d".to_string(),
+                    "/v:on".to_string(),
+                    "/c".to_string(),
+                    command,
+                ],
```
baszalmstra, line 84: This requires comments.

crates/rattler_build_script/src/native_runner/cmd_exe.rs
```diff
@@ -33,8 +34,43 @@ IF "%CONDA_BUILD%" == "" (
...
+        if let Some(machine) = windows_machine_transition(
+            context.runtime().process_platform(),
+            context.build().platform(),
+        ) {
+            let script_name = build_script_path
+                .file_name()
+                .expect("generated build script has a filename")
+                .to_string_lossy();
+            let command = format!(
+                "start /b /wait /machine {} cmd.exe /d /c {} & exit /b !ERRORLEVEL!",
```
baszalmstra: What happens if scrip_name contains a space?

crates/rattler_build_script/src/execution.rs
```diff
@@ -136,6 +128,23 @@ impl ExecutionArgs {
 
         replacements
     }
+
+    /// Normalizes mutable Windows process architecture variables when the
+    /// configured build environment deliberately switches between x64 and ARM64.
+    pub fn apply_platform_environment(&mut self) {
```
baszalmstra: I dont like that this is a mutable function. Do we set the variables ourselves? Cant we have this behavior there?

crates/rattler_build_script/src/native_runner/cmd_exe.rs
baszalmstra: Added comments covering the `/machine` transition, `/wait`, delayed `ERRORLEVEL` expansion, and the explicit `exit /b` needed to propagate the child status.

crates/rattler_build_script/src/native_runner/cmd_exe.rs
baszalmstra: Quoted the script name with `quote_arg` and added coverage for a filename containing whitespace.

crates/rattler_build_script/src/execution.rs
baszalmstra: Moved this into activation and environment construction. `ExecutionContext` now exposes the architecture as a pure value, and `ExecutionArgs` is no longer mutated.

PR #2676, opened 2026-07-24
fix: Ignore case when querying variables from `RuntimeEnv` on Windows

### LINKED ISSUE #2674: [Windows] Interpreter system fallback mechanism doesn't work
Environment variable names are case insensitive on Windows, and the canonical variable name of `PATH` is `Path` on Windows.
So this part will always return empty string:
https://github.com/prefix-dev/rattler-build/blob/caafc953cff42f762963a43b031928cc184ff3a6/crates/rattler_build_script/src/runtime.rs#L51-L53

paperchalice:
The canonical name of `PATH` is `Path` on Windows, and `env: HashMap` is not case insensitive.
Fixes #2674.

### DIFF
--- a/crates/rattler_build_script/src/runtime.rs
+++ b/crates/rattler_build_script/src/runtime.rs
@@ -44,7 +44,14 @@ impl RuntimeEnv {
 
     /// Looks up an environment variable by name.
     pub fn var(&self, name: &str) -> Option<&str> {
-        self.env.get(name).map(String::as_str)
+        if self.platform.is_windows() {
+            self.env
+                .iter()
+                .find(|(key, _)| key.to_uppercase() == name.to_uppercase())
+                .map(|(_, val)| val.as_str())
+        } else {
+            self.env.get(name).map(String::as_str)
+        }
     }
 
     /// The value of `PATH`, or an empty string when it is unset.
--- a/crates/rattler_build_script/src/runtime.rs
+++ b/crates/rattler_build_script/src/runtime.rs
@@ -48,16 +48,6 @@ impl EnvironmentVariables {
     fn iter(&self) -> impl Iterator<Item = (&str, &str)> {
         self.values.iter().map(|(k, v)| (k.as_str(), v.as_str()))
     }
-
-    fn set_case_insensitive(&mut self, case_insensitive: bool) {
-        if self.case_insensitive_keys.is_some() != case_insensitive {
-            let values = std::mem::take(&mut self.values);
-            *self = Self::new(case_insensitive);
-            for (name, value) in values {
-                self.insert(name, value);
-            }
-        }
-    }
 }
 
 /// The environment rattler-build is running in: the process environment
@@ -80,10 +70,6 @@ impl RuntimeEnv {
             runtime_env.env.insert(name, value)
         }
         runtime_env
-    }
-
-    pub fn set_case_insensitive(&mut self, case_insensitive: bool) {
-        self.env.set_case_insensitive(case_insensitive);
     }
 
     /// Creates a runtime environment with an empty variable set and the given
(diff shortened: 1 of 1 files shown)


baszalmstra:
I wonder if we can make this more efficient by changing the type and allowing sensitive or case-insensitive get functions.

baszalmstra:
I dont think I was clear in my last comment. This only fixes `var()`; `vars()` and `with_var()` stay case-sensitive, so the
map can still hold both `Path` and `PATH`.

Could we put it in the type instead? A map keeping the original keys plus an
ASCII-lowercased index, with `RuntimeEnv` deciding once at construction whether
lookups fold case. `vars()` then still reports `Path` as `Path`, which we need
when handing the env to spawned processes.

This also needs a test or two!

paperchalice:
Should we use crates like unicase or case_insensitive_hashmap here?

wolfv:
Maybe something like this could work: 

```rust
#[derive(Debug, Clone)]
struct EnvironmentVariables {
    values: HashMap<String, String>,
    case_insensitive_keys: Option<HashMap<String, String>>,
}

impl EnvironmentVariables {
    fn new(case_insensitive: bool) -> Self {
        Self {
            values: HashMap::new(),
            case_insensitive_keys: case_insensitive.then(HashMap::new),
        }
    }

    fn get(&self, name: &str) -> Option<&str> {
        let name = self
            .case_insensitive_keys
            .as_ref()
            .and_then(|keys| keys.get(&name.to_ascii_lowercase()))
            .map_or(name, String::as_str);
        self.values.get(name).map(String::as_str)
    }

    fn insert(&mut self, name: String, value: String) {
        if let Some(keys) = &mut self.case_insensitive_keys {
            let normalized = name.to_ascii_lowercase();
            if let Some(original) = keys.get(&normalized) {
                self.values.insert(original.clone(), value);
                return;
            }
            keys.insert(normalized, name.clone());
        }
        self.values.insert(name, value);
    }

    fn iter(&self) -> impl Iterator<Item = (&str, &str)> {
        self.values.iter().map(|(k, v)| (k.as_str(), v.as_str()))
    }

    fn set_case_insensitive(&mut self, case_insensitive: bool) {
        if self.case_insensitive_keys.is_some() != case_insensitive {
            let values = std::mem::take(&mut self.values);
            *self = Self::new(case_insensitive);
            for (name, value) in values {
                self.insert(name, value);
            }
        }
    }
}
```

crates/rattler_build_script/src/runtime.rs
```diff
@@ -10,29 +10,88 @@ use std::collections::HashMap;
...
+        };
+        for (name, value) in std::env::vars() {
+            runtime_env.env.insert(name, value)
         }
+        runtime_env
+    }
+
+    pub fn set_case_insensitive(&mut self, case_insensitive: bool) {
+        self.env.set_case_insensitive(case_insensitive);
     }
```
baszalmstra: Do we really need this? Its unused isnt it? Same for the function in the EnvironmentVaraibles type.

crates/rattler_build_script/src/runtime.rs
```diff
@@ -10,29 +10,74 @@ use std::collections::HashMap;
...
 
 impl RuntimeEnv {
     /// Captures the real process environment variables and the current platform.
     pub fn current() -> Self {
-        Self {
-            env: std::env::vars().collect(),
-            process_platform: Platform::current(),
+        let current_platform = Platform::current();
+        let mut runtime_env = Self {
+            env: EnvironmentVariables::new(current_platform.is_windows()), // std::env::vars().collect(),
```
baszalmstra:
```suggestion
            env: EnvironmentVariables::new(current_platform.is_windows()),
```

baszalmstra:
@paperchalice Can you fix the CI issues?

wolfv:
Thank you!

PR #2702, opened 2026-07-30
feat: add runner to build configuration

baszalmstra:
The `Runner` API now exists in `rattler_build_script`, but the build configuration has no place to carry a runner. Library integrations that already pass `Configuration` through the build and test code would otherwise need to keep the execution backend as separate state.

This adds an `Arc<dyn Runner>` to `Configuration` and a `with_runner` builder method. The default stays `LocalRunner`, so existing callers keep the current behavior. CLI builds construct the local runner from the selected environment isolation mode, keeping those settings together. The runner types are also re-exported through the core script API.

This is configuration and API plumbing only. Existing script execution is unchanged.

## Testing

- `pixi run cargo-fmt-check`
- `pixi run cargo check --workspace --all-targets`
- `pixi run cargo clippy -p rattler_build_core --all-targets --all-features -- -D warnings -Dclippy::dbg_macro`
- `pixi run cargo check --manifest-path py-rattler-build/rust/Cargo.toml`
- `pixi run test`

### DIFF
--- a/crates/rattler_build_core/src/tool_configuration.rs
+++ b/crates/rattler_build_core/src/tool_configuration.rs
@@ -5,6 +5,7 @@ use std::{collections::HashMap, path::PathBuf, sync::Arc};
 
 use clap::ValueEnum;
 use rattler::package_cache::PackageCache;
+use rattler_build_script::runner::{LocalRunner, Runner};
 use rattler_conda_types::{ChannelConfig, Platform};
 #[cfg(feature = "s3")]
 use rattler_networking::s3_middleware;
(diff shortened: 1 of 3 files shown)


crates/rattler_build_core/src/tool_configuration.rs
```diff
@@ -566,3 +583,29 @@ pub async fn resolve_s3_credentials(
     let resolved = rattler_s3::ResolvedS3Credentials::from_sdk().await?;
     Ok(resolved)
 }
+
+#[cfg(test)]
+mod tests {
```
wolfv: Not sure if these tests add a lot of value lol

wolfv (changes requested):

wolfv (approved):
Rebased onto current main and removed the low-value runner plumbing tests.

PR #2703, opened 2026-07-30
fix: Handle copy_metadata errors for reflinked files

jmakovicka:
With rattler-build 0.68 and later, using git sources, copying of git objects from cache to the working directory fails with

FileSystem error: 'Permission denied (os error 13)'


dhirschfeld:
Just ran into this one - thanks for the fix!

image

PR #2710, opened 2026-07-31
fix: pixi config loading - warn less on unknown keys

wolfv:
This reduces the number of warnings on unknown keys (e.g. coming from the default `pixi.toml` that contains a key like `detached-environemnts = "/foo/bar"`.


Hofer-Julian (changes requested):
This feels like a workaround to me tbh.
I think we should just provide API in `rattler_config` for this

Issue #2657, opened 2026-07-15
Windows: test cleanup fails with `Access is denied (os error 5)` when a test spawns a lingering child process (MSVC mspdbsrv.exe)


danx12:
## Problem description

On Windows, a package **test that passes** still fails the whole run because rattler-build cannot remove its temporary test directory afterwards:

```
│ Testing commands:
│ ... cl /nologo /EHsc hello.c
│ hello.c
│ ... hello.exe
│ hello from cl.exe
╰─────────────────── (took 2 minutes)
Error:   × Test failed: failed to remove directory `...\test\test_<pkg>`: Access is denied. (os error 5)
```

The compiler command (`cl.exe`) succeeds and the program runs correctly. The error happens purely in the post-test cleanup (`remove_dir_all` of the test prefix).

Root cause: MSVC's `cl.exe` spawns a helper, `mspdbsrv.exe` (the PDB symbol server), which keeps running for a short while after `cl.exe` exits and holds an open handle inside the test directory. When rattler-build immediately tries to delete the directory, Windows refuses with `os error 5` (Access is denied). This is a lingering-child-process file-lock, not a real test failure.

This looks like the same class of problem previously reported in #1431 (closed/completed), which also shows `Access is denied. (os error 5)` / `os error 32` during build/test cleanup on Windows for Go- and JVM-based recipes.
## Request

Currently the only options are to disable the test entirely or add a recipe-side sleep hack, since there is no flag to skip only the cleanup step.

## Minimal reproducer

Any recipe whose test invokes the MSVC compiler reproduces it, because `cl.exe` always launches `mspdbsrv.exe`. Sketch:

```yaml
tests:
  - script:
      - if: win
        then:
          - cl /nologo /EHsc hello.c   # spawns mspdbsrv.exe
          - hello.exe
    files:
      recipe:
        - hello.c
```

`hello.c`:

```c
#include <stdio.h>
int main(void) { printf("hello from cl.exe\n"); return 0; }
```

The test passes, then rattler-build fails removing the test prefix with `Access is denied. (os error 5)`.

## Environment

- rattler-build: **0.69.0**
- OS: Windows Server container (running as an administrator account)

lucascolley:
It may be that I hit the same problem in https://github.com/conda-forge/circleci-feedstock/pull/16 and https://github.com/conda-forge/staged-recipes/pull/32124

eunos-1128:
[...]

PR #2729, opened 2026-08-11
fix: use `remove_dir_all_force` helper in `run_test.rs`

lucascolley:
I've been having some troubles recently which seem similar to https://github.com/prefix-dev/rattler-build/issues/2657. Claude Opus 5 reckoned this might be a fix.


Hofer-Julian (approved):
Let's give it a try

PR #2748, opened 2026-08-18
feat: auto-detect variant variables used in build scripts

wolfv:
## What

Recipes often consume variant variables directly from the environment of their build script (e.g. `${TARGET}` in the clang recipe's `build.sh`) without referencing them anywhere in the recipe, which previously required forwarding them manually via `build.variant.use_keys`.

This PR makes rattler-build scan build scripts for references to variant configuration keys. Matches count as used variables: they expand the variant matrix and participate in the build hash of the outputs whose scripts reference them — closing the gap with conda-build, whose script scanning is why recipes ported from conda-forge break today.

> Original request: *"Some users are complaining that we do not 'auto-extract' used variables from build.sh / build.bat files, or the script content. E.g. in the clang dev recipe they have to explicitly forward TARGET which is used as `${TARGET}` in the bash script. They would want us to scan those files and cross-reference as 'used variables' with the variant configuration."*

## How

Per review discussion the detection is a **literal key search with a sigil heuristic** — no per-interpreter syntax parsing, no regexes. A variant key counts as used when its normalized name occurs in the script text:

1. as a standalone word (not embedded in a longer identifier), case-sensitive, and
2. directly preceded by a *usage sigil*: one of `$` `{` `#` `!` `%` `:` `"` `'`, or an `env.` prefix.

That single rule covers the access spellings of every supported interpreter:

- matches: `$TARGET`, `${TARGET}`, `${TARGET:-x}`, `%TARGET%`, `!TARGET!`, `$env:TARGET`, `os.environ["TARGET"]`, `ENV['TARGET']`, `Sys.getenv("TARGET")`, `$ENV{TARGET}`, `$env.TARGET`, `process.env.TARGET`
- rejected: bare prose mentions (`# adjust TARGET here`), command names equal to a variant key (running `cmake` while the config pins `cmake`), assignments (`make TARGET=install`), longer identifiers (`$TARGET_ARCH`), different casing (`$target`)

The sigil set is the union across interpreters rather than per-dialect, because scripts routinely embed one interpreter in another (`bash` running `python -c "... os.environ['TARGET'] ..."`). The normalized key spelling is searched because that is exactly the name `env_vars_from_variant` exports to the script environment. Only names explicitly declared in the variant configuration can ever match; `build.variant.ignore_keys` opts out, `use_keys` covers undetectable usages (dynamically constructed names, `if defined TARGET`, `printenv TARGET`).

Structure:

- **`variable_scan` module in `rattler_build_script`**: the word-bounded, sigil-prefixed matcher (`find_referenced_variables`) plus `Script::detect_used_variables()`, which resolves script sources read-only — inline commands, explicit `script.file`, single-string command-or-path, and the default `build.sh`/`build.bat` discovery next to the recipe.
- **`script_variables` module in `rattler_build_recipe`**: walks all stage0 build plans (top-level, multi-output outputs, staging outputs, `build.steps`) and each evaluated stage1 output.
- **Two hooks in `variant_render`**: script-detected keys seed the combination matrix (`collect_used_variables`), and after evaluation each output's variant records the keys *its own* script references (analogous to `use_keys`) so build strings differ per value. Outputs that don't reference the key render identically and dedupe downstream.

## Performance

Negligible. Script execution is untouched; the scan only runs during rendering and is a plain substring search per candidate key over a few-KB script.

## Behavior change / discussion points

- **Hash churn**: existing recipes whose scripts reference a variant-config key with a sigil will grow additional variants / different hashes on rebuild.
- Possible follow-up: a log line when the scan adds keys that were not otherwise used ("detected `TARGET` in `build.sh`; add it to `build.variant.ignore_keys` to opt out").

## Test plan

- 10 unit tests in `variable_scan.rs`: access spellings across all interpreters, rejected bare mentions/assignments/command names, identifier boundaries and case sensitivity, empty candidates, file discovery and resolution.
- 7 integration tests in `variant_render.rs`: `build.sh` discovery expanding the matrix with distinct hashes, inline scripts, sigil-matching semantics, `build.bat` on a Windows build platform, a Python `script.file`, a multi-output recipe where only the output using `$TARGET` records it, and a negative test that names absent from the variant config are ignored.
- All 343 recipe-crate and 91 script-crate tests plus clippy pass. Docs section in `docs/variants.md` documents the rule with match/no-match examples.

https://claude.ai/code/session_016dFvfEwnyMU6HLqedzLaMX


Hofer-Julian:
I am also not a big fan of searching scripts for env vars, but after thinking a bit about it, I think we should go for it. The root problem IMO is that both `bash` and `cmd` silently accept references to environment variables that don't exist. That makes it very easy to miss that the environment variable isn't actually available.

I do have one request though: let's be less smart about this. How about we search for the literal `ENV_VAR` and completely ignore syntax like `${ENV_VAR}`. That is what tailwind does as well, and I don't see how that would be a big problem in practice.

wolfv:
We could try to do that but we would have to have a list of all available variant keys. It would also mean that we might take things mentioned in comments ... but that is actually also the case with the current implementation.

Hofer-Julian:
> have to have a list of all available variant keys

That's doable, right?

wolfv:
Implemented the "less smart" approach in 1f44c68 — and yes, the key list was already available at both call sites (the variant config keys when seeding the matrix, and the combination keys when recording each output's variant), so this was a straightforward inversion.

**New rule:** a variant key counts as used when its normalized name occurs literally in the script text as a standalone word — case-sensitive, bounded by non-identifier characters. So `${TARGET}`, `$TARGET`, `%TARGET%`, `os.environ["TARGET"]`, and a bare mention of `TARGET` all match, while `$TARGET_ARCH` and `$target` do not match the key `TARGET`. The normalized spelling is what `env_vars_from_variant` exports to the script environment, so the searched name is exactly the name a script would actually reference.

This deleted all nine dialect regexes (and the `regex` dependency from `rattler_build_script`) — every interpreter is now covered by the same rule, and the docs shrank from a spelling table to two sentences.

One trade-off worth having on the record: variant configs frequently contain package-name keys (`zlib`, `openssl`, `cmake`, ...), and those words appear in scripts more often than `$`-prefixed forms did — e.g. a `build.sh` that runs `cmake` while the variant config pins a `cmake` version would now record that pin in the hash. Word-boundary + case-sensitive matching keeps this contained, `ignore_keys` opts out per recipe, and arguably "the pinned tool version is in the hash" errs in the safe direction — but it is more matching than the previous syntax-aware scan did.

---
_Generated by [Claude Code](https://claude.ai/code)_

Hofer-Julian (approved):

wolfv:
Added the sigil heuristic in c33b02b, as a middle ground between the literal search and full syntax parsing.

**Rule:** a variant key counts as used when it occurs word-bounded, case-sensitive, and **directly preceded by a usage sigil** — one of `$` `{` `#` `!` `%` `:` `"` `'`, or an `env.` prefix. That single character set covers the access spelling of every supported interpreter:

| matches | rejected |
|---|---|
| `$TARGET`, `${TARGET}`, `${TARGET:-x}`, `${#TARGET}` | `# adjust TARGET here` (prose/comment) |
| `%TARGET%`, `%TARGET:.=_%`, `!TARGET!` | running `cmake` while the config pins `cmake` |
| `$env:TARGET`, `GetEnvironmentVariable("TARGET")` | `make TARGET=install` (assignment, not usage) |
| `os.environ["TARGET"]`, `ENV['TARGET']`, `Sys.getenv("TARGET")`, `$ENV{TARGET}` | `Config.cmake` (plain `.` is not a sigil — only `env.`) |
| `$env.TARGET`, `process.env.TARGET` | `$TARGET_ARCH`, `$target` |

One deliberate deviation from the per-dialect framing: the sigil set is the **union** across interpreters rather than chosen per script dialect, because scripts routinely embed one interpreter in another (`bash` running `python -c "... os.environ['TARGET'] ..."` is a common recipe pattern) — per-dialect sets would miss those, and the union barely widens any single dialect's set. Still no regexes: it's a substring search plus a one-character look-behind.

Known misses (real usages without a sigil): cmd's `if defined TARGET` and bash's `printenv TARGET` — `build.variant.use_keys` covers those, and the docs say so.

---
_Generated by [Claude Code](https://claude.ai/code)_

wolfv:
Posted this in a discussion in Zulip:

Hi all, we have a controversial rattler-build change to search for "variant keys" in the build scripts. It emulates something that conda-build did and that I always disliked: searching in the build script (bash, cmd.exe, Python, ...) for "used keys". For example, using `$TARGET` or `${TARGET}` in a bash script would add the `TARGET` variable as a "used key". 

Of course, parsing bash/cmd.exe etc. scripts is not so easy. The real way would be to use a proper parser that knows the context, what a comment is, what is executed or not, etc. - the things we _actually_ do in the recipes themselves (where we trace through the recipe to find the "used variables".

We don't have that luxury for bash or cmd.exe, so for example a $TARGET that is used in an `if` clause which never triggers on `osx-arm64` would still count towards the variant. Or if you use a comment `# ${TARGET}` it would _still_ count towards the variant, because properly parsing/tracing arbitrary bash / cmd.exe scripts is not feasible.

I wanted to gather your thoughts on whether we should add this to rattler-build or not ... 

The current way of doing this _properly_ is to forward the variant keys to the script using jinja (which will pick up the keys from the variant automatically):

```yaml
script:
  env:
     FOO: ${{ FOO }}
     TARGET: ${{ TARGET }}
  file: build.sh
```

## Why are used variables bad?

Used variables do two things:

- they change the hash
- they can create a "build matrix" (if the key in the variant config specifies more than one element, e.g. `python: ["3.12", "3.13", "3.14"]` -> leads to 3 x N builds

Why is changing the hash problematic? 

A changing hash leads to duplicate builds not being detected. For example, `rattler-build --skip-existing` relies on the hash being stable, in order to skip builds. A stable hash thus _minimizes_ the amount of packages being needlessly built.

wolfv:
For the variables forwarding thing in rattler build we could also propose: using real jinja in build scripts, or to have a initial comment line that we evaluate as yaml or so

PR #2751, opened 2026-08-18
fix: select platform-appropriate legacy test scripts

### LINKED ISSUE #2750: Legacy package tests run both Unix and Windows scripts
## Summary

When a legacy conda-build package contains both `info/test/run_test.sh` and `info/test/run_test.bat`, rattler-build executes both files on every platform. Conda-build intentionally includes both files in noarch packages, so valid noarch artifacts fail when tested with rattler-build.

## Reproduction

Use a noarch package containing both legacy scripts and run `rattler-build test` with rattler-build 0.74.0.

On Unix, rattler-build runs `run_test.sh` and then attempts `run_test.bat`, which fails because `cmd` is unavailable. On Windows, it runs `run_test.bat` and then also runs `run_test.sh`.

A downstream reproduction is conda-forge/conda-feedstock#312. Its Linux jobs fail with `interpreter 'cmd' was not found in the build environment`. Its Windows jobs complete the batch test, then execute the Unix test and fail inside it.

## Expected behavior

Run `run_test.sh` on Unix, run `run_test.bat` on Windows, and run `run_test.py` on every supported platform.

## Cause

`legacy_tests_from_folder()` collects both command scripts without consulting the concrete host platform, even though the documentation identifies them as Unix and Windows test files respectively.

The selector should use `host_platform`, since `target_platform` is `noarch` for these artifacts and `current_platform` can differ during cross or emulated builds.

## Affected versions

Reproduced with rattler-build 0.72.2 and 0.74.0. Current main at `3b0edc315efac949088f21c373c2182a7b66fb5e` has the same behavior.

## References

- https://github.com/conda-forge/conda-feedstock/pull/312
- https://github.com/conda-forge/conda-feedstock/actions/runs/32146419778/job/95741147551
- https://github.com/conda-forge/conda-feedstock/actions/runs/32146419778/job/95741147830
- https://github.com/conda/conda-build/blob/336e4af5d3a6c3dbb96090349660384d4eb492ca/conda_build/create_test.py#L63-L98
- https://github.com/conda/conda-build/blob/336e4af5d3a6c3dbb96090349660384d4eb492ca/conda_build/build.py#L3175-L3187

jezdez:
## Summary

Legacy noarch packages produced by conda-build can contain both Unix and Windows test scripts. Filter legacy command tests using the concrete host platform so `run_test.sh` runs only on Unix and `run_test.bat` runs only on Windows. Continue running `run_test.py` on every supported platform.

Fixes #2750

### DIFF
--- a/crates/rattler_build_core/src/package_test/run_test.rs
+++ b/crates/rattler_build_core/src/package_test/run_test.rs
@@ -303,7 +303,10 @@ impl Tests {
     }
 }
 
-async fn legacy_tests_from_folder(pkg: &Path) -> Result<(PathBuf, Vec<Tests>), std::io::Error> {
+async fn legacy_tests_from_folder(
+    pkg: &Path,
+    host_platform: Platform,
+) -> Result<(PathBuf, Vec<Tests>), std::io::Error> {
     let mut tests = Vec::new();
 
     let test_folder = pkg.join("info/test");
@@ -312,6 +315,11 @@ async fn legacy_tests_from_folder(pkg: &Path) -> Result<(PathBuf, Vec<Tests>), s
         return Ok((test_folder, tests));
     }
 
+    let command_test = if host_platform.is_windows() {
+        "run_test.bat"
+    } else {
+        "run_test.sh"
+    };
     let mut read_dir = tokio::fs::read_dir(&test_folder).await?;
 
     while let Some(entry) = read_dir.next_entry().await? {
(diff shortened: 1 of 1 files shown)


baszalmstra (approved):

isuruf (commented):
Does this select the specific file for running the tests or does this also only package the specific file? If the latter, we need to package both files, but use one only for running.

wolfv:
I believe legacy noarch packages already package both files. 

The only question I have is whether this should be `build_platform` or `host_platform` ... whcih can differ if running tests in a cross-compiled environment :) Which is also precisely what makes this difficult because if `build_platform == win64` and `host_platform == linux-64` then ... what to select?!

We could also add a condition to only do thids when `build_platform == host_platform`.

isuruf:
`build_platform` should be fine. When there's no emulation, we would have passed `--no-test` anyway

jezdez:
Thanks. Both files remain packaged. This only filters the legacy scripts after the package has been extracted for testing.

I switched the selector to `build_platform` in e2d16ef8 to match conda-build, which packages both variants for noarch packages and chooses `.bat` or `.sh` from the platform running the tests.

wolfv:
Thanks!

PR #2758, opened 2026-08-21
fix: emit canonical MatchSpecs for v3 packages

baszalmstra:
## Why

This closes a gap in our CEP 48 implementation. Packages marked with `repodata_revision: 3` were still writing positional legacy MatchSpecs to `info/index.json`, even though their v3 records require the canonical `name[...]` representation.

For revision 3 and newer, rattler-build now writes canonical MatchSpecs for `depends`, `constrains`, and `extra_depends`. Legacy packages keep their existing positional representation. The recipe parser uses the same revision threshold, so future revisions consistently support the v3 recipe fields too.

The index-shape tests also cover duplicate dependencies to make sure canonicalization does not change the existing deduplication behavior.

Related rattler indexer change: https://github.com/conda/rattler/pull/2718

## Testing

- `pixi run cargo test -p rattler_build_core test_canonical_matchspec_format_revision_threshold`
- `pixi run cargo test -p rattler_build_recipe test_render_future_repodata_revision_supports_v3_features`
- `pixi run cargo check -p rattler_build_core -p rattler_build_recipe`
- `RUSTC_WRAPPER= pixi run build-release`
- `pixi run pytest test/end-to-end/test_v3_repodata.py -v` (7 passed)
- `pixi run cargo fmt --check`

### DIFF
--- a/test/end-to-end/test_v3_repodata.py
+++ b/test/end-to-end/test_v3_repodata.py
@@ -146,9 +146,11 @@ def _read_map(data: bytes, offset: int, length: int) -> tuple[dict[Any, Any], in
 def assert_v3_index_json(index_json: dict[str, Any]) -> None:
     assert index_json["repodata_revision"] == 3
     assert index_json["flags"] == ["cuda", "blas:openblas"]
+    assert index_json["depends"].count('python[version=">=3.10"]') == 1
+    assert index_json["constrains"] == ['python[version="<4"]']
     assert index_json["extra_depends"] == {
-        "full": ["pandas >=2", "rich[extras=[jupyter]]"],
-        "plot": ["matplotlib >=3.8"],
+        "full": ['pandas[version=">=2"]', "rich[extras=[jupyter]]"],
+        "plot": ['matplotlib[version=">=3.8"]'],
     }
     assert 'scipy[when="python>=3.10"]' in index_json["depends"]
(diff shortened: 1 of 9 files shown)


test/end-to-end/test_v3_repodata.py
```diff
@@ -146,9 +146,11 @@ def _read_map(data: bytes, offset: int, length: int) -> tuple[dict[Any, Any], in
 def assert_v3_index_json(index_json: dict[str, Any]) -> None:
     assert index_json["repodata_revision"] == 3
     assert index_json["flags"] == ["cuda", "blas:openblas"]
+    assert index_json["depends"].count('python[version=">=3.10"]') == 1
+    assert index_json["constrains"] == ['python[version="<4"]']
```
Hofer-Julian, line 150: These pass without the canonicalization change too.

Legacy `Display` also normalizes, so `python >=3.10` and `python[version=">=3.10"]` already deduped to one entry before. I checked with the parent commit's binary.

Channel and URL specs would actually test this. `to_canonical_string()` strips credentials, so two deps that only differ by a `/t/<token>/` prefix become one under `--v3` but stay two under legacy.

Hofer-Julian (approved):
Comment scope fixed in the commit I pushed. The rest looks good to me.

PR #2761, opened 2026-08-22
feat: add nested output subpackages

wolfv:
## Summary

Add nested `subpackages` to a package output. A parent output builds once, child packages select files from that result, and selected files are removed from the parent.

```yaml
outputs:
  - package:
      name: mylib
    build:
      script: cmake --install build --prefix "$PREFIX"

    subpackages:
      - package:
          name: mylib-dev
        split:
          files:
            - include/**
            - lib/pkgconfig/**
        requirements:
          run:
            - ${{ pin_subpackage("mylib", exact=True) }}
        about:
          summary: Development files for mylib
```

This is syntactic and semantic support for the common “build once, split runtime/dev/docs packages” case. It avoids hand-authoring a staging output plus several sibling package outputs and, importantly, defines what happens to the parent files.

## Semantics

- The parent source preparation and build execute exactly once in an automatically generated private staging cache.
- Every child receives the union of its static `split.files` selection and its generated selection.
- The parent receives the complement of all child selections, followed by its own `build.files` filter.
- A file selected by two children is an error. Splits are partitions, not an order-dependent copy operation.
- Child version, build metadata, requirements, tests, and `about` can be overridden:
  - version defaults to the parent version;
  - omitted requirements inherit the parent package-time requirements, while an explicit section (including `{}`) replaces them;
  - omitted tests inherit; an explicit tests section replaces them;
  - `about` and non-executable `build` fields overlay the parent;
  - child `build.script`, `build.steps`, `build.files`, and `build.skip` are rejected with targeted guidance.
- Common build/host requirements and run exports flow through the private staging cache. Explicit child build/host requirements remain possible for child-specific packaging/post-processing needs.
- The common build always sees the finalized parent `PKG_NAME`, `PKG_VERSION`, `PKG_BUILDNUM`, `PKG_BUILD_STRING`, and `PKG_HASH`, including when only a child is requested with `--up-to`.

## Dynamic selection protocol

A child may add `split.script`. It runs once after the common build, in the same environment and prepared work tree:

```yaml
split:
  files:
    - include/**
  script: |
    find "$PREFIX/lib/cmake" -type f -printf 'lib/cmake/%P\n' \
      >> "$RATTLER_BUILD_SUBPACKAGE_FILES"
```

The script must create `RATTLER_BUILD_SUBPACKAGE_FILES` and write one prefix-relative path or glob per line. Blank lines and comments are ignored. Absolute paths, `..`, missing manifests, and non-regular manifests are rejected. Generated manifests are cached with the common staging result and included in the staging cache identity through the selector configuration.

The existing `RATTLER_BUILD_PACKAGE_FILES` override is retained for the parent build and applied before taking the parent complement.

## Design thoughts

I looked at Melange/Wolfi for comparison. Melange models subpackages next to one origin package and gives each a pipeline; reusable pipelines such as `split/dev`, `split/static`, and `split/manpages` physically move files from `targets.destdir` to `targets.subpkgdir`. That model has two especially good ideas:

1. subpackages are structurally owned by the build that produced their files;
2. common policies should eventually be reusable presets rather than repeated glob lists.

Rattler-build differs because one recipe already has multiple independently buildable package outputs and staging outputs. Nesting under a **package output** therefore seems more precise than a second top-level subpackage system: each existing output can own a split family.

I chose a declarative partition rather than exposing a mutable child destination directory. It makes overlap an explicit error, removes ordering effects between children, is portable across restored staging prefixes, and allows rebuild/render metadata to describe the split. `split.script` is an escape hatch for discovery, but it emits a manifest rather than moving files itself.

Possible follow-ups, deliberately left out of this PR:

- reusable selection policies inspired by Melange, e.g. `split: {uses: dev}` for headers, pkg-config, CMake, unversioned linker symlinks, and static archives;
- conditional/range-generated subpackages;
- allowing a split family to consume an explicitly named staging output instead of owning a private one;
- deciding whether tests should inherit by default or default to empty (this PR chooses inheritance plus explicit replacement);
- supporting one syntax on single-output recipes as sugar for converting them to a multi-output recipe;
- richer generated-manifest operations such as explicit exclusions (static `split.files` already supports include/exclude mappings).

Current scope is intentionally one nested level. A parent with subpackages cannot also specify `inherit`, because the parent build itself is the source of the private staging cache.

## Correctness and safety details

- All produced files are retained in the private cache even when the parent has `build.files`; the parent filter is applied only after splitting.
- Child selectors participate in used-variable and variant discovery.
- Parent identities are matched and propagated after build-string finalization for each variant.
- Selection overlap reporting is deterministic.
- Generated entries cannot escape the package prefix and cached selector outputs cannot be symlinks.
- Existing `$RATTLER_BUILD_PACKAGE_FILES` entries are now lexically normalized, closing a relative `../` traversal gap while preserving absolute/relative in-prefix paths.
- Selector scripts and generated files are part of serialized rendered metadata, so package rebuilds preserve the behavior.

## Validation

- `cargo test --workspace`
- `cargo check --all-targets`
- `cargo clippy --all-targets -- -D warnings`
- Rust formatting, Ruff, Prettier (YAML), Typos, and `git diff --check`
- 4 focused cross-platform E2E cases:
  - one common build partitioned into runtime, development, and docs artifacts;
  - static globs plus a script-generated CMake glob;
  - requirements/about/build-number overrides and exact parent pinning;
  - `--up-to` child-only build preserving parent `PKG_*` identity;
  - overlap rejection;
  - traversal rejection for generated manifests.

The normal package tests embedded in all three successful artifacts also pass.

### DIFF
--- a/crates/rattler_build_core/src/consts.rs
+++ b/crates/rattler_build_core/src/consts.rs
@@ -22,3 +22,12 @@ pub const RATTLER_BUILD_PACKAGE_FILES: &str = "RATTLER_BUILD_PACKAGE_FILES";
 /// and namespaced so that build scripts are extremely unlikely to clobber it
 /// by accident.
 pub const PACKAGE_FILES_LIST_NAME: &str = ".rattler-build-package-files";
+
+/// Selector-script output containing prefix-relative paths/globs for one nested subpackage.
+pub const RATTLER_BUILD_SUBPACKAGE_FILES: &str = "RATTLER_BUILD_SUBPACKAGE_FILES";
+
+/// Directory under the restored work tree holding generated subpackage selections.
+pub const SUBPACKAGE_FILES_DIR: &str = ".rattler-build-subpackage-files";
+
+/// Cached copy of the parent build's explicit package file list.
+pub const SUBPACKAGE_PARENT_FILES: &str = "parent-package-files.txt";
--- a/crates/rattler_build_core/src/packaging.rs
+++ b/crates/rattler_build_core/src/packaging.rs
@@ -1,7 +1,7 @@
 //! This module contains the functions to package a conda package from a given
 //! output.
 use std::{
-    collections::{HashMap, HashSet},
+    collections::{BTreeMap, HashMap, HashSet},
     io::Write,
     path::{Component, Path, PathBuf},
 };
@@ -103,6 +103,20 @@ pub enum PackagingError {
 
     #[error("Package file `{0}` listed in $RATTLER_BUILD_PACKAGE_FILES does not exist")]
     PackageFileMissing(PathBuf),
+
+    #[error(
+        "Invalid generated subpackage pattern `{0}`: patterns must be prefix-relative and cannot contain `..`"
+    )]
+    InvalidSubpackagePattern(String),
+
+    #[error("Generated subpackage selection `{0}` must be a regular file")]
+    InvalidSubpackageSelectionFile(PathBuf),
+
+    #[error("Subpackage selector did not produce `{0}`")]
+    MissingSubpackageSelectionFile(PathBuf),
+
+    #[error("File `{path}` is selected by more than one subpackage: {packages}")]
+    OverlappingSubpackageSelection { path: PathBuf, packages: String },
 }
 
 /// Split a path into the longest leading directory prefix that contains no glob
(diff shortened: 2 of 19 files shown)


wolfv:
CI update: all checks relevant to this change now pass on Linux, macOS, and Windows x64 (including all four new subpackage E2E tests). The remaining red Windows E2E check is the repository-wide R failure: `test_r_interpreter` and `test_interpreter_detection[r]` crash inside R with `*** stack smashing detected ***`. The same two tests fail on unrelated runs [32538342782](https://github.com/prefix-dev/rattler-build/actions/runs/32538342782) and [32534818668](https://github.com/prefix-dev/rattler-build/actions/runs/32534818668). A failed-job rerun reproduced only those two failures. Windows 11 ARM then cannot download the x64 artifact because the x64 job does not upload it after failing. No subpackage tests fail in the latest run.

PR #2768, opened 2026-08-25
fix: allow legacy-compatible optional dependencies

baszalmstra:
Plain `requirements.extras` produces additive `extra_depends` metadata, which can be represented in legacy packages. Requiring `--v3` prevented recipes from publishing those optional dependency groups without using the V3 repodata layout.

This removes the blanket gate and keeps the MatchSpec-level V3 checks. An extra containing `when`, `flags`, or `extras` still needs `--v3`.

This also updates to the latest rattler release. Every manifest that directly uses `rattler_conda_types` moves to the new 0.52 minor release; the other released rattler crate patches stay in their respective `Cargo.lock` files. V3 repodata now canonicalizes MatchSpecs, so the local publish assertion expects the canonical extra dependency strings.

This is the rattler-build follow-up to conda/rattler#2721.

## Testing

- `pixi run build-release`
- `pixi run python -m pytest test/end-to-end/test_v3_repodata.py -q --snapshot-warn-unused --snapshot-details`
- `pixi run cargo test -p rattler_build_recipe --all-targets`
- `pixi run --environment lint cargo-clippy`
- `pixi run cargo check --manifest-path py-rattler-build/rust/Cargo.toml`
- `pixi run cargo check --manifest-path crates/rattler_build_playground/Cargo.toml`

I also built a no-`--v3` package with `extra_depends` and verified its `info/index.json` has no `repodata_revision`. A fresh local channel indexed with `conda-index` preserved the `extra_depends` record in `repodata.json` and likewise has no `repodata_revision`. Prefix.dev currently rejects the upload to the `baszalmstra` channel as a V3 repodata upload. That is the same stale revision classification addressed by conda/rattler#2721, so the hosted validation needs that change deployed first.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4742,9 +4742,9 @@ dependencies = [
 
 [[package]]
 name = "rattler"
-version = "0.48.4"
+version = "0.48.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6595b5f9878f0e2787da607235c158c7c125d4c4f145cfcc3febade9c66c3a4f"
+checksum = "2c15b3a9ff3a90a018a35810cc4cd2dab365963193667e418671766d3c7a8e19"
 dependencies = [
  "anyhow",
  "astral-reqwest-middleware",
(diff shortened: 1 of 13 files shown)


wolfv:
> An extra containing when, flags, or extras still needs --v3.

Any matchspec, also in regular `depends` requiring an _extra_ should still fail IMO.

wolfv:
Just to be sure, I think we should also reject:

```yaml
package:
  name: foobar
  version: 0.1.0

requirements:
  run:
     - foo [extras=["bla"]]
```

baszalmstra:
Confirmed. A regular dependency such as `foo [extras=["bla"]]` still fails without `--v3` with `invalid bracket key: extras`. I added an end-to-end regression case in 783c80f8.

Hofer-Julian (approved):

PR #2769, opened 2026-08-25
feat: Add an option --pass-all-variants-as-env

isuruf:
usually only used variables are sent to build scripts as env variables. this options sends all single element variables to build scripts. this is useful for conda-forge for variables like CROSSCOMPILING_EMULATOR

### DIFF
--- a/crates/rattler_build_core/src/env_vars.rs
+++ b/crates/rattler_build_core/src/env_vars.rs
@@ -545,24 +545,6 @@ pub fn env_vars_from_variant(
         .collect()
 }
 
-/// Convert the variant keys that only have a single possible value into
-/// environment variables, excluding language-specific keys that are handled
-/// separately. `all_variants` is expected to hold every key and all of its
-/// possible values from the variant configuration, not just the ones a
-/// recipe uses.
-pub fn env_vars_from_single_value_variants(
-    all_variants: &BTreeMap<NormalizedKey, Vec<Variable>>,
-) -> HashMap<String, Option<String>> {
-    let single_value_variants: BTreeMap<NormalizedKey, Variable> = all_variants
-        .iter()
-        .filter_map(|(k, values)| match values.as_slice() {
-            [value] => Some((k.clone(), value.clone())),
-            _ => None,
-        })
-        .collect();
-    env_vars_from_variant(&single_value_variants)
-}
-
 #[cfg(test)]
 mod test {
     use super::*;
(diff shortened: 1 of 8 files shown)


isuruf:
@wolfv thoughts on this? If you agree, I can fix the CI and add some tests

wolfv:
Hmm I don't think I can agree with this change as it makes it harder to recreate Conda-forge recipes.

What would you say if we add a top level "build env" key that will be used in all outputs?

Or actually, for a slightly hacky solution, if you use the variant keys in the global context then they should be available in all build environments. Or you can possibly use them with the global build.variants.use_keys: [foo, bar] and that will (probably) applied to all outputs (if not we should fix it!)

isuruf:
No, requiring changes to hundreds of recipes to add `CROSSCOMPILING_EMULATOR` is too much work and error prone. Also, some variables like `QEMU_LD_PREFIX` are used by compiler activation variables and the recipe author doesn't even know about them. See https://github.com/conda-forge/ctng-compiler-activation-feedstock/blob/main/recipe/activate-gcc.sh#L201 and https://github.com/conda-forge/ctng-compiler-activation-feedstock/pull/146

wolfv:
I've merged #2802 which allows you to define "use_keys" at the top:

```
...

build:
  variants:
     use_keys: ["BLA", "FOO"]
```

These are folded into _every_ output (and staging cache) if defined at the top level which should allow you to write these recipes pretty cleanly. I think together with the lint PR in `conda-smithy` that's an acceptable solution. 

WDYT?

cc @h-vetinari

h-vetinari:
Thanks for working on this!

> These are folded into _every_ output (and staging cache) if defined at the top level which should allow you to write these recipes pretty cleanly.

If every cross-compiled recipe now needs to set
```yaml
build:
  variants:
     use_keys: ["CROSSCOMPILING_EMULATOR", "<etc>"]
```
then that doesn't sound like a feasible solution IMO. That is, unless you'd provide a special key / mechanism in the variant config, which we can use to set all the required variables to be active _once_ in the global pinning, rather than 100s of times in individual feedstocks.

wolfv:
Can we write a CEP to define these as "always used"?

isuruf:
That's gonna take a long time. We need a quick resolution for tests not being run in v1 recipes.

wolfv:
I've made a separate proposal in #2809 - wdyt about that?

PR #2783, opened 2026-08-29
fix: diagnose malformed `skip` expressions instead of ignoring them

### LINKED ISSUE #2780: BUG: mismatched bracket in `skip:` not diagnosed in any way
In https://github.com/conda-forge/clang-win-activation-feedstock/pull/70, I had missed a bracket in a skip condition at some point
```yaml
build:
  number: ${{ build_number }}
  # for win-arm64, only build native clang compilers for now
  skip: target_platform == "win-arm64" or cross_target_platform == "win-arm64") and target_platform != cross_target_platform
       ^                                                                      ^
       !                                                                      !
```

However, rattler-build happily rerender the recipe (ignoring the intent of the skip), and starts building. In fact, it seems completely unfazed by the mismatched brackets. This should IMO be a hard error. It's hard to miss when one's building outputs that are doomed to fail, but skips (and more generally: boolean evaluation) happens in so many places that it's easy to slip into something that doesn't fail the build, but causes problems later.

wolfv:
Fixes #2780

A `skip` entry with an unbalanced bracket was silently treated as "do not skip", so the output got built anyway. `is_skipped()` had `Err(_) => { /* continue */ }`, so an expression that failed to compile just contributed `false`.

`skip` was the only lenient path — `if:` conditions, `${{ }}` templates and variant-config conditionals all propagate Jinja errors. And unlike an `if:` condition, which becomes a `JinjaExpression` (validated at construction), `skip` entries are plain strings, so nothing parsed them until evaluation.

## Fix

**Parse time** — concrete `skip` entries go through `JinjaExpression::new`, a pure syntax check that resolves no variables. Recurses into `if`/`then`/`else`. Templates are left alone: already validated, and their expression is only known once rendered.

**Eval time** — `is_skipped` returns `Result<bool, ParseError>` and propagates *syntax errors*, with the span of the entry. Catches conditions that only exist after a template renders.

## Why only syntax errors are fatal

Propagating every evaluation failure was too aggressive. Under `UndefinedBehavior::Strict` a bare name is undefined (falsy), but almost any operator on it errors:

| Expression (empty variant) | Result |
|---|---|
| `is_abi3` | `Ok(undefined)` → falsy |
| `match(python, "<3.9")` | `Ok(true)` |
| `not is_abi3` | **Err** `UndefinedError` |
| `python == "3.12"` | **Err** `UndefinedError` |
| `undefined_thing \| length` | **Err** `InvalidOperation` |
| `target_platform == "win-arm64") and true` | **Err** `SyntaxError` |

So strict propagation would make `skip: python == "3.12"` a hard failure whenever `python` isn't in the variant, and would break variant discovery, which evaluates skips against incomplete combinations by design. `SyntaxError` is the one kind a missing variable can never produce. `InvalidOperation` is ambiguous (`cmp(...)` is a real bug, `undefined_thing | length` is undefined-driven), so it stays lenient.

`discover_new_variant_keys_from_evaluation` already used `.unwrap_or_default()` and keeps its leniency.

## Result

```
Error:   × Failed to parse recipe
  ╰─▶ Jinja template error: invalid `skip` expression: Failed to parse Jinja
      expression: syntax error: unexpected input after expression
   ╭─[recipe.yaml:9:9]
 8 │ build:
 9 │   skip: target_platform == "win-arm64" or cross_target_platform == "win-arm64") and …
   ·         ───────┬───────
   ╰────
```

Side effect: `test-data/recipes/test-parsing/recipe_bad_skip{,_multi}.yaml` (fixtures with `asd;asd;123`, orphaned from the old parser) now error as intended.

## Tests

Parser tests (single value, list entry, conditional branch, valid-expression control); eval tests for propagation plus `test_is_skipped_tolerates_undefined_variables` guarding the six shapes above; e2e `test_unbalanced_bracket_in_skip`; doc note.

`cargo test --workspace` green except pre-existing `test_system_tool` (no `patchelf` in the dev container).

## Not changed

An undefined variable in a skip condition stays silent (`skip: python == "3.12"` with no `python` quietly means "don't skip"). Same class of silent failure, but wider blast radius and not what was hit here. A `tracing::warn!` would be natural; `rattler_build_recipe` has no `tracing` dep, and the shared path would need to distinguish variant discovery from real evaluation to avoid spurious warnings. Happy to follow up.

### DIFF
--- a/crates/rattler_build_recipe/src/stage0/evaluate.rs
+++ b/crates/rattler_build_recipe/src/stage0/evaluate.rs
@@ -825,7 +825,7 @@ pub fn evaluate_skip_list(
     list: &ConditionalList<String>,
     context: &EvaluationContext,
 ) -> Result<bool, ParseError> {
-    let conditions: Vec<String> =
+    let conditions: Vec<(String, Option<Span>)> =
         evaluate_conditional_list(list.as_slice(), context, |value, ctx| {
             // Skip expressions are Jinja boolean expressions, not templates
             // We need to render them through Jinja to track accessed variables,
@@ -833,11 +833,15 @@ pub fn evaluate_skip_list(
             if let Some(expr_str) = value.as_concrete() {
                 // Wrap in ${{ }} to render as Jinja expression (for variable tracking)
                 let template = format!("${{{{ {} }}}}", expr_str);
-                // Ignore the result - we just want to trigger variable tracking
+                // Ignore the result - we just want to trigger variable tracking.
+                // A render failure is not fatal here: skip conditions may reference
+                // variant keys that are not part of this combination, which strict
+                // template rendering rejects but expression evaluation tolerates.
+                // Malformed expressions are reported by the evaluation below instead.
                 let _ = render_template(&template, ctx, value.span());
 
                 // Always return the original expression string
-                Ok(Some(expr_str.clone()))
+                Ok(Some((expr_str.clone(), value.span().copied())))
             } else if let Some(template) = value.as_template() {
                 // Already a template, render it
                 let rendered = render_template(template.source(), ctx, value.span())?;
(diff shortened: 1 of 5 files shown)


wolfv:
@Hofer-Julian do you remember if we on purpose allow "undefined" variables in skip conditions?

Hofer-Julian:
> @Hofer-Julian do you remember if we on purpose allow "undefined" variables in skip conditions?

Yeah, there were recipes relying on that behavior.

Issue #2730, opened 2026-08-11
BUG: output with no `files:` constraint inherits entire staging output as content


h-vetinari:
In https://github.com/conda-forge/clangdev-feedstock/issues/471, we removed most install scripts, but kept the one for `python-clang`, which looks like
```yaml
  - package:
      name: python-clang
    inherit: clangdev-stage
    build:
      noarch: python
      # Building this output using this recipe is only supported on unix
      # It can still be installed on Windows as it is marked as `noarch: python`
      skip: not (linux and x86_64 and ${{ variant }} == "default")
      string: h${{ hash }}_${{ build_number }}
      script:
        file: build_python_bindings
    requirements:
      host:
        - python ${{ python_min }}.*
      run:
        - python >=${{ python_min }}
        - libclang ==${{ version }}
```
The script is more or less trivial, in that it only does
```bash
mkdir -p "${SP_DIR}"
cp -r clang/bindings/python/clang "${SP_DIR}"
```

In particular, it should be impossible for anything outside of `${SP_DIR}` to be part of that package. What ends up happening however, is that the entire staging output is copied into `python-clang` (see [contents](https://conda-metadata-app.streamlit.app/?q=conda-forge%2Fnoarch%2Fpython-clang-22.1.8-h7055128_5.conda)), which obviously blows up its noarch-ness: https://github.com/conda-forge/clangdev-feedstock/issues/478

My understanding was that an output required at least one of `build.script` or `build.files`; while one can argue whether that's the right thing philosophically, it's a major footgun to just copy everything by default if `files:` isn't present (with no error or lint).

carterbox:
I also ran into this when defining a metapackage (empty package) for dav1d.

https://github.com/conda-forge/dav1d-feedstock/pull/24/changes#diff-ad6aed43f0cf479348ccccec8e7185c4b70582192329f11908dd22a731771702R91

h-vetinari:
#2784 intentionally did not solve this, saying
> Deliberately does **not** change what an _absent_ `files:` key does on an inheriting output — that's #2730, and flipping it would silently empty the common "package everything from staging" pattern.

[...]

PR #2784, opened 2026-08-30
fix: `files: []` packages nothing instead of everything

### LINKED ISSUE #2753: BUG: `files: []` not respected for output that inherits from staging build
In the clangdev feedstock we have
```yaml
  - package:
      name: clang-no-conda-cfg
    inherit: clangdev-stage
    build:
      skip: with_cfg
      string: ${{ variant }}_h${{ hash }}_${{ build_number }}
      files: []
      variant:
        down_prioritize_variant: ${{ 0 if variant == "default" else 1 }}
    requirements:
      run:
        - clang ==${{ version }} ${{ variant }}_nocfg_*_${{ build_number }}
    tests:
      - script:
          - echo "pure metapackage"
```

That's because simply all outputs used `inherit: clangdev-stage` by default. However, what ends up happening is that the output gets [populated](https://conda-metadata-app.streamlit.app/?q=conda-forge%2Flinux-64%2Fclang-no-conda-cfg-22.1.8-default_h2327a0a_6.conda&with_broken=true) with the entire content of the staging output, instead of being a metapackage as intended.

This is black-on-white against the `files: []` specification, and IMO a severe bug.

wolfv:
Fixes #2753

`build.files` was a plain `GlobVec`, so an explicitly empty list was indistinguishable from an absent key. Every consumer skipped the filter with `if !files.is_empty()`, so `files: []` packaged everything instead of nothing.

Same root cause, two more symptoms:
- A conditional `files` list with no matching branch evaluates to empty (the `libclang-cpp`-on-Windows case).
- The output/top-level merge used `if output.files.is_empty() { toplevel.files }`, so `files: []` on an output silently inherited the top-level globs.

Not staging-specific — a plain single-output recipe with `files: []` also packaged everything. Staging just makes it severe: nothing in a restored cache is in a `PrefixRecord`, so the whole tree counts as new.

## Fix

`files` becomes `Option<GlobVec>`, mirroring `about.license_file` (#2598). `None` (no key) packages everything, as before; `Some` lets the globs decide, and `GlobVec::is_match` already returns false for an empty vec. `Files::from_prefix`/`from_paths` take `Option<&GlobVec>`; the merge becomes `output.files.or(toplevel.files)`.

Deliberately does **not** change what an *absent* `files:` key does on an inheriting output — that's #2730, and flipping it would silently empty the common "package everything from staging" pattern.

## Verified

| output | before | after |
|---|---|---|
| `files: []`, inherits staging | `['include/core.h', 'lib/libcore.so']` | `[]` |
| conditional `files`, no branch matches | `['include/core.h', 'lib/libcore.so']` | `[]` |
| `files: []`, plain single output | `['lib/should-not-be-packaged.txt']` | `[]` |
| no `files:` key, inherits staging | `['include/core.h', 'lib/libcore.so']` | unchanged |
| `files: [include/**]` | `['include/core.h']` | unchanged |

Unit tests on `Files` and on the merge, plus e2e `test_staging_empty_files_selection` with a new fixture. Snapshot churn is 25 `files: []` lines disappearing from stage0 snapshots — that is the fix showing up.

`cargo test --workspace` green except pre-existing `test_system_tool` (no `patchelf` in the dev container). fmt and clippy clean.

### DIFF
--- a/docs/reference/recipe_file.md
+++ b/docs/reference/recipe_file.md
@@ -798,6 +798,30 @@ build:
       - include/**/private.h
 ```
 
+Leaving the `files` key out entirely packages every new file. Giving the key
+hands the selection to the globs, so an empty list packages nothing:
+
+```yaml title="recipe.yaml"
+build:
+  # a metapackage: no files, whatever the build produced
+  files: []
+```
+
+The same applies when a conditional list has no matching branch - the key is
+still present, so nothing is selected on the platforms that match no branch:
+
+```yaml title="recipe.yaml"
+build:
+  files:
+    - if: linux
+      then: lib/libfoo.so
+    - if: osx
+      then: lib/libfoo.dylib
+    # on Windows no branch matches, so the package is empty
+```
+
+This matters most for outputs that `inherit` from a staging output, where the
+files restored from the staging cache all count as new.
 
 ### Python specific options
 
--- a/docs/reference/recipe_file.md
+++ b/docs/reference/recipe_file.md
@@ -798,17 +798,17 @@ build:
       - include/**/private.h
 ```
 
-Leaving the `files` key out entirely packages every new file. Giving the key
-hands the selection to the globs, so an empty list packages nothing:
+Omitting `files` packages every new file; giving it hands the selection to the
+globs, so an empty list packages nothing:
 
 ```yaml title="recipe.yaml"
 build:
-  # a metapackage: no files, whatever the build produced
+  # a metapackage, whatever the build produced
   files: []
 ```
 
-The same applies when a conditional list has no matching branch - the key is
-still present, so nothing is selected on the platforms that match no branch:
+A conditional list with no matching branch is still an explicit key, so it too
+packages nothing:
 
 ```yaml title="recipe.yaml"
 build:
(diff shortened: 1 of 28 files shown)


Hofer-Julian (approved):

Hofer-Julian:
As soon as CI is green we can merge this

docs/reference/recipe_file.md
```diff
@@ -798,6 +798,30 @@ build:
       - include/**/private.h
 ```
 
+Omitting `files` packages every new file; giving it hands the selection to the
+globs, so an empty list packages nothing:
```
h-vetinari, line 802: Could we make it so that the presence of `script.file: build_foo` counts the same as `files: [<not_everything_but_just_whatever_the_script_may_install>]`? That's what caught me out in the closely related #2730. My expectation was that if I provide an install script that will explicitly pick/copy stuff into the prefix, then _that's_ what defines the content of the output. Having to then add a  `files: [...]` guard in addition to the custom script is IMO both redundant as well as yet another thing that can easily fall out of sync.

The "inherit everything" approach still seems like a dangerous default to me in general, but I guess it does make sense if the output looks like
```yaml
  - package:
      name: foo
    inherit: foo-stage
    build:
      # neither files: nor script.file:
```

h-vetinari: This approach should still be compatible with what you wrote in the OP (which I saw only now)
> Deliberately does **not** change what an _absent_ `files:` key does on an inheriting output — that's #2730, and flipping it would silently empty the common "package everything from staging" pattern.

Hofer-Julian:
@h-vetinari Let's discuss your suggestion in an issue :)

h-vetinari:
> @h-vetinari Let's discuss your suggestion in an issue :)

Moved my comment to #2730

PR #2787, opened 2026-08-31
fix: run link checks when relocation is disabled

### LINKED ISSUE #2771: Setting `binary_relocation: false` disables link checking
We set `binary_relocation: false` in NVIDIA-managed conda recipes to avoid `patchelf` [corrupting our binaries](https://github.com/NixOS/patchelf/issues/492). We typically do the patching ourselves [in the build script](https://github.com/conda-forge/libcublas-feedstock/blob/424c62eec78dd7dae468c90016633132b034af1c/recipe/build.sh#L31).

When using rattler-build, `binary_relocation: false` disables both binary relocation AND link checking. We expect `binary_relocation` to only affect binary relocation NOT link checking.

For example:

```yaml
    binary_relocation: true
    overdepending_behavior: error
    overlinking_behavior: error
```
I would expect this configuration to perform link checking even if the binaries were not patched/relocated.

cc: @billysuh7 @jakirkham

rootsec1:
`binary_relocation: false` currently returns from post-processing before dynamic binaries are discovered, so it also disables requested overlinking and overdepending checks. The same coupling prevents those checks from running on Windows, where relocation itself is intentionally skipped.

This change discovers every supported binary independently of the relocation selection. Relinking still happens only for paths selected by `binary_relocation`, and never on Windows; the complete discovered set is then passed to the existing linking checks. NoArch and WASM exclusions are unchanged.

Regression coverage exercises an undeclared zlib link with relocation disabled on Linux and Windows, plus a declared-dependency Linux build that records the pre-processing binary hash and verifies the packaged binary is byte-identical.

Verified locally with:
- `pixi run test-end-to-end-ci` (220 passed, 18 skipped)
- `pixi run cargo test --workspace --all-targets --features=recipe-generation -- --nocapture`
- `pixi run -e pre-commit lint`
- `pixi run cargo fmt --all -- --check`
- `pixi run cargo clippy --workspace --all-targets --features=recipe-generation -- -D warnings`

Closes #2771

### DIFF
--- a/crates/rattler_build_core/src/post_process/relink.rs
+++ b/crates/rattler_build_core/src/post_process/relink.rs
@@ -156,7 +156,6 @@ pub fn relink(temp_files: &TempFiles, output: &Output) -> Result<(), RelinkError
         // skip linking checks for wasm
         || target_platform.arch() == Some(Arch::Wasm32)
         || target_platform.arch() == Some(Arch::Wasm64)
-        || relocation_config.is_none()
     {
         return Ok(());
     }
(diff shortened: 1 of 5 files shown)


rootsec1:
Closing this for now to keep my active contribution queue bounded. The implementation and verification are preserved, and I may revisit it after the current reviews settle.

wolfv (approved):

PR #2790, opened 2026-09-01
feat: warn on overlapping output files and unused staging files

wolfv:
Ref #2730 (see also #2753)

Adds two cross-output consistency checks that run after all outputs are built:

- **Overlapping files** — two outputs of the same recipe package the same file, so installing both clobbers it. With staging caches this usually means an output forgot to narrow its `files` selection.
- **Unused staging files** — files from a staging cache were not included in any output that inherits from it (the "silently dropped or silently duplicated" ambiguity from #2730).

Both are warnings by default; `--error-overlapping-files` and `--error-unused-staging-files` turn them into hard errors. Warnings are also recorded on the involved outputs so they appear in build/markdown summaries.

```
⚠ Outputs 'liba' and 'libb' both package 1 file(s), which will clobber each other on installation:
    - lib/libcore.so
⚠ 1 file(s) from staging cache 'stage' were not included in any of the outputs that inherit from it (liba, libb):
    - include/unused.h
```

## How

`create_package` now records the prefix-relative files each output selected for packaging (before noarch path remapping) in the build summary. The checks in the new `output_checks` module compare those sets:

- Overlap: pairwise within one recipe, on prefix-relative paths — so a noarch and an arch output shipping the same prefix file are caught even though their packaged paths differ. Same-name outputs are never compared (they can't be co-installed); variant repeats of a pair are deduped.
- Unused: per cache directory (one per staging output + variant selection), against `prefix_files` from the cache's `metadata.json`, minus files packaging always drops (`CACHEDIR.TAG`, `.pyo`, …). A cache is only judged when **all** of its inheriting outputs were packaged in the run — a consumer skipped by `--skip-existing`/`--noarch-build-platform` or failed under `--continue-on-failure` leaves usage unknown, so no false errors (verified manually with `--skip-existing=local --error-unused-staging-files`).

## Notes

- `test-data/recipes/staging/multiple-staging-caches.yaml` now (correctly) warns: `libcore` and `core-headers` genuinely both package `include/core.h`.
- Flag plumbing follows `--error-prefix-in-binary`; py-rattler-build passes the defaults (can be exposed as kwargs later).
- CLI docs regenerated with `rattler_build_docs`; prose section added to `docs/multiple_output_cache.md`.

## Tests

- e2e `test_staging_consistency_checks`: both warnings fire on a new fixture, and each flag makes the build fail.
- e2e `test_staging_checks_quiet_when_all_files_used`: a recipe whose outputs cover the cache and share nothing builds clean under both strict flags.
- Full workspace tests green except pre-existing `test_system_tool` (no `patchelf` in the dev container); all 23 staging e2e tests pass; fmt and clippy clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_016Rj3CRpTRFpCtsnCVdaaQd)_

### DIFF
--- a/crates/rattler_build_core/src/output_checks.rs
+++ b/crates/rattler_build_core/src/output_checks.rs
@@ -1,0 +1,188 @@
+//! Cross-output consistency checks that run after all outputs of a build are packaged.
+
+use std::{
+    collections::{BTreeSet, HashMap, HashSet},
+    path::PathBuf,
+};
+
+use fs_err as fs;
+
+use crate::{packaging::file_mapper::filter_file, staging::StagingCacheMetadata, types::Output};
+
+/// How many offending files to list per finding before truncating.
+const MAX_LISTED_FILES: usize = 10;
+
+fn format_file_list(files: &BTreeSet<PathBuf>) -> String {
+    let listed = files
+        .iter()
+        .take(MAX_LISTED_FILES)
+        .map(|p| format!("  - {}", p.display()))
+        .collect::<Vec<_>>()
+        .join("\n");
+    if files.len() > MAX_LISTED_FILES {
+        format!(
+            "{}\n  … and {} more",
+            listed,
+            files.len() - MAX_LISTED_FILES
+        )
+    } else {
+        listed
+    }
+}
+
+fn report(findings: Vec<String>, error: bool, flag: &str) -> miette::Result<()> {
+    if findings.is_empty() {
+        return Ok(());
+    }
+    if error {
+        Err(miette::miette!("{}", findings.join("\n\n")))
+    } else {
+        for finding in &findings {
+            tracing::warn!("{}", finding);
+        }
+        tracing::warn!("Pass `{}` to turn this warning into an error.", flag);
+        Ok(())
+    }
+}
+
+/// Warn (or error) when two outputs of the same recipe package the same file.
+///
+/// Installing both packages into one environment would clobber the file, and with
+/// staging caches this usually means an output forgot to narrow its `files` selection.
+/// Compared on prefix-relative paths as selected for packaging (before noarch path
+/// remapping), so a noarch and an arch output that ship the same file are caught too.
+/// Outputs with the same package name are never compared: they cannot be co-installed.
+pub fn check_overlapping_files(outputs: &[Output], error: bool) -> miette::Result<()> {
+    let mut by_recipe: HashMap<&std::path::Path, Vec<&Output>> = HashMap::new();
+    for output in outputs {
+        by_recipe
+            .entry(output.build_configuration.directories.recipe_path.as_path())
+            .or_default()
+            .push(output);
+    }
+
+    let mut findings = Vec::new();
+    for outputs in by_recipe.values() {
+        // Variants of the same pair would repeat the same finding, so keep one per name pair.
+        let mut seen_pairs = HashSet::new();
+        for (i, a) in outputs.iter().enumerate() {
+            let Some(files_a) = a.packaged_prefix_files() else {
+                continue;
+            };
+            for b in &outputs[i + 1..] {
+                if a.name() == b.name() {
+                    continue;
+                }
+                let Some(files_b) = b.packaged_prefix_files() else {
+                    continue;
+                };
+                let overlap: BTreeSet<PathBuf> = files_a.intersection(&files_b).cloned().collect();
+                if overlap.is_empty() || !seen_pairs.insert((a.name().clone(), b.name().clone())) {
+                    continue;
+                }
+                b.record_warning(&format!(
+                    "packages {} file(s) that are also packaged by output '{}'",
+                    overlap.len(),
+                    a.name().as_normalized(),
+                ));
+                findings.push(format!(
+                    "Outputs '{}' and '{}' both package {} file(s), which will clobber each other on installation:\n{}",
+                    a.name().as_normalized(),
+                    b.name().as_normalized(),
+                    overlap.len(),
+                    format_file_list(&overlap),
+                ));
+            }
+        }
+    }
+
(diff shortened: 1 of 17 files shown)


crates/rattler_build_core/src/packaging.rs
```diff
@@ -1042,6 +1042,13 @@ impl Output {
...
         };
 
+        let relative_new_files: BTreeSet<PathBuf> = files_after
+            .new_files
+            .iter()
+            .filter_map(|f| f.strip_prefix(host_prefix).ok().map(Path::to_path_buf))
+            .collect();
+        self.record_packaged_prefix_files(relative_new_files);
+
         package_conda(self, tool_configuration, &files_after)
```
pb01ka, line 1052: As far as I can understand,

- `files_after` is sent to `package_conda`
- In `package_conda` there are two things I'd like to note,
  - [`let mut tmp = files.to_temp_folder(output)?;`](https://github.com/prefix-dev/rattler-build/blob/754646c521f0abf0b65577d9909e202c5b453382/crates/rattler_build_core/src/packaging.rs#L855)
    - which has [`if file_mapper::filter_pyc(f, &self.old_files)`](https://github.com/prefix-dev/rattler-build/blob/754646c521f0abf0b65577d9909e202c5b453382/crates/rattler_build_core/src/packaging/file_finder.rs#L335). This suggests that `pyc` files are removed.
    - it also has, [`if let Some(dest_file) = output.write_to_dest(f, &self.prefix, temp_dir.path())?`](https://github.com/prefix-dev/rattler-build/blob/754646c521f0abf0b65577d9909e202c5b453382/crates/rattler_build_core/src/packaging/file_finder.rs#L339) and it [calls `filter_file`](https://github.com/prefix-dev/rattler-build/blob/754646c521f0abf0b65577d9909e202c5b453382/crates/rattler_build_core/src/packaging/file_mapper.rs#L106) which essentially filters `.pyo`, `.la`, `.DS_Store`.

This is making me curious that wouldn't it be better to call record packaged files after calling `package_conda`. What do you think on this? Please let me know.

py-rattler-build/rust/src/cli_api.rs
```diff
@@ -145,6 +145,8 @@ pub fn build_recipes_py(
         ContinueOnFailure::from(continue_on_failure),
         error_prefix_in_binary,
         allow_symlinks_on_windows,
+        false, // error_overlapping_files
+        false, // error_unused_staging_files
```
pb01ka, line 149: Is there any reason to not expose these two flags to Python API?

crates/rattler_build_core/src/packaging.rs
pb01ka: Another question here,

- Since `relative_new_files` is consumed at - `self.record_packaged_prefix_files(relative_new_files);` before `package_conda`.
- Suppose `package_conda` fails due to some reason - the recorded packaged prefix files still has the data from `relative_new_files`
- My question is - should we be creating and sending `relative_new_files` after calling `package_conda`?

Am I making sense?

crates/rattler_build_core/src/output_checks.rs
```diff
@@ -0,0 +1,179 @@
...
+                    || (a_platform != rattler_conda_types::Platform::NoArch
+                        && b_platform != rattler_conda_types::Platform::NoArch
+                        && a_platform != b_platform)
+                {
+                    continue;
+                }
+                let Some(files_b) = b.packaged_prefix_files() else {
+                    continue;
+                };
+                let overlap: BTreeSet<PathBuf> = files_a.intersection(&files_b).cloned().collect();
```
pb01ka, line 78: As far as I can understand, this checks for intersection between `files_a` and `files_b` which is needed and correct. However files of every `outputs[i]` (our `b`) are checked for intersection against `outputs[0...i - 1]` (our `a`s). I propose the following alternative which might be better (if correct),

- Create a hash map, `files -> owners`, say `f2o`, where `owners` will be a vector.
- For each `outputs[i]` - list all its `files` and insert in `f2o[files[j]].append(outputs[i])`.
- Once `f2o` is ready, only check for those files which have more than one owners. And run the loop:

```rust
overlaps = hash map of (owner1, owner2) -> overlap count
for i in range(len(f2o[file])):
  for j in range(i + 1, len(f2o[file])):
    # run your name and platform checks
    # if passed just continue
    # otherwise store or increase the overlap counter in overlaps hashmap using
    # f2o[file][i] and f2o[file][j]

for overlap in overlaps:
    # record warning officially here
    # since now we have all the counts
```

- The ordering of outputs will be preserved since they are inserted in the same order in the `f2o` hashmap as they appear in original `outputs` vector.

The loop in the diff has O(n * n * number_of_files_per_owner) complexity. The above idea has O(total_number_of_files_across_all_owners + n * n + m * m * total_number_of_files_across_all_owners). `m < n`. 

The benefit of the above proposal relies on the assumption that `m << n` or `m` will be a constant. The loop in the diff will always be in worst case mode. However, the alternative will match the worst case only if `m = n` i.e., all files are shared by all outputs which I think will not be that common in practical.

What do you think on this? Please let me know.

pb01ka (commented):
I reviewed this on best effort basis using my best possible judgement. Please correct me wherever you feel the need to do so. TY.

P.S. I enjoyed going through the diff. It required me to juggle between different files of the code base which is fun in its own sense.

crates/rattler_build_core/src/packaging.rs
wolfv: Thanks, yes, that made sense. Now we use the properly filtered files :)

crates/rattler_build_core/src/output_checks.rs
wolfv: I think we don't need to worry crazily about performance here at this point. Maybe later!

PR #2795, opened 2026-09-02
fix: Set SRC_PATH in tests

### LINKED ISSUE #2793: `SRC_DIR` is not set in the test environment
**rattler-build version:** 0.75.0

During the **build** phase rattler-build exports `SRC_DIR`. During the **test**
phase it does not — even though `CONDA_BUILD=1`, `CONDA_BUILD_STATE=TEST`,
`PREFIX`, `PKG_NAME`, `PKG_VERSION` and `PKG_BUILD_STRING` are all still set, and
`HOME` is set to exactly the directory that would be `SRC_DIR`.

conda-build sets `SRC_DIR` during tests (pointing at its test working
directory), so scripts that rely on it break only under rattler-build.

## Reproducer

`recipe.yaml`:

```yaml
schema_version: 1

package:
  name: srcdir-repro
  version: "0.1.0"

build:
  script:
    - echo "build SRC_DIR=[${SRC_DIR:-<unset>}]"

tests:
  - script:
      - echo "test SRC_DIR=[${SRC_DIR:-<unset>}]"
      - test -n "${SRC_DIR:-}"
```

```
rattler-build build --recipe recipe.yaml
```

**Expected:** `SRC_DIR` set in both phases.

**Actual:**

```
│ │ build SRC_DIR=[$SRC_DIR]
...
│ test SRC_DIR=[<unset>]
│ × error Script failed with status 1
```

(`$SRC_DIR` in the build line is rattler-build's own log redaction of a real
path — it is genuinely set there. In the test phase the shell prints the literal
`<unset>` fallback.)

Incidentally, the log redactor shows that rattler-build already knows the test
work directory *as* `SRC_DIR`: it rewrites the test prefix
`.../test_srcdir-reproJfEQAy/test_run_env` as `$SRC_DIR_run_env`. So the work
directory is registered under that name internally, it just isn't exported to
the test script.

## Why it matters

conda-forge's compiler activation scripts interpolate `${SRC_DIR}` into
`CFLAGS`/`CXXFLAGS`/`FFLAGS` when `CONDA_BUILD=1`. With `SRC_DIR` unset those
flags come out malformed, which breaks recipes that run a compiler in their
tests (we hit this in conda-forge/go-feedstock and had to work around it in the
test script).

## Suggested fix

Export `SRC_DIR` in the test environment, pointing at the test working
directory — the same path `HOME` already points at. That matches conda-build and
is consistent with the other conda-build-compatible variables rattler-build
already provides.

---

xref https://github.com/conda-forge/go-feedstock/pull/338

pavelzw:
closes #2793

### DIFF
--- a/crates/rattler_build_core/src/env_vars.rs
+++ b/crates/rattler_build_core/src/env_vars.rs
@@ -495,15 +495,21 @@ pub fn vars(output: &Output, build_state: &str) -> HashMap<String, Option<String
 /// Return environment variables that should be set during the test process.
 ///
 /// This is a minimal subset of [`vars`] appropriate for test scripts — it does
-/// not include build-time paths (`RECIPE_DIR`, `SRC_DIR`, …) or pip settings.
+/// not include build-time paths (`RECIPE_DIR`, …) or pip settings. `SRC_DIR`
+/// points at the test working directory, matching conda-build's behavior of
+/// setting it to the directory the test runs in (conda-forge's compiler
+/// activation scripts interpolate `${SRC_DIR}` into compiler flags whenever
+/// `CONDA_BUILD` is set).
 pub fn test_vars(
     target_platform: Platform,
     build_platform: Platform,
     host_platform: Platform,
+    work_dir: &Path,
 ) -> HashMap<String, Option<String>> {
     let mut vars = HashMap::new();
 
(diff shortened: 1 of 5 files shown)


wolfv:
Thanks!

wolfv (approved):

PR #2796, opened 2026-09-03
fix: schedule sibling outputs when their dependencies are ready

baszalmstra:
## Why

Multi-output recipes can currently build in an order that is topologically valid but still cannot be installed. For example:

```mermaid
graph LR
    B[consumer B] -->|build or host| A[wrapper A]
    A -->|run| C[runtime C]
```

The hard dependency graph allows `A, B, C`, but building `B` after `A` still fails because `A` cannot be installed until `C` exists. Treating every runtime dependency as a hard topological edge is not an option either: runtime cycles are valid and would become false build cycles.

This also affects `py-rattler-build`, where `Stage0Recipe.run_build()` builds the rendered variants one at a time in render order.

## What changed

Hard build requirements still determine whether the graph is valid. Plain sibling runtime dependencies are now a soft ordering preference, with a stable fallback when they form a cycle. The build loop then uses the dependencies finalized by earlier builds to pick the next output whose local build and host closure is actually available.

```mermaid
flowchart LR
    R[Rendered outputs] --> S[Hard sort plus soft runtime preference]
    S --> Q[BuildQueue]
    Q --> C{Local dependency closure ready?}
    C -->|yes| B[Build output]
    B --> P[Record finalized dependencies]
    P --> Q
    C -->|no| Q
```

The queue itself is generic and only handles stable rotation. `OutputBuildQueue` owns the package-specific rules: exact sibling pins, finalized runtime closures, strong run exports, ignored exports, merged build/host environments, and test readiness. I kept the queue linear; these lists are small compared with the work of building an output, and an event-driven dependency index would add quite a bit of machinery here.

`--up-to` now selects sibling runtime packages and relevant run exports contributed by direct build or host dependencies. It does not pull in run exports emitted by the target itself. Deferred tests are also drained after the final build attempt, so a failed or test-skipped final output no longer leaves earlier tests behind silently.

Strict builds return a typed error when every remaining output is blocked. With `--continue-on-failure`, the queue tries the first blocked output and lets the normal build error handling continue.

## Testing

- Generic queue tests: 3 passed.
- Topological-sort tests: 5 passed.
- Plain runtime ordering tests: 2 passed.
- `--up-to` unit tests: 2 passed.
- Focused scheduler end-to-end tests against the debug binary: 13 passed. These cover plain runtime dependencies, runtime cycles, skipped local outputs, strong and transitive run exports, both ignore forms, merged environments, exact variant pins, deferred tests, and blocked queues.
- `cargo clippy --bin rattler-build -- -D warnings`
- `cargo fmt --all -- --check`
- Ruff on the Python regression test.

I added a Python-facing render-order regression test, but could not run that test locally. The editable extension build stops in `aws-lc-sys` because its build environment is missing CMake/NASM. The same sorting path is covered by the Rust tests and the CLI end-to-end test.


crates/rattler_build_recipe/src/variant_render.rs
```diff
@@ -567,96 +590,189 @@ fn filter_up_to(
...
-                }
+            soft_fallback.get_or_insert(index);
+            if !variant_dependencies_are_added(
+                index,
+                &variants,
+                name_to_indices,
+                &added,
+                extract_soft_dependency_names(&variants[index].recipe),
+            ) {
+                continue;
```
Hofer-Julian, line 818: This soft check silently produces cross-variant `pin_subpackage(exact=True)` dependencies.

Deferring a variant until its runtime provider is emitted can push it past a later variant of its own *hard* provider. `subpackages` in `src/lib.rs` is keyed by name and overwritten per emitted variant, so the consumer then resolves its pin against the wrong variant.

Recipe with `flavor: [a, b]`:

```yaml
outputs:
  - package: {name: foo-core}
    build: {script: echo "core ${{ flavor }}"}
  - package: {name: foo}
    build: {script: echo "foo ${{ flavor }}"}
    requirements:
      host: ["${{ pin_subpackage('foo-core', exact=True) }}"]
      run: [foo-data]          # the only line that matters
  - package: {name: foo-data}
    build: {noarch: generic, script: echo data}
```

Render order becomes `foo-core(a), foo-data, foo-core(b), foo(a), foo(b)`, so both `foo` variants read `subpackages["foo-core"]` after flavour b overwrote it:

| | flavour a | flavour b |
|---|---|---|
| `ec1ce3d3` | `foo-core ==1.0.0 hadb1255_0` | `h4e707db_0` |
| this branch | `foo-core ==1.0.0 h4e707db_0` | `h4e707db_0` |

The build exits 0 with no warning. The flavour-a package's own hash still encodes flavour a, but it depends on the flavour-b core. Drop the two `run:` lines and it is correct again.

`recipe-sibling-pin-variant.yaml` happens not to hit this shape, so the new test gives false confidence here.

src/build_queue.rs
```diff
@@ -0,0 +1,613 @@
...
+            Dependency::PinSubpackage(pin) => {
+                let name = &pin.pin_subpackage.name;
+                let subpackage = consumer
+                    .build_configuration
+                    .subpackages
+                    .get(name)
+                    .ok_or_else(|| BuildQueueError::MissingPinnedSubpackage {
+                        subpackage: name.as_normalized().to_string(),
+                        consumer: consumer.identifier(),
+                    })?;
```
Hofer-Julian: A `pin_subpackage` in *test* requirements now aborts the whole build.

Test requirements create no ordering edge (`extract_dependency_names` covers build/host, run pins, run_export pins and exact run_constraint pins, not `tests[].requirements`), so a pin to a later-defined sibling is legitimately absent from `subpackages`. Turning that into a fatal error breaks recipes that built fine before:

```yaml
outputs:
  - package: {name: early}
    build: {noarch: generic, script: echo early}
    tests:
      - script: [echo test]
        requirements:
          run: ["${{ pin_subpackage('late', exact=True) }}"]
  - package: {name: late}
    build: {noarch: generic, script: echo late}
```

`ec1ce3d3` exits 0, this branch exits 1 with `sibling 'late' pinned by 'early-1.0.0-h4616a5c_0' is missing from the build plan`.

`serialize_test.rs` already handles the identical lookup by falling back to just the package name when the subpackage is not in the map. I would do the same here and return `Ok(None)` instead of erroring, so readiness degrades to name matching.

src/build_queue.rs
```diff
@@ -0,0 +1,613 @@
...
+        visited: &mut HashSet<usize>,
+        closure: &mut Vec<usize>,
+    ) -> bool {
+        if !self
+            .planned_outputs
+            .iter()
+            .any(|output| Self::identifier_matches_spec(spec, output))
+        {
+            return true;
+        }
```
Hofer-Julian, line 181: This decides "not installable" purely from `planned_outputs` and `done_outputs`, with no way for a channel to satisfy the spec. In strict mode that aborts builds the solver used to resolve fine.

Seed `fb-consumer 1.0.0` into the output dir, then build:

```yaml
context: {version: "2.0.0"}
recipe: {name: fb, version: "${{ version }}"}
outputs:
  - package: {name: fb-consumer}
    build: {noarch: generic, script: echo consumer}
    requirements:
      host: [fb-provider ==${{ version }}]
  - package: {name: fb-provider}
    build: {noarch: generic, script: echo provider}
    requirements:
      run: [fb-consumer >=1.0.0]   # satisfiable from the channel by 1.0.0
```

`ec1ce3d3` exits 0 and builds all three artifacts. This branch exits 1 after building only `fb-provider`, with `could not determine a buildable output; blocked outputs: fb-consumer-2.0.0-h4616a5c_0`. `--continue-on-failure` is the only escape hatch today.

`recipe-sibling-blocked-cycle.yaml` locks this in as expected behaviour, but it does not distinguish the genuinely unsatisfiable case from the one a channel can satisfy. Either consult available repodata before declaring a spec locally blocked, or make `Blocked` advisory in the default policy too and let the solver produce the error.

src/build_queue.rs
```diff
@@ -0,0 +1,613 @@
...
+            let mut candidate_closure = closure.clone();
+            if self.collect_install_closure(
+                output_index,
+                &mut candidate_visited,
+                &mut candidate_closure,
+            ) {
+                *visited = candidate_visited;
+                *closure = candidate_closure;
+                return true;
+            }
```
Hofer-Julian, line 199: This tries every matching `done_output` and clones `visited`/`closure` per candidate with no memoization of failures, so a failing walk costs O(variants ^ depth).

With 4 variants and a chain of sibling run deps I measured 6.4s at 10 levels and 5m19s at 13, against 2.3s on `ec1ce3d3`, which fails fast with a precise solver error. The 64x jump for three extra levels is exactly 4³.

Caching `output_index -> bool` for the duration of one `can_build` would fix it.

crates/rattler_build_recipe/src/variant_render.rs
```diff
@@ -567,96 +590,189 @@ fn filter_up_to(
...
+                    add_run_export_targets(&provider.requirements().run_exports.weak);
+                }
+            }
+        }
+
+        for dependency in &requirements.run {
+            if let Some(name) = dependency.name()
+                && let Some(indices) = name_to_indices.get(name)
+            {
+                pending.extend(indices);
```
Hofer-Julian: Following plain `run` deps transitively here means `--up-to` now selects the whole runtime closure, which is a bigger behaviour change than it looks.

For `base <- middle (run: plugin) <- top`, `--up-to middle` selects `base, middle` on `ec1ce3d3` and `base, plugin, middle` here. For a three output `a -> b -> c -> a` run cycle, `--up-to a` selects just `a` on `ec1ce3d3` and all three here, so in any recipe with a runtime cycle `--up-to <anything>` is now equivalent to building everything.

It is also not conditioned on the test strategy, so `--up-to X --test=skip` still drags in a closure only the test env would need.

The help text in `src/opt.rs` still says just "Build recipes up to the specified package". Worth documenting, and worth skipping the run-dep expansion when tests are skipped.

src/build_queue.rs
```diff
@@ -0,0 +1,613 @@
...
+        subpackage: String,
+        /// Output containing the pin.
+        consumer: String,
+    },
+
+    /// Every pending output depends on another unavailable output in the same plan.
+    #[error("could not determine a buildable output; blocked outputs: {blocked_outputs}")]
+    NoBuildableOutput {
+        /// Comma-separated identifiers retained in queue order.
+        blocked_outputs: String,
```
Hofer-Julian: This error is a lot less actionable than what it replaces. For `recipe-sibling-blocked-cycle.yaml`, `ec1ce3d3` prints:

```
× Failed to resolve dependencies
╰─▶ Cannot solve the request because of: repro-blocked-provider ==1.0.0
    cannot be installed because there are no viable options:
    └─ repro-blocked-provider 1.0.0 would require
       └─ repro-blocked-consumer ==1.0.0, for which no candidates were found.
```

and this branch prints `could not determine a buildable output; blocked outputs: repro-blocked-consumer-1.0.0-h...`, which does not say which edge blocks it. Carrying the first unsatisfied dependency per blocked output would keep the diagnostic useful.

crates/rattler_build_recipe/src/variant_render.rs
```diff
@@ -448,14 +456,16 @@ fn build_recipe_graph(
...
 }
 
-/// Perform a stable topological sort on the dependency graph using Kahn's algorithm.
+/// Performs a stable topological sort on the hard dependency graph.
 ///
-/// Uses a BTreeSet so that items with no dependency relationship preserve
-/// their original order. Detects cycles and reports a package name involved.
+/// Among hard-ready items, plain sibling runtime dependencies are preferred.
+/// If every hard-ready item is soft-blocked, the first one is selected so runtime
+/// cycles retain recipe order. Hard cycles still produce an error.
```
Hofer-Julian: "cycles retain recipe order" only holds for the two-output cycle in the new test. With `a -> b -> c -> a` on plain run deps I get `a, c, b`, deterministically across three render and three build runs. The order is valid and stable, so this is just the comment overstating it.

crates/rattler_build_recipe/src/variant_render.rs
```diff
@@ -542,14 +562,17 @@ fn build_recipe_name_index(
...
 }
 
-/// Filter the dependency graph to only include items needed to build `up_to` package.
+/// Filters to the outputs needed to build and test `up_to`.
 ///
-/// Performs a DFS post-order traversal from all variants of the target package,
-/// returning only the transitive dependencies and the target itself as item indices.
+/// Build-order dependencies come from `graph`. Sibling runtime dependencies and
+/// relevant run exports from direct build or host dependencies are followed for
+/// selection only, so they do not create ordering cycles.
```
Hofer-Julian: `filter_up_to` never walks `recipe.tests()`, so "needed to test it" overstates what it does. On `test-data/recipes/race-condition/recipe.yaml`, `--up-to my-package-a` selects only `my-package-a` while its script test requires `my-package`. That test then defers forever and lands in the unconditional drain at the end of the build loop.

Hofer-Julian (commented):
I let an agent run an adversarial review on this

Hofer-Julian (approved):
Thank you!

PR #2801, opened 2026-09-07
feat: expose exclude-newer policies for Python builds, tests, and debug sessions

jezdez:
Add a reusable `ExcludeNewer` Python object carrying global, package, and channel cutoffs plus unknown-timestamp handling. Pass it through the existing `exclude_newer` argument on build, standalone package test, and debug APIs. Existing build datetime arguments and positional calls remain compatible, and debug sessions retain the policy when adding packages.

Historical cutoffs currently exclude the newly built package from its own test environment. Exempt the build output channel and temporary test artifact channel while preserving explicit package override precedence. External channels still obey the policy, including authenticated channel URLs.

This supplies the upstream API support for conda/conda-build#6124, which adds `--exclude-newer` for `recipe.yaml` recipes as part of conda/conda#15759. The conda-build integration needs a release containing these bindings.

### DIFF
--- a/py-rattler-build/rust/src/build.rs
+++ b/py-rattler-build/rust/src/build.rs
@@ -19,7 +19,7 @@ use std::{
 };
 
 use crate::error::RattlerBuildError;
-use crate::exclude_newer::exclude_newer_policy;
+use crate::exclude_newer::PyExcludeNewer;
 use crate::render;
 use crate::repodata_revision::PyRepodataRevision;
 use crate::run_async_task;
@@ -250,7 +250,7 @@ use std::path::Path;
 /// directly without needing to write temporary files.
 #[pyfunction]
 #[allow(clippy::too_many_arguments)]
-#[pyo3(signature = (rendered_variant, tool_config, output_dir, channels, progress_callback=None, recipe_path=None, no_build_id=false, package_format=None, no_include_recipe=false, exclude_newer=None, env_isolation=PyEnvironmentIsolation::Strict, sibling_variants=Vec::new(), repodata_revision=None, *, exclude_newer_package=None, exclude_newer_channel=None, exclude_newer_include_unknown_timestamp=false))]
+#[pyo3(signature = (rendered_variant, tool_config, output_dir, channels, progress_callback=None, recipe_path=None, no_build_id=false, package_format=None, no_include_recipe=false, exclude_newer=None, env_isolation=PyEnvironmentIsolation::Strict, sibling_variants=Vec::new(), repodata_revision=None))]
 pub fn build_rendered_variant_py(
     py: Python<'_>,
     rendered_variant: render::PyRenderedVariant,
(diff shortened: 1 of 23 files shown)


wolfv:
Thanks!

py-rattler-build/rust/src/build.rs
```diff
@@ -249,7 +250,7 @@ use std::path::Path;
 /// directly without needing to write temporary files.
 #[pyfunction]
 #[allow(clippy::too_many_arguments)]
-#[pyo3(signature = (rendered_variant, tool_config, output_dir, channels, progress_callback=None, recipe_path=None, no_build_id=false, package_format=None, no_include_recipe=false, exclude_newer=None, env_isolation=PyEnvironmentIsolation::Strict, sibling_variants=Vec::new(), repodata_revision=None))]
+#[pyo3(signature = (rendered_variant, tool_config, output_dir, channels, progress_callback=None, recipe_path=None, no_build_id=false, package_format=None, no_include_recipe=false, exclude_newer=None, env_isolation=PyEnvironmentIsolation::Strict, sibling_variants=Vec::new(), repodata_revision=None, *, exclude_newer_package=None, exclude_newer_channel=None, exclude_newer_include_unknown_timestamp=false))]
```
wolfv: I am wondering if we should turn this into a Python object that carries these three fields?

wolfv:
Any ideas how this should (or should not) translate to the `rattler-build` CLI?

jezdez:
> Any ideas how this should (or should not) translate to the `rattler-build` CLI?

I'd love to see a similar option like in https://github.com/conda/conda-build/pull/6113

py-rattler-build/rust/src/build.rs
jezdez: Yeah, that would work for me, want me to take a stab at it?

jezdez: Implemented in [`fbe0e7b` (this PR)](https://github.com/prefix-dev/rattler-build/pull/2801/commits/fbe0e7b4642a6b0bb353234010a286945d33c1c1)

jezdez:
@wolfv, rebased onto current `main` and added the `ExcludeNewer` Python object as suggested. It carries the global cutoff, package and channel overrides, and unknown-timestamp handling. Build, package test, and debug methods accept it through `exclude_newer`. Existing build methods still accept a datetime, while the new test and debug arguments take the policy object only. The conda-build integration in conda/conda-build#6124 is updated too.

For the CLI, I think repeatable `--exclude-newer-package NAME=TIMESTAMP|none` and `--exclude-newer-channel CHANNEL=TIMESTAMP|none` flags would be a straightforward addition alongside the existing global cutoff. The core policy handling is already in place. Retaining a cutoff automatically between separate `debug setup` and `debug host-add` invocations would need session-state handling. These CLI additions are still a proposal.

CI is green on 6638c1fa.

PR #2803, opened 2026-09-07
chore: update dependencies

wolfv:
## Summary
- upgrade Cargo dependencies to their latest releases, using compatibility ranges for rattler crates where practical
- upgrade Pixi dependencies with semver-style constraints and refresh lockfiles
- adopt Ruff 0.16 defaults and fix the newly enabled diagnostics
- adapt cargo-deny, zizmor, and archive entry sizing to updated tool APIs
- declare the `zstandard` end-to-end test dependency explicitly

## Testing
- `pixi run -e pre-commit lint`
- `cargo test --workspace --all-targets`
- `cargo check --manifest-path py-rattler-build/rust/Cargo.toml --all-targets`
- `pixi run --manifest-path py-rattler-build/pixi.toml -e test test`
- `pixi lock --check --manifest-path pixi.toml`
- `pixi lock --check --manifest-path py-rattler-build/pixi.toml`


ruff.toml
```diff
@@ -0,0 +1 @@
+lint.select = ["E4", "E7", "E9", "F"]
```
Hofer-Julian: Would be cool to use the defaults and let the agent fix problems with that

Hofer-Julian: Those are nice improvements!

PR #2810, opened 2026-09-09
fix: always use CROSSCOMPILING_EMULATOR variant

wolfv:
Forward `CROSSCOMPILING_EMULATOR` from the runtime environment into build scripts. Variant configuration can override the inherited value, and configured values are always included in the build matrix and package hash.

Includes regression coverage and documentation.


isuruf:
There's `CMAKE_TEST_LAUNCHER` and `CMAKE_CROSSCOMPILING_EMULATOR` that we pass as well.

wolfv:
Good point — added `CMAKE_TEST_LAUNCHER` and `CMAKE_CROSSCOMPILING_EMULATOR` to the same forwarding/variant handling, tests, and docs.

wolfv:
@isuruf couldn't those be injected by the CMake activation script based on the value from CROSSCOMPILING_EMULATOR?

isuruf:
Test launcher for x86_64-v4 and the cross compiling emulator are two different things even though we use CROSSCOMPILING_EMULATOR for both. So, yes CMAKE_CROSs can be derived, but not the test launcher.

We also have some windows specific env var for skipping tests if no emulator. Need to dig that up

wolfv:
Moved the CMake-specific mapping into conda-forge/cmake-feedstock#273 and narrowed this PR back to the generic `CROSSCOMPILING_EMULATOR` variable.

isuruf:
Thank you!

PR #2823, opened 2026-09-19
fix: retry source downloads that arrive empty or truncated

chrisburr:
This works around the issue described in [gitlab-org/gitlab#471955](https://gitlab.com/gitlab-org/gitlab/-/work_items/471955) where you can get an empty body with a HTTP 200.

I'm opening as getting a completely empty body could make sense to detect rather than return a checksum error that the content is `SHA256("")`. That said, I'm not sure if this is a good idea to merge, so feel free to close.


chrisburr:
To give an idea of how much I've seen this issue, this will hopefully succeed on it's 6th attempt: https://github.com/conda-forge/fastframes-feedstock/actions/runs/35524438966/job/106139719902?pr=23

PR #2825, opened 2026-09-22
docs: Document that  --env-isolation none is needed in sccache tips

### LINKED ISSUE #2824: sccache/ccache documented tips do not work with default environment isolation
The documentation at https://rattler-build.prefix.dev/latest/tips_and_tricks/#using-sccache-or-ccache-with-rattler-build suggests to use `sccache` with:

```sh
export CMAKE_C_COMPILER_LAUNCHER=sccache
export CMAKE_CXX_COMPILER_LAUNCHER=sccache

rattler-build build --no-build-id --recipe ./path/to/recipe.yaml
```

However, since https://github.com/prefix-dev/rattler-build/pull/2372 (rattler-build 0.62.0), the default environment isolation is `strict`, so these variables are not forwarded to the build script.

I wonder the minimal way to document this to work is to restore the `<0.62.0` behaviour, i.e. document:

```sh
export CMAKE_C_COMPILER_LAUNCHER=sccache
export CMAKE_CXX_COMPILER_LAUNCHER=sccache

rattler-build build --env-isolation none --no-build-id --recipe ./path/to/recipe.yaml
```

Ideally, I think it would be cool to have some way to whitelist env variables to pass via command line (a bit like bazel does) without the need to modify the recipe with environment specific details. I guess this could be also useful for example to pass `MAKEFLAGS` in case a global jobserver is used.

traversaro:
Updated build command to include --env-isolation none flag for environment variable visibility.

Fix https://github.com/prefix-dev/rattler-build/issues/2824 .


wolfv:
Thanks!

Hofer-Julian:
Wait, should we really recommend isolation none here?

Can't we just recommend to forwarding these env vars instead?

Hofer-Julian:
With concrete examples and all

traversaro:
> Wait, should we really recommend isolation none here?
> 
> Can't we just recommend to forwarding these env vars instead?

See https://github.com/prefix-dev/rattler-build/issues/2824 . The advantage of the existing sccache integration is that you do not need to modify the recipe, if there was a way to just forward the required env variables I would be happy to use it.

wolfv:
Maybe we should add "escape" vars as a config either from CLI or the variant config (like a special key, similar to what I proposed in #2809 but without pulling them in to the variant).

Hofer-Julian:
> The advantage of the existing sccache integration is that you do not need to modify the recipe

Mmmh, I always added the configuration in the recipe, but I can see why it can be useful to avoid that.

Hofer-Julian:
> Maybe we should add "escape" vars as a config either from CLI 

That sounds like a good idea to me!

